PR #445, opened 2023-02-14
Command cleanup, integration test fixes and code coverage

kartikgupta-db:
- Cleanup command visibility. Only show commands like attach cluster after login. 
- Enable codecov
- Fix integration tests


packages/databricks-vscode/package.json
```diff
@@ -39,6 +39,8 @@
     "activationEvents": [
         "onCommand:databricks.*",
         "onTaskType:databricks",
+        "onView:configurationView",
```
fjakobs, line 42: VS Code told me that those are inferred automatically. But it this might only be for new versions so we'd have to bump the minimal VSCode version.

packages/databricks-vscode/package.json
```diff
@@ -53,81 +55,96 @@
...
                 "title": "Start synchronization (full sync)",
                 "category": "Databricks",
+                "enablement": "databricks.context.activated && databricks.context.loggedIn",
                 "icon": "$(sync)"
             },
             {
                 "command": "databricks.sync.stop",
                 "title": "Stop synchronization",
                 "category": "Databricks",
+                "enablement": "databricks.context.activated && databricks.context.loggedIn",
```
fjakobs, line 112: nice!

packages/databricks-vscode/package.json
```diff
@@ -162,32 +179,44 @@
...
+                "icon": "$(debug-stop)",
+                "enablement": "databricks.context.activated && databricks.context.loggedIn",
+                "category": "Databricks"
             },
             {
                 "command": "databricks.utils.openExternal",
                 "title": "Open link externally",
-                "icon": "$(link)"
+                "icon": "$(link)",
+                "enablement": "false",
```
fjakobs, line 197: Why is this disabled?

packages/databricks-vscode/src/extension.ts
```diff
@@ -24,15 +24,18 @@ import {showQuickStartOnFirstUse} from "./quickstart/QuickStart";
...
 import {PackageJsonUtils, UtilsCommands} from "./utils";
 import {ConfigureAutocomplete} from "./language/ConfigureAutocomplete";
 import {WorkspaceFsCommands, WorkspaceFsDataProvider} from "./workspace-fs";
 import {generateBundleSchema} from "./bundle/GenerateBundle";
+import {CustomWhenContext} from "./vscode-objs/CustomWhenContext";
 
 export async function activate(
     context: ExtensionContext
 ): Promise<PublicApi | undefined> {
+    CustomWhenContext.setActivated(false);
```
fjakobs, line 37: shouldn't we move this line down after the version and workspace checks?

packages/databricks-vscode/src/test/e2e/resources/dummy-test/package.json
```diff
@@ -4,25 +4,20 @@
     "description": "A dummy extension to run integration tests",
     "version": "0.0.1",
     "engines": {
-        "vscode": "^1.74.0"
+        "vscode": "^1.69.0"
```
fjakobs, line 7: Did this break the tests?

packages/databricks-vscode/src/test/e2e/resources/dummy-test/package.json
```diff
@@ -4,25 +4,20 @@
...
-        "watch": "tsc -watch -p ./",
-        "pretest": "yarn run compile && yarn run lint",
-        "lint": "eslint src --ext ts"
-    },
+    "activationEvents": [
+        "*"
+    ],
+    "main": "./src/extension.js",
     "devDependencies": {
         "@types/node": "^18.13.0",
```
fjakobs, line 17: I assume we can get rid of all these dependencies.

fjakobs (changes requested):

packages/databricks-vscode/src/extension.ts
kartikgupta-db: This is to make sure that the extension context is set to deactivated when it fails to activate due to any reason. Which includes failure due to version and workspace checks.

packages/databricks-vscode/src/test/e2e/resources/dummy-test/package.json
kartikgupta-db: Yep. We need the dummy extension to support the same versions as the main extension (and hence wdio).

packages/databricks-vscode/package.json
kartikgupta-db: Good catch. Mistake.

packages/databricks-vscode/package.json
kartikgupta-db: Yeah. This is what the autocomplete tells me as well. But this broke our tests in an annoying way.

fjakobs (approved):

PR #473, opened 2023-02-17
Improve AAD login and prevent infinite loops when trying to use AAD

fjakobs:


packages/databricks-vscode/src/configuration/AzureCliCheck.ts
```diff
@@ -23,13 +26,32 @@ export interface ErrorResult {
...
     let step: KEYS | undefined = start;
     while (step && steps[step]) {
+        counter += 1;
+        if (counter > maxSteps) {
+            throw new OrchestrationLoopError();
+        }
         const result: StepResult<S, KEYS> = await steps[step]();
+        logger?.info(
+            `Azire CLI check: ${step}`,
+            util.inspect(result, {depth: 5})
```
kartikgupta-db, line 52: Logger will automatically do inspect internally. So this should be enough
```suggestion
            result
```

packages/databricks-vscode/src/configuration/AzureCliCheck.ts
```diff
@@ -161,25 +212,44 @@ export class AzureCliCheck implements Disposable {
...
+            m = e.message.match(
+                /Expected aud claim to be: ([a-z0-9-]+?)\/?, but was: ([a-z0-9-]+)\/?/
+            );
+            if (m) {
+                return {
+                    aud: m[1],
+                    canLogin: false,
+                };
+            }
+
```
kartikgupta-db, line 249: AppID can be variable???

packages/databricks-vscode/src/configuration/AzureCliCheck.ts
```diff
@@ -73,26 +98,43 @@ export class AzureCliCheck implements Disposable {
...
                     return {
                         type: "success",
                         result: true,
                     };
                 } else {
-                    return {
-                        type: "next",
-                        next: "findCli",
-                    };
+                    if (loginAttempts >= 3) {
```
kartikgupta-db, line 125: We should increase this. It can reach 3 very quickly
loginAttempt = 1 : App ID incorrect
loginAttempt = 2 : tenant ID incorrect
loginAttempt = 3 : Throws error here

packages/databricks-vscode/src/configuration/AzureCliCheck.ts
kartikgupta-db: We should infact do a logger?.error for errors because that handles non enumerable fields in errors as well.

packages/databricks-vscode/src/configuration/AzureCliCheck.ts
fjakobs: For staging workspaces there is a different AppID

packages/databricks-vscode/src/configuration/AzureCliCheck.ts
fjakobs, line 125:
```suggestion
                    if (loginAttempts >= 4) {
```

fjakobs: bumpted it to 4

kartikgupta-db (approved):

PR #475, opened 2023-02-17
Make integration tests less flaky.

kartikgupta-db:
Most test failures are happening because the host input box is closed when quickstart page appears. This waits for the quickstart page to appear before starting to run tests. 

Renamed quickstart page so that it is unique in case some other extension dependency (like python) decides to publish a quickstart page and break our tests again.


packages/databricks-vscode/src/quickstart/QuickstartCommands.ts
```diff
@@ -6,7 +6,7 @@ export class QuickstartCommands {
     openQuickstartCommand() {
         return async () => {
             const uri = Uri.file(
-                this.context.asAbsolutePath("README.quickstart.md")
+                this.context.asAbsolutePath("DATABRICKS.quickstart.md")
```
fjakobs, line 9: The file is referenced  in the `README.md`, which needs to be updated as well.

What was the reason was renamed?

fjakobs (approved):

PR #478, opened 2023-02-17
Hide existing repos

fjakobs:
No longer offer existing Repos  as sync destination. This avoids the risk of users accidentally overwriting existing files and notebooks.

Before:
Screenshot 2023-02-17 at 16 28 00

After: 
Screenshot 2023-02-17 at 16 27 35


saadansari-db (commented):
I really like it. I wonder if we can add a blurb in there that says "Only Reps create by your IDE shown here" - would like to get @PaulCornellDB word smithing here

PaulCornellDB:
Is there enough space in the "Command Palette" to say much about this to address the "why can't I pick a sync destination that existed before I started using this extension?"

Perhaps we might add a "lowercase-i-inside-of-a-circle" icon next to "Sync Destination" in the "Configuration" pane, that when clicked states something such as "You can only add or select sync destinations that are created by this extension." And if you have more space, add "This prevents the accidental overwriting of existing content in destinations that this extension did not create."

kartikgupta-db (approved):

PR #480, opened 2023-02-17
RFC: Remove Cluster and WSFS views

fjakobs:
This PR removes the cluster and WSFS views as they are not project centric and don't really add to the inner dev loop experience.

Similar functionality can be provided by 3rd party extensions like https://marketplace.visualstudio.com/items?itemName=paiqo.databricks-vscode


fjakobs:
@kartikgupta-db I wonder if we can also simplify the cluster manager after merging this PR

packages/databricks-vscode/src/cluster/ClusterTreeNode.ts
```diff
@@ -0,0 +1,103 @@
...
+        label: "Driver",
+        description: element.details.driver_node_type_id,
+    });
+
+    if (element.details.autoscale) {
+        children.push({
+            label: "Worker",
+            description: `${element.details.node_type_id}, ${element.details.autoscale.min_workers}-${element.details.autoscale.max_workers} workers`,
+        });
+    } else if (element.details.num_workers || 0 > 0) {
```
kartikgupta-db, line 70: 0 > 0 ?

packages/databricks-vscode/src/extension.ts
```diff
@@ -106,30 +105,17 @@ export async function activate(
...
-        window.registerTreeDataProvider(
-            "workspaceFsView",
-            workspaceFsDataProvider
-        ),
         commands.registerCommand(
             "databricks.wsfs.attachSyncDestination",
             workspaceFsCommands.attachSyncDestination,
             workspaceFsCommands
         ),
-        commands.registerCommand(
```
kartikgupta-db, line 128: We can remove wsfs.attachSyncDestination as well and remove these commands from package.json.

kartikgupta-db (commented):
This does simplify a bunch of things!

packages/databricks-vscode/src/extension.ts
fjakobs: +1 will do

PR #486, opened 2023-02-20
Bump bricks

pietern:
This pulls in a change that modifies the args to `bricks sync SRC DST`.


pietern:
The working directory was always implied to be equal to the workspace root through `BricksTasks.ts`.

There is some ambiguity wrt execution vs returning arguments in `CliWrapper.ts` and friends. Ideally we centralize everything related to spawn options and running bricks into a single file. As is, we do not pick up the proxy settings when calling non-sync commands, for example.

kartikgupta-db (approved):

PR #494, opened 2023-02-22
[DECO-531] Transition sync state to ERROR if the underlying sync process fails

shreyas-goenka:


shreyas-goenka:
Tested this with a modified bricks cli that times out at 1 second

https://user-images.githubusercontent.com/88374338/220666087-fd8a97bf-86ea-43e0-89db-22cc040f8b10.mov

fjakobs (approved):

PR #497, opened 2023-02-23
Add an option to silence autocomplete dialog.

fjakobs:
I'm currently getting this dialog even for projects that are not Databricks projects.

PR #498, opened 2023-02-23
Fix parsing `.databrickscfg` profiles that contain a dot

fjakobs:
fixes #497


packages/databricks-sdk-js/src/config/KnownConfigLoader.ts
```diff
@@ -49,3 +49,44 @@ export class KnownConfigLoader implements Loader {
...
+            if (typeof obj[key] === "object") {
+                topLevel.push(key);
+                resp = {
+                    ...resp,
+                    ..._flattenIniObject(obj[key], topLevel, resp),
+                };
+            } else {
+                props[key] = obj[key];
+            }
+        }
```
kartikgupta-db, line 80: This will reduce all nested objects to dots. But we only want to flatten objects  that
1. Start from direct children of the root ini object
2. Form a single chain of nested objects with no branches

Eg for the file
```ini
[a.b.c]
d = something

[one]
d = something
[.two]
e = something
```

We want to flatten the first object but for the second one we should rather throw an error and let the user know that it is an invalid cfg.

fjakobs: This is actually what I want. I just want to treat the dot as a regular character. Added a unit test to document my intentions here.

kartikgupta-db (approved):

PR #520, opened 2023-03-01
Add refresh button to refresh results of a Workflow run

### LINKED ISSUE #470: Run as a Workflow on Databricks is not showing fetched results
Tried to run a notebook using `Run as a Workflow on Databricks` and it gets on hold in `Fetching results` status even though the status in Task Run Details is `Succeeded`.

Screenshot 2023-02-16 at 3 27 41 PM

It will be cool if vscode can recognized the `.py` files as a notebooks format connected to the databricks cluster. i guess it is in the road map.

vscode already recognize the file as an interactive file:

image

but only shows the local kernels:

image

kartikgupta-db:
Fixes #470 

https://user-images.githubusercontent.com/88345179/222450373-98e74cd1-1e36-4c8d-a5da-24609d9c44c3.mov

### DIFF
--- a/packages/databricks-vscode/resources/webview-ui/job.html
+++ b/packages/databricks-vscode/resources/webview-ui/job.html
@@ -34,13 +34,29 @@
                 font-size: small;
             }
 
+            .vscode-button {
+                background-color: var(--vscode-button-background);
+                color: var(--vscode-button-foreground);
+                border-color: var(--vscode-button-border);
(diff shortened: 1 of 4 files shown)


fjakobs:
I love it. I wonder if we should do this in the background say every 30sec to make the experience more like in the webapp. Do you know if we can call this API while the task is still running?

fjakobs (approved):

PR #530, opened 2023-03-04
[Experimental] DBConnectv2 and dbutils integration

kartikgupta-db:


packages/databricks-vscode/resources/python/00-databricks-startup.py
```diff
@@ -0,0 +1,42 @@
...
+    if "databricks.env" in os.listdir(databricksDir):
+        with open(os.path.join(databricksDir, "databricks.env"), "r") as fil:
+            for line in fil.readlines():
+                segs = line.split("=")
+                os.environ[segs[0]] = '='.join(segs[1:])
+
+loadEnv()
+del loadEnv
+
+spark = SparkSession.builder.getOrCreate()
```
fjakobs, line 27: are these globals available in notebook cells?

kartikgupta-db: yes

fjakobs: Nice

PR #532, opened 2023-03-06
Use structured output from sync command

pietern:
This removes the need for complex state tracking and monitors for "start" and "complete" events in the sync command stdout. As we no longer forward log messages to the VS code terminal pane this change includes rudimentary formatting of these messages into something human readable.

We should consider if it is worth keeping this pane or if the status should be consolidated into the "sync" pane.


packages/databricks-vscode/src/cli/BricksSyncParser.ts
```diff
@@ -103,38 +106,23 @@ export class BricksSyncParser {
...
+    // This function processes the JSON output from bricks sync and parses it
+    // to figure out if a synchronization step is in progress or has completed.
+    public processStdout(data: string) {
+        const logLines = data.split("\n");
+        for (let i = 0; i < logLines.length; i++) {
+            const line = logLines[i];
+            try {
+                this.processLine(line);
+            } catch (error: any) {
+                // Log error?
```
pietern, line 122: @kartikgupta-db Which logger should I use here?

pietern:
Note: tests will fail because a bump is required.

packages/databricks-vscode/src/cli/BricksSyncParser.ts
kartikgupta-db: @pietern  Please use the extension logger here. 
`NamedLogger.getOrCreate(Loggers.Extension)`

fjakobs:
A little video is always appreciated

fjakobs (approved):

packages/databricks-vscode/src/cli/BricksSyncParser.ts
pietern: Thanks, done.

PR #546, opened 2023-03-10
Publish extension also to OpenVSIX

fjakobs:
This is required for VSCode distributions that are not published by Microsoft such as Gitpod, or VSCodium as they are not allowed to install extension from the MS Marketplace.

Screenshot 2023-03-10 at 11 30 55


pietern (approved):
LGTM

kartikgupta-db (approved):

nfx (approved):

PR #565, opened 2023-03-15
Prevent relogin when project.json is updated on first login

kartikgupta-db:


fjakobs (approved):
Code looks good. Let's add a random UUID to the auto generated folder name to prevent conflicts. Can be a follow up PR

fjakobs (approved):

PR #587, opened 2023-03-21
Files in Workspace: Show prompt to enable files in workspace and add deprecation banner for repos

tanmay-db:
## Changes
- We provide prompt for users to:
  - Switch to workspace using button if they are on repos and if they meet the conditions to use workspace
  - Ignore for this workspace
- Add deprecation banners for Repos
- Bug Fix: We have a bug in our tooling for syncing files where we don't sync workspace directory which contains the nested files because it is under .gitignore, this leads to files not being synced. Fix is to define workspace under a temporary directory. 

Note: This PR doesn't contain making workspace as default. Flipping that switch separately in another PR: https://github.com/databricks/databricks-vscode/pull/602.

image
image

## Tests
- Add integration test for workspace
- Update existing integration test
- Manual testing


packages/databricks-vscode/src/sync/SyncDestination.ts
```diff
@@ -10,7 +10,7 @@ import {Loggers} from "../logger";
 import {ConnectionManager} from "../configuration/ConnectionManager";
 
 export const REPO_NAME_SUFFIX = ".ide";
-export type SyncDestinationType = "repo" | "workspace";
+export type SyncDestinationType = "repo [deprecated]" | "workspace";
```
kartikgupta-db, line 13: Let us not change the type. This is not visible for the user.

packages/databricks-vscode/src/workspace-fs/WorkspaceFsAccessVerifier.ts
```diff
@@ -65,30 +85,38 @@ export class WorkspaceFsAccessVerifier implements Disposable {
...
+                "Please switch to workspace for better experience, repos has been deprecated";
+            const selection = await window.showErrorMessage(
+                message,
+                "Switch to Workspace",
+                "Ignore"
+            );
+
+            if (selection === "Switch to Workspace") {
+                switchToWorkspace();
+            }
```
kartikgupta-db, line 119: For files in workspace to be enabled:
1. The cluster must be > 11.2
2. The files in workspace option must be enabled on workspace level. 

The 2nd case is not handled here. The way to check for 2nd case is to try uploading a file to the workspace and checking if it is able to upload correctly. This function does that https://github.com/databricks/databricks-vscode/blob/main/packages/databricks-vscode/src/workspace-fs/WorkspaceFsAccessVerifier.ts#L95. 

Can you please add handling for this case as well.

packages/databricks-vscode/src/vscode-objs/WorkspaceState.ts
```diff
@@ -4,6 +4,20 @@ import {ExtensionContext} from "vscode";
...
+        return this.context.workspaceState.get(
+            "databricks.switch.to.workspace",
+            false
+        );
+    }
+
+    set skipSwitchToWorkspace(value: boolean) {
+        this.context.workspaceState.update(
+            "databricks.switch.to.workspace",
+            true
```
kartikgupta-db, line 17:
```suggestion
            value
```

packages/databricks-vscode/src/configuration/ConfigurationDataProvider.ts
```diff
@@ -7,11 +7,18 @@ import {
     TreeDataProvider,
     TreeItem,
     TreeItemCollapsibleState,
+    Command,
```
fjakobs, line 10: let's clean this up

packages/databricks-vscode/src/configuration/ConfigurationDataProvider.ts
```diff
@@ -272,16 +284,29 @@ export class ConfigurationDataProvider
             ];
 
             if (
-                syncDestination.remoteUri.name !== syncDestination.localUri.name
+                workspaceConfigs.syncDestinationType === "repo" &&
+                this.workspaceState.filesInWorkspaceFf &&
```
fjakobs, line 288: `filesInWorkspaceFf` is a strange name. What does `Ff` stand for?

packages/databricks-vscode/src/configuration/ConfigurationDataProvider.ts
```diff
@@ -272,16 +284,29 @@ export class ConfigurationDataProvider
...
+                this.wsfsAccessVerifier.isEnabled
             ) {
                 children.push({
-                    label: "The remote sync destination name does not match the current vscode workspace name",
-                    tooltip:
-                        "If syncing to directory with a different name is the intended behaviour, this warning can be ignored",
+                    label: {
+                        highlights: [[0, 10]],
+                        label: "Deprecated",
```
fjakobs, line 294: I wonder if we can find a better term than `Deprecated`. I also wonder how many users have to stay on repos

packages/databricks-vscode/src/workspace-fs/WorkspaceFsAccessVerifier.ts
```diff
@@ -3,23 +3,77 @@ import {
...
 
-async function switchToRepos() {
+export async function switchToRepos() {
     await workspaceConfigs.setSyncDestinationType("repo");
     commands.executeCommand("workbench.action.reloadWindow");
 }
 
+export async function switchToWorkspace() {
+    await workspaceConfigs.setSyncDestinationType("workspace");
+    commands.executeCommand("workbench.action.reloadWindow");
```
fjakobs, line 19: Is it really required to reload the window? The experience would be better if we wouldn't have to.

packages/databricks-vscode/src/workspace-fs/WorkspaceFsAccessVerifier.ts
```diff
@@ -3,23 +3,77 @@ import {
...
+            dbrVersionParts[1] !== "x" &&
+            dbrVersionParts[1] < 2)
+    );
+}
+
+export async function switchToWorkspacePrompt(
+    workspaceState: WorkspaceStateManager
+) {
+    const message =
+        "Repos as sync destination is deprecated. Please switch to using workspace as the sync destination.";
```
fjakobs, line 36: We should try to find more positive wording. Like

```
The Databricks extension works better when using the workspace as sync destination. Do you want to switch?

Yes, No, Never
```

packages/databricks-vscode/src/vscode-objs/WorkspaceState.ts
```diff
@@ -4,6 +4,38 @@ import {ExtensionContext} from "vscode";
...
+        }
+        return randomNum;
+    }
+
+    get filesInWorkspaceFf() {
+        return this.fixedRandom <= 0.1;
+    }
+
+    get skipSwitchToWorkspace() {
+        return this.context.workspaceState.get(
```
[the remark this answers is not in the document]
kartikgupta-db: Nope

packages/databricks-vscode/src/configuration/ConfigurationDataProvider.ts
kartikgupta-db: Users will only see deprecated if they are on a cluster with dbr >= 11.2

PR #603, opened 2023-03-21
Add Pull Request Template

tanmay-db:
## Changes

PR template to add structure for all pull requests. 

## Tests

Copy pasting the template in this description to see how it looks, would be able to test the changes in Github after this merges (for PRs after this)

## Public Tickets

N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -4,8 +4,8 @@
 ## Tests
 <!-- How is this tested? -->
 
-## Ticket / Issue
+## Public Tickets
 <!--
-Internal ticket -- Jira ticket number only (no links), example: DECO-TicketNumber
-Public ticket (eg: github issue) -- Link to the public ticket
+Link to public tickets (if any). Please don't mention internal tickets. 
 -->
+NA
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,14 @@
...
+## Tests
+
+<!-- How is this tested? -->
+
+## Jira Ticket
+
+<!--
+For linking Jira ticket with PR. This will show up on the ticket.
+Not to be added in PR title as these will need to be removed in the change logs.
+-->
```
kartikgupta-db, line 14: Would be great if you can put in examples here. We do not want people to commit full links to our internal JIRA.

Also, for open source repos, we do not track all issues in JIRA. Infact we want to minimise JIRA exposure and track as much of the work in github issues as possible (so that users have better visibility). Could you add a section for this with examples as well? Wdyt about making it higher priority than JIRA in the template?

tanmay-db: Thanks for the review. Yeah agree, we shouldn't have people add full links, just adding the number should be good since jira automatically picks it up, example just writing:

## Ticket Number (if any)
DECO-613

should be enough. I did the same in this PR and it links it:
image

> github issues as possible (so that users have better visibility). Could you add a section for this with examples as well? Wdyt about making it higher priority than JIRA in the template?
I think having another section for open source issues should be okay but I was thinking on having a single section where people can link (if open source) / mention (if internal jira) the tickets to make PR template less bulky.

tanmay-db: updated the PR

fjakobs (changes requested):

PR #611, opened 2023-03-24
Use waiter to handle long running calls

fjakobs:
## Changes
Use waiter to handle long running calls inspired by the Python SDK.

Previously we had two functions for long running operations:

- `cluster.create(): Promise<CreateClusterResponse>`
- `cluster.createAndWait(): Promise<ClusterInfo>`

where the first would just make the API call to trigger the creation and the second function would wait for completion and return the result of the last polling operation.

This change collapses these two function into one:

```create(...): Promise<Waiter<CreateClusterResponse, ClusterInfo>>```

The returned waiter is compatible with the return value of the previous `create` call. If users want to wait they can call `.wait()` on the return value, which will be equivalent to the `createAndWait` call.

## Example

```typescript
        this.details = await (
            await this.clusterApi.delete(
                {
                    cluster_id: this.id,
                },
                new Context({cancellationToken: token})
            )
        ).wait({
            onProgress: async (clusterInfo) => {
                this.details = clusterInfo;
                if (onProgress) {
                    await onProgress(clusterInfo);
                }
            },
        });
```

### DIFF
--- a/packages/databricks-sdk-js/src/.codegen/api.ts.tmpl
+++ b/packages/databricks-sdk-js/src/.codegen/api.ts.tmpl
@@ -10,6 +10,7 @@ import {CancellationToken} from "../../types"
 import {ApiError, ApiRetriableError} from "../apiError";
 import {context, Context} from "../../context"
 import {ExposedLoggers, withLogContext} from "../../logging";
+import {Waiter, asWaiter} from "../../wait";
 
 {{range $i, $s := .Services }}
 export class {{$s.PascalName}}RetriableError extends ApiRetriableError {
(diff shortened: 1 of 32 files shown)


packages/databricks-sdk-js/src/.codegen/api.ts.tmpl
```diff
@@ -30,88 +31,102 @@ export class {{$s.PascalName}}Service {
...
-		        `{{range  .PathParts}}{{.Prefix}}{{if .Field}}${request.{{.Field.Name}}}{{ else if .IsAccountId }}${this.client.accountId}{{end}}{{ end }}`
-	        {{- else}}"{{.Path}}"{{end}}
-			return (await this.client.request(
-				path,
-				"{{.Verb}}",
-				{{if .Request -}}request{{else}}undefined{{end}}, 
-				context
-			) as {{if .Response}}{{if .Response.ArrayValue}}Array<model.{{.Response.ArrayValue.PascalName}}>{{else}}model.{{.Response.PascalName}}{{end}}{{else}}model.EmptyResponse{{end}})
-		}
+		const path = {{if .PathParts -}}
```
nfx, line 41: once i push out the SDK generator improvement, you can simplify this code even more. stay tuned.

kartikgupta-db (approved):

PR #621, opened 2023-03-28
Write .databricks.env file with the required env variables

kartikgupta-db:


packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
```diff
@@ -0,0 +1,173 @@
...
+                    return;
+                }
+                this.writeFile();
+                this.emitToTerminal();
+            }, this)
+        );
+    }
+
+    async getPatToken() {
+        const headers: Record<string, string> = {};
```
fjakobs, line 84: This should use the metadata service so we don't leak PATs to the file system.
I understand that currently, we can't do that yet because DBConnect doesn't use the SDK yet. Let's at least leave a TODO.

packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
```diff
@@ -0,0 +1,173 @@
...
+        const host = this.connectionManager.databricksWorkspace?.host.authority;
+        const pat = await this.getPatToken();
+        if (!cluster || !pat || !host) {
+            return;
+        }
+        /* eslint-disable @typescript-eslint/naming-convention */
+        return {
+            DATABRICKS_CLUSTER_ID: cluster.id,
+            SPARK_REMOTE: `sc://${host}:443/;token=${pat};use_ssl=true;x-databricks-cluster-id=${cluster.id}`,
+            DATABRICKS_HOST: host,
```
fjakobs, line 111: Why do we need HOST and PAT here?

fjakobs (approved):

PR #630, opened 2023-03-31
[WIP] Add Telemetry to VS Code Extension

mgyucht:
## Changes
This PR adds telemetry to the VS code extension.
1. The `src/telemetry` directory is introduced in the `databricks-vscode` package, containing the main telemetry interface. Telemetry is based on the @vscode/extension-telemetry library.
2. A wrapper function is defined for registering commands, ensuring that command execution is instrumented.
3. An EventEmitter is added to ConnectionManager exposing changes to DatabricksWorkspace (which happens on login/logout). The telemetry library registers a handler to track the latest anonymized user metadata, which is automatically added to events when recorded.

Other changes introduced in this PR:
* Added a development mode tracker, based on environment variables set by the extension's launch configuration.
* Refactored cloud type to a single named type called `Cloud`.

## Tests
- [x] Manual tests: added console.log and tried to run a command.
Screenshot 2023-03-31 at 18 52 22
One interesting point: user metadata listener is only called after the command execution completes, so we only see the user ID in following commands but not in the login command itself.

- [ ] Unit tests: ...


packages/databricks-vscode/package.json
```diff
@@ -628,7 +628,7 @@
         "version": "0.0.27"
     },
     "scripts": {
-        "vscode:prepublish": "rm -rf out && yarn run package:compile && yarn run package:wrappers:write && yarn run package:copy-webview-toolkit",
+        "vscode:prepublish": "rm -rf out && yarn run package:compile && yarn run package:wrappers:write && yarn run package:copy-webview-toolkit && yarn run generate-telemetry",
```
mgyucht, line 631: The command without this change fails on my machine, not sure how to test this.

.gitignore
```diff
@@ -18,3 +18,6 @@ packages/databricks-sdk-js/src/config/testdata/azure/Library
 *.d.ts.map
 *.js.map
 tsconfig.tsbuildinfo
+
+# Telemetry file, automatically generated by packages/databricks-vscode/scripts/generateTelemetry.ts
+packages/databricks-vscode/telemetry.json
```
fjakobs, line 23: This line should go into the `.gitignore` inside of `packages/databricks-vscode`

packages/databricks-vscode/src/configuration/auth/AuthProvider.ts
```diff
@@ -13,6 +13,7 @@ const extensionVersion = require("../../../package.json")
 
 import {AzureCliCheck} from "../AzureCliCheck";
 
+// TODO: Resolve this with SDK's AuthType.
```
fjakobs, line 16: I don't get this TODO

packages/databricks-vscode/src/telemetry/index.ts
```diff
@@ -0,0 +1,144 @@
...
+import {DatabricksWorkspace} from "../configuration/DatabricksWorkspace";
+import {isDevExtension} from "../utils/developmentUtils";
+import {
+    DEV_APP_INSIGHTS_CONFIGURATION_STRING,
+    EventType,
+    EventTypes,
+    PROD_APP_INSIGHTS_CONFIGURATION_STRING,
+} from "./constants";
+import crypto from "crypto";
+import bcrypt from "bcrypt";
```
fjakobs, line 11: Is `bcrypt` a binary dependency? If se we might have to check if our Windows builds still work.

packages/databricks-vscode/src/telemetry/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    const userName = databricksWorkspace.userName;
+    // Salt is the first 22 characters of the hash of the username, meant to guard against
+    // rainbow table attacks. 2^7 iterations takes ~13ms.
+    const salt = crypto
+        .createHash("sha256")
+        .update(userName)
+        .digest("hex")
+        .substring(0, 22);
+    const bcryptSalt = `$2b$07$${salt}`;
+    const hashedUserName = await bcrypt.hash(userName, bcryptSalt);
```
fjakobs, line 56: I think this is sufficiently irreversible.

packages/databricks-vscode/src/telemetry/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    EventType,
+    EventTypes,
+    PROD_APP_INSIGHTS_CONFIGURATION_STRING,
+} from "./constants";
+import crypto from "crypto";
+import bcrypt from "bcrypt";
+
+export {Events, EventTypes} from "./constants";
+
+let telemetryReporter: TelemetryReporter | undefined;
```
fjakobs, line 15: Not a fan of globals. Why can't we just instantiate a TelemetryReporter in `extension.ts`?

fjakobs: I'd prefer to sub class `TelemetryReporter` and create a single instance instead.

fjakobs (changes requested):
Overall looking really good.

Main feedback:
1. check if we really need a new event
2. don't use globals for the telemetry reporter

fjakobs:
also integration tests are failing with:

```
Error: R] Could not resolve "mock-aws-s3"

    node_modules/@mapbox/node-pre-gyp/lib/util/s3_setup.js:43:28:
      43 │     const AWSMock = require('mock-aws-s3');
         ╵                             ~~~~~~~~~~~~~

  You can mark the path "mock-aws-s3" as external to exclude it from the bundle, which will remove this error. You can also surround this "require" call with a try/catch block to handle this failure at run-time instead of bundle-time.

Error: R] Could not resolve "aws-sdk"

    node_modules/@mapbox/node-pre-gyp/lib/util/s3_setup.js:76:[22](https://github.com/databricks/databricks-vscode/actions/runs/4618953599/jobs/8167050258?pr=630#step:12:23):
      76 │   const AWS = require('aws-sdk');
         ╵                       ~~~~~~~~~

  You can mark the path "aws-sdk" as external to exclude it from the bundle, which will remove this error. You can also surround this "require" call with a try/catch block to handle this failure at run-time instead of bundle-time.

Error: R] Could not resolve "nock"

    node_modules/@mapbox/node-pre-gyp/lib/util/s3_setup.js:112:[23](https://github.com/databricks/databricks-vscode/actions/runs/4618953599/jobs/8167050258?pr=630#step:12:24):
      112 │   const nock = require('nock');
          ╵                        ~~~~~~
```

mgyucht:
> also integration tests are failing with:
> 
> ```
> Error: R] Could not resolve "mock-aws-s3"
> 
>     node_modules/@mapbox/node-pre-gyp/lib/util/s3_setup.js:43:28:
>       43 │     const AWSMock = require('mock-aws-s3');
>          ╵                             ~~~~~~~~~~~~~
> 
>   You can mark the path "mock-aws-s3" as external to exclude it from the bundle, which will remove this error. You can also surround this "require" call with a try/catch block to handle this failure at run-time instead of bundle-time.
> 
> Error: R] Could not resolve "aws-sdk"
> 
>     node_modules/@mapbox/node-pre-gyp/lib/util/s3_setup.js:76:[22](https://github.com/databricks/databricks-vscode/actions/runs/4618953599/jobs/8167050258?pr=630#step:12:23):
>       76 │   const AWS = require('aws-sdk');
>          ╵                       ~~~~~~~~~
> 
>   You can mark the path "aws-sdk" as external to exclude it from the bundle, which will remove this error. You can also surround this "require" call with a try/catch block to handle this failure at run-time instead of bundle-time.
> 
> Error: R] Could not resolve "nock"
> 
>     node_modules/@mapbox/node-pre-gyp/lib/util/s3_setup.js:112:[23](https://github.com/databricks/databricks-vscode/actions/runs/4618953599/jobs/8167050258?pr=630#step:12:24):
>       112 │   const nock = require('nock');
>           ╵                        ~~~~~~
> ```

Seems to be a known issue with bcrypt: https://github.com/kelektiv/node.bcrypt.js/issues/964. I'm addressing by switching to the pure JS implementation bcryptjs as suggested. This comes with a 30% performance hit, so the username hashing will take around 17ms on average now.

packages/databricks-vscode/src/configuration/auth/AuthProvider.ts
fjakobs: +1 makes sense

PR #656, opened 2023-04-13
Treat cluster in RESIZING state as running clusters

### LINKED ISSUE #618: [BUG] Cluster state RESIZING is interpreted as not running
**Describe the bug**
When triggering a workflow while connected to an autoscaling cluster which is currently in the state "RESIZING", the cluster is considered not running. I expect the workflow to be executed while the cluster is in RESIZING state.

**To Reproduce**
1. While the connected cluster is in the state "RESIZING"
2. Click on 'Run File as Workflow on Databricks'
4. See error popup: 'The attached cluster is not running'.

**System information:**

1. Paste the output ot the `Help: About` command (CMD-Shift-P).
```
Version: 1.76.1 (user setup)
Commit: 5e805b79fcb6ba4c2d23712967df89a089da575b
Date: 2023-03-08T16:32:00.131Z
Electron: 19.1.11
Chromium: 102.0.5005.196
Node.js: 16.14.2
V8: 10.2.154.26-electron.0
OS: Windows_NT x64 10.0.19045
Sandboxed: Yes
```

3. Databricks Extension Version
`v0.3.7`

kartikgupta-db:
## Changes
* Clusters in RESIZING state should allow execution to be queued on them. They should be treated same as RUNNING clusters. This PR makes it so.
 
fixes #618

### DIFF
--- a/packages/databricks-vscode/src/cluster/ClusterListDataProvider.ts
+++ b/packages/databricks-vscode/src/cluster/ClusterListDataProvider.ts
@@ -72,13 +72,13 @@ export class ClusterListDataProvider
     public static clusterNodeToTreeItem(element: Cluster): TreeItem {
         let icon: ThemeIcon;
         switch (element.state) {
+            case "RESIZING":
             case "RUNNING":
                 icon = new ThemeIcon("debug-start");
                 break;
 
             case "RESTARTING":
             case "PENDING":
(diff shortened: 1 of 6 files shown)


fjakobs (approved):
Code looks good but tests are failing.

PR #675, opened 2023-04-25
Move autcompletion for globals to `__builtins__.py` from internal stubs.

kartikgupta-db:
## Changes
We are deprecating internal type stubs in favour of python sdk. Python sdk currently only provides typings for dbutils. This adds typings for other globals to `__builtins__.py`. 

**Removed docstrings** for these globals since they were 10+ lines each, adding a big chunk to the user's builtins file. 

* displayHTML()
* display


packages/databricks-vscode/src/language/ConfigureAutocomplete.ts
```diff
@@ -12,10 +12,26 @@ import {Loggers} from "../logger";
 import {WorkspaceStateManager} from "../vscode-objs/WorkspaceState";
 import {MsPythonExtensionWrapper} from "./MsPythonExtensionWrapper";
 import {DbConnectInstallPrompt} from "./DbConnectInstallPrompt";
+import * as os from "node:os";
```
fjakobs, line 15: can you remove this import?

fjakobs (changes requested):

fjakobs (approved):

PR #686, opened 2023-05-02
Fix multiple quote issue for .databricks.env file.

kartikgupta-db:
## Changes
* Variables being added to .databricks.env were being added with multiple quotes. This was because of the following sequence of events:  
  * Vscode would try to search for user env file (.env) and not find it. 
  * It would then try to search for mspython python.envFile and not find it. 
  * Finally it will set the userEnvFile to .env and the mspython python.envFile to .databricks/.databricks.env.
  * At the same time, if another instance of vscode is running in the same project, it will go through the same steps, but this time find mspython python.envFile to be .databricks/.databricks.env. It will use this value for userEnvFile. 
  * Now both values are set to .databricks/.databricks.env. This is an issue. 

This is fixed by
* Stripping quotes
* More strict checking for .databricks/.databricks.env appearing in userEnvFile. 

This pr also adds an integration test for running .ipynb as jobs.

### DIFF
--- a/packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
+++ b/packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
@@ -20,6 +20,7 @@ import {logging} from "@databricks/databricks-sdk";
 import {Loggers} from "../logger";
 import {Context, context} from "@databricks/databricks-sdk/dist/context";
 import {MsPythonExtensionWrapper} from "../language/MsPythonExtensionWrapper";
+import {NamedLogger} from "@databricks/databricks-sdk/dist/logging";
 
 export class DatabricksEnvFileManager implements Disposable {
     private disposables: Disposable[] = [];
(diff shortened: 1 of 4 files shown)


packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
```diff
@@ -125,6 +135,13 @@ export class DatabricksEnvFileManager implements Disposable {
         );
     }
 
+    private checkValidUserEnvPath(
```
andrewnester, line 138: Since it's a type guard function I'd suggest to rename it to `isValidUserEnvPath`

packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
```diff
@@ -263,7 +280,11 @@ export class DatabricksEnvFileManager implements Disposable {
         const data = Object.entries({
             ...databricksEnvVars,
             ...userEnvVars,
-        }).map(([key, value]) => `${key}="${value}"`);
+        }).map(([key, value]) => {
+            value = value?.startsWith('"') ? value.slice(1) : value;
+            value = value?.endsWith('"') ? value.slice(0, -1) : value;
```
andrewnester, line 285: You can simplify these 2 lines by converting it to simple regex
```
value = value.replaceAll("^\"|\"$", "");
```

packages/databricks-vscode/src/test/e2e/run_job_on_cluster_with_workspace.e2e.ts
```diff
@@ -129,6 +154,72 @@ describe("Run job on cluster with workspace", async function () {
...
+        await startSyncIfStopped();
+        await waitForSyncComplete();
+
+        const workbench = await driver.getWorkbench();
+        const editorView = workbench.getEditorView();
+        await editorView.closeAllEditors();
+
+        // open file
+        const input = await workbench.openCommandPrompt();
+        await sleep(200);
```
andrewnester, line 167: Isn't it making test flaky? What if command prompt appears after 201 ms? Is there or more reliable way to wait for certain element to appear on the screen?

PR #705, opened 2023-05-11
DB Connect Unified Auth Support

kartikgupta-db:
## Changes
* Add support for the following authentication types, for dbconnect integration
  * Azure CLI
  * Databricks U2M oauth (aws)
  * Databricks-cli profile
  
## Tests
* manually

### DIFF
--- a/packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
+++ b/packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
@@ -240,18 +240,29 @@ export class DatabricksEnvFileManager implements Disposable {
             return;
         }
         const cluster = this.connectionManager.cluster;
-        const host = this.connectionManager.databricksWorkspace?.host.authority;
-        const pat = await this.getPatToken();
         const userAgent = await this.userAgent();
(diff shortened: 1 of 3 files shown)


packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
```diff
@@ -240,18 +240,29 @@ export class DatabricksEnvFileManager implements Disposable {
...
+        const sparkEnvVars: Record<string, string> = {};
+        if (pat && host) {
+            sparkEnvVars[
+                "SPARK_REMOTE"
+            ] = `sc://${host}:443/;token=${pat};use_ssl=true;x-databricks-cluster-id=${cluster.id};user_agent=vs_code`; //;user_agent=${encodeURIComponent(userAgent)}`
+        }
+
         /* eslint-disable @typescript-eslint/naming-convention */
         return {
+            ...authEnvVars,
```
fjakobs, line 263: Shouldn't we instead just rely on the metadata service URL? I just removed all the "env" logic from the AuthProvider.

fjakobs (approved):
Once we have metadata service support in Python SDK we need to revert `toEnv`.

PR #731, opened 2023-05-25
SDK: Port error handling code from GO SDK

fjakobs:
## Changes
Port error handling code from GO SDK

### DIFF
--- a/databricks-vscode.code-workspace
+++ b/databricks-vscode.code-workspace
@@ -26,6 +26,8 @@
         },
         "[typescript]": {
             "editor.defaultFormatter": "esbenp.prettier-vscode"
-        }
+        },
+        "python.envFile": "${workspaceFolder}/.databricks/.databricks.env",
+        "databricks.python.envFile": "${workspaceFolder}/.env"
     }
 }
--- a/databricks-vscode.code-workspace
+++ b/databricks-vscode.code-workspace
@@ -26,8 +26,6 @@
         },
         "[typescript]": {
             "editor.defaultFormatter": "esbenp.prettier-vscode"
-        },
-        "python.envFile": "${workspaceFolder}/.databricks/.databricks.env",
-        "databricks.python.envFile": "${workspaceFolder}/.env"
+        }
     }
 }
--- a/packages/databricks-sdk-js/src/api-client.ts
+++ b/packages/databricks-sdk-js/src/api-client.ts
@@ -9,27 +9,12 @@ import {Context} from "./context";
 import {Headers, Config} from "./config/Config";
 import retry, {RetriableError} from "./retries/retries";
 import Time, {TimeUnits} from "./retries/Time";
+import {HttpError, parseErrorFromResponse} from "./apierr";
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const sdkVersion = require("../package.json").version;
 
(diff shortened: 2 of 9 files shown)


packages/databricks-sdk-js/src/apierr.ts
```diff
@@ -0,0 +1,141 @@
...
+                    "Please delete repos before creating new repos"
+                )
+            ) {
+                return false;
+            }
+            return true;
+        }
+        if (this.statusCode >= 400) {
+            if (this.isTransientError(this.message)) {
+                return true;
```
kartikgupta-db, line 45: I like this. Let's also do something similar for other retriable errors in sdk. Maybe we can have a common "SDKError". Adding it to this ticket https://databricks.atlassian.net/browse/DECO-709

databricks-vscode.code-workspace
```diff
@@ -26,6 +26,8 @@
         },
         "[typescript]": {
             "editor.defaultFormatter": "esbenp.prettier-vscode"
-        }
+        },
+        "python.envFile": "${workspaceFolder}/.databricks/.databricks.env",
+        "databricks.python.envFile": "${workspaceFolder}/.env"
```
kartikgupta-db, line 31:
```suggestion
        }
```

packages/databricks-sdk-js/src/api-client.ts
```diff
@@ -163,91 +152,38 @@ export class ApiClient {
...
-        }
-
         let responseJson: any;
         try {
             responseJson =
                 responseText.length === 0 ? {} : JSON.parse(responseText);
         } catch (e) {
             logAndReturnError(url, options, responseText, e, context);
-            throw new ApiClientResponseError(responseText, e);
+            throw new Error(`Can't parse reposne as JSON: ${responseText}`);
```
kartikgupta-db, line 184: We should send it through `parseError` as well.

kartikgupta-db: Let's unify this and https://github.com/databricks/databricks-vscode/pull/731/files#diff-dd9ade02b57b4295433e36501f1d12a708bd8bc95cdb04af866d851f5984691aR147-R151, 
into 1 error - IOError.

packages/databricks-sdk-js/src/apierr.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export class ApiError extends Error {
+    constructor(
+        public readonly message: string,
+        public readonly errorCode: string,
+        public readonly statusCode: number,
+        public readonly response: any
+    ) {
+        super(message);
+    }
```
kartikgupta-db, line 29: The stack trace shows a bunch of stuff before showing the actual function where the error is raised.
```
Error: 4465905737085886 unauthorized to perform ReadAction on /org_admin_conf/enableProjectTypeInWorkspace
      at parseErrorFromResponse (src\apierr.ts:5:45)
      at fn (src\api-client.ts:5:2007)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at retry (src\retries\retries.ts:2:5792)
      at ApiClient.request (src\api-client.ts:5:573)
```

One fix is, instead of returning the `ApiError` from `parseError` we instead return `ApiErrorDetails`. We then use the pattern,
`throw new ApiError(parseApiError)`. This will give a much cleaner stack trace.

kartikgupta-db (approved):

PR #733, opened 2023-05-26
SDK: Add support for Azure client secret auth

fjakobs:
## Changes
SDK: Add support for Azure client secret auth

Ported from https://github.com/databricks/databricks-sdk-go/blob/main/config/auth_azure_client_secret.go
 
## Tests
Manually tested with Azure SP credentials.


packages/databricks-sdk-js/src/AccountClient.ts
```diff
@@ -282,12 +282,6 @@ export class AccountClient {
         }
 
         this.config = config as Config;
-
-        this.config.ensureResolved();
-        if (!this.config.accountId || !this.config.isAccountClient()) {
-            throw new Error("invalid Databricks Account configuration");
-        }
```
kartikgupta-db, line 289: Why did you move it out? I don't understand the use of moving it out.

packages/databricks-sdk-js/examples/workspace.ts
```diff
@@ -1,4 +1,4 @@
-import {Config, WorkspaceClient} from "..";
+import {Config, WorkspaceClient} from "../src";
```
kartikgupta-db, line 1: Let's make examples it's own package. Easier for people to understand imports that way and easier to get started.

packages/databricks-sdk-js/src/AccountClient.ts
fjakobs: `ensureResolved` is async but since this is the constructor we can't await for this. 
This means that the check below never actually worked. I moved the check into the actual calls.

packages/databricks-sdk-js/package.json
```diff
@@ -35,6 +35,7 @@
         "ini": "^4.1.1",
         "inversify": "^6.0.1",
         "node-fetch-commonjs": "^3.2.4",
+        "openid-client": "^5.4.2",
```
nfx, line 38: why do we need it?

packages/databricks-sdk-js/src/.codegen/AccountClient.ts.tmpl
```diff
@@ -27,12 +27,6 @@ export class AccountClient {
         }
 
         this.config = config as Config;
-
-        this.config.ensureResolved();
```
nfx, line 31: why do we need to move it out of constructor?

packages/databricks-sdk-js/src/config/AzureClientSecretCredentials.ts
```diff
@@ -0,0 +1,81 @@
...
+    getAzureLoginAppId,
+} from "./Azure";
+import {
+    AuthType,
+    Config,
+    CredentialProvider,
+    RequestVisitor,
+    Headers,
+} from "./Config";
+import {Issuer, TokenSet} from "openid-client";
```
nfx, line 14: please re-implement it without external dependencies as in Go / Python / Java SDKs - it's just resolving a couple of URLs.

packages/databricks-sdk-js/src/config/AzureClientSecretCredentials.ts
```diff
@@ -0,0 +1,81 @@
...
+                    resource: [env.serviceManagementEndpoint],
+                });
+            }
+
+            if (config.azureResourceId) {
+                headers["X-Databricks-Azure-Workspace-Resource-Id"] =
+                    config.azureResourceId;
+            }
+            headers["Authorization"] = `Bearer ${innerToken.access_token}`;
+            headers["X-Databricks-Azure-SP-Management-Token"] =
```
nfx, line 62: shouldn't it get a surroudning `if`?

nfx (changes requested):

kartikgupta-db (approved):

packages/databricks-sdk-js/src/AccountClient.ts
kartikgupta-db: Makes sense. Thanks!

packages/databricks-sdk-js/src/config/AzureClientSecretCredentials.ts
fjakobs: what should be in that `if`?

packages/databricks-sdk-js/package.json
fjakobs: I removed the dependency

kartikgupta-db (approved):

PR #745, opened 2023-06-08
`cd` to directory of source file when using `%sh` magic in WSFS notebooks

### LINKED ISSUE #734: [BUG] `%pip` and `%sh pip` can't be used to install local libraries when running a job from files in workspace.
**Describe the bug**
> it ends up running in an ephemeral path which means relative paths don't seem to work properly. this is impacting both %pip install as well as %sh pip install. when doing this via Repos mode, SINGLE_USER cluster would work fine for both. USER_ISOLATION clusters however would have the issue with %pip install but not %sh pip install

kartikgupta-db:
## Changes
* %sh spawns a new shell every time, with the working directory set to an ephemeral directory in the driver. 
* This PR appends a `cd directory_of_source_file` to all %sh commands, hence getting exactly the same behaviour as files in repos. 

## Tests
* integration tests

fixes #734


packages/databricks-vscode/resources/python/notebook.workflow-wrapper.py
```diff
@@ -1,12 +1,54 @@
-#This file is autogenerated by the Databricks Extension for VS Code
+# This file is autogenerated by the Databricks Extension for VS Code
```
fjakobs, line 1:
```suggestion
# This cell is autogenerated by the Databricks Extension for VS Code
```

packages/databricks-vscode/src/test/e2e/run_job_on_cluster_with_workspace.e2e.ts
```diff
@@ -54,6 +54,8 @@ describe("Run job on cluster with workspace", async function () {
             [
                 "# Databricks notebook source",
                 `spark.sql('SELECT "hello world"').show()`,
+                "# COMMAND ----------",
+                "# MAGIC %sh ls",
```
fjakobs, line 58: I would use `pwd`

packages/databricks-vscode/src/test/e2e/run_job_on_cluster_with_repo.e2e.ts
```diff
@@ -53,6 +53,8 @@ describe("Run job on cluster with repo", async function () {
             [
                 "# Databricks notebook source",
                 `spark.sql('SELECT "hello world"').show()`,
+                "# COMMAND ----------",
+                "# MAGIC %sh ls",
```
fjakobs, line 57: Use `pwd`

packages/databricks-vscode/resources/python/notebook.workflow-wrapper.py
```diff
@@ -1,12 +1,54 @@
...
+                    "%sh\n",
+                    f"cd {src_file_dir}\n",
+                    rest.strip() + "\n",
+                    *lines[1:]
+                ]
+
+        return lines
+
+    ip = get_ipython()
+    ip.input_transformers_cleanup.append(parse_databricks_magic_lines)
```
fjakobs, line 47: 🤯

fjakobs: do we know if this has any unwanted side-effects?

packages/databricks-vscode/resources/python/notebook.workflow-wrapper.py
```diff
@@ -1,12 +1,54 @@
...
+                break
+
+        if first.startswith("%"):
+            magic = first.split(" ")[0].strip().strip("%")
+            rest = ' '.join(first.split(" ")[1:])
+
+            if magic == "sh":
+                return [
+                    "%sh\n",
+                    f"cd {src_file_dir}\n",
```
fjakobs, line 39: the dir needs to be escaped in case the path contains spaces or special characters.

fjakobs (changes requested):

packages/databricks-vscode/resources/python/notebook.workflow-wrapper.py
kartikgupta-db: Right. Good catch.

packages/databricks-vscode/resources/python/notebook.workflow-wrapper.py
kartikgupta-db: As long as none of the other cells start with a %, we are returning them untouched.

fjakobs (approved):

PR #752, opened 2023-06-13
Default values for injected widgets in notebook wrapper

kartikgupta-db:
## Changes
* Add default values for injected widgets in notebook wrapper. This enables users to use the wrapper interactively in Databricks UI, without having to copy the injected variables. 

## Tests
* unit
* integration
* manual


fjakobs:
love it

PR #754, opened 2023-06-14
Kill execution if cluster is not running or sync destination is not attached

kartikgupta-db:
## Changes
* The execution was blocked if cluster was not running and the popup to start the cluster was shown. There was no way to start a new run until the prompt was dismissed.  Similar blocking also happened for "attach sync destination" prompt. 
* This PR makes these prompts non blocking and terminates the runs when the prompts are shown. A new run can be started at any point, irrespective of prompts being dismissed or not. 

## Tests
* manually

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
         "eslint-config-prettier": "^8.8.0",
         "eslint-plugin-no-only-tests": "^3.1.0",
         "ts-mockito": "^2.6.1",
-        "typescript": "^5.1.3"
+        "typescript": "^5.0.0"
     },
     "resolutions": {
         "json5": "2.2.2"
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
         "eslint-config-prettier": "^8.8.0",
         "eslint-plugin-no-only-tests": "^3.1.0",
         "ts-mockito": "^2.6.1",
-        "typescript": "^5.1.3"
+        "typescript": "^5.0.0"
```
kartikgupta-db, line 42: Downgraded because our current eslint plugins don't support ts >5.0.0

fjakobs (approved):

PR #766, opened 2023-06-19
Add version information to `Update dbconnect` popup message

kartikgupta-db:
## Changes
image


fjakobs (approved):
Would be great to add a screenshot so I can see how it looks

PR #769, opened 2023-06-21
[DECO-1112] Allow jupyter code lenses to parse and interpret of databricks notebooks (dbnbs) 

kartikgupta-db:
## Changes
* [x] Update `jupyter.interactiveWindow.cellMarker` settings to allow vscode to interpret databricks notebooks cells as jupyter code cells https://code.visualstudio.com/docs/python/jupyter-support-py

## Tests
* [ ] unit tests
* [ ] integration tests


packages/databricks-vscode/src/language/notebooks/DatabricksNbCellLimits.ts
```diff
@@ -0,0 +1,28 @@
+import {workspaceConfigs} from "../../vscode-objs/WorkspaceConfigs";
+import {FileUtils} from "../../utils";
+import {Uri} from "vscode";
+import {ConnectionManager} from "../../configuration/ConnectionManager";
+
+export async function setDbnbCellLimits(
```
tanmay-db, line 6: we are only setting the cell limits here right? Where are we parsing?

kartikgupta-db: This enables vscode to parse the notebooks correctly. Only `# MAGIC` cells do not work right now. They will be handled in the python init script in a separate PR.

tanmay-db (approved):

PR #771, opened 2023-06-22
Seperate versioning for sdk and extension

kartikgupta-db:
## Changes
* Allow for seperate versioning of sdk and extension.
* SDK remains in preview (0.x). For simplicity, for every release we need to release both packages - sdk and extension (specify non 0 versions for both). 
If either of the 2 have 0.0.0 as a version, we just get a changelog and no release PR. (https://github.com/databricks/databricks-vscode/actions/runs/5347157952, https://github.com/databricks/databricks-vscode/actions/runs/5347160929)
Eg. PR https://github.com/databricks/databricks-vscode/pull/775

* root package versioning follows vscode versioning

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -42,7 +42,7 @@ jobs:
         run: bash -c '$([[ "${{ inputs.extension_version }}" =~ ^([0-9]+\.){2}[0-9]+ ]])'
 
       - name: Check SDK Version
-        run: bash -c '$([[ "${{ inputs.extension_version }}" =~ ^([0-9]+\.){2}[0-9]+ ]])'
+        run: bash -c '$([[ "${{ inputs.sdk_version }}" =~ ^([0-9]+\.){2}[0-9]+ ]])'
 
       - run: |
           git config --global user.name "releasebot"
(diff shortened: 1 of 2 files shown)


.github/workflows/release-pr.yml
```diff
@@ -1,88 +1,107 @@
...
-                default: 0.0.0
-                type: string
-
-    workflow_call:
-        inputs:
-            version:
-                default: 0.0.0
-                type: string
+  workflow_dispatch:
+    inputs:
```
fjakobs, line 5: indentation changed from 4 to 2 spaces, which makes review harder.

The other workflows use 4 spaces indent so maybe change this back.

.github/workflows/release-pr.yml
```diff
@@ -1,88 +1,107 @@
...
-
-            - name: Create PR
-              if: ${{ inputs.version != '0.0.0' }}
-              run: |
-                  gh pr create -B main -H ${{ steps.create-branch.outputs.branch_name }} --title "Release: v${{ inputs.version }}" --body-file ${{ steps.create-changelog.outputs.changelog_file }}
-                  rm ${{ steps.create-changelog.outputs.changelog_file }}
-              env:
-                  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  release-pr:
+    runs-on: "macos-latest"
```
fjakobs, line 30: nit: can we make it run on Linux? Faster and cheaper.

fjakobs (changes requested):

.github/workflows/release-pr.yml
kartikgupta-db: Not yet. Linux has slightly different behaviour for shell commands (eg awk), so it is hard to test. We can do it in a followup when I am more confident with the devloop.

fjakobs (approved):

PR #789, opened 2023-06-29
Block changes to config unless we are in a databricks project

kartikgupta-db:
## Changes
* We were changing env file configs when a user activated the extension by mistake. This PR fixes this by blocking changes unless we are sure that we are in a databricks project
  * Check if a `.databricks/.databricks.env` exists 
  * If not, then wait for user to connect to a workspace

## Tests
* manual


packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -530,7 +530,7 @@ export class ConnectionManager {
     }
 
     async waitForConnect(): Promise<void> {
-        if (this._state === "CONNECTING" || this._state === "DISCONNECTED") {
+        if (this._state !== "CONNECTED") {
```
tanmay-db, line 533: just curious, what other states does this cover?

tanmay-db (approved):

PR #797, opened 2023-07-05
Show errors in notebook init scripts

kartikgupta-db:
## Changes

https://github.com/databricks/databricks-vscode/assets/88345179/a7d0c37b-4770-4639-a9f8-932472bb903c

### DIFF
--- a/packages/databricks-vscode/src/language/notebooks/NotebookInitScriptManager.ts
+++ b/packages/databricks-vscode/src/language/notebooks/NotebookInitScriptManager.ts
@@ -1,0 +1,99 @@
+import {Disposable, Uri, ExtensionContext} from "vscode";
+import path from "path";
+import {PackageJsonUtils} from "../../utils";
+import {exists} from "fs-extra";
+import {mkdir, cp, rm} from "fs/promises";
+import {glob} from "glob";
+import {ConnectionManager} from "../../configuration/ConnectionManager";
+export class NotebookInitScriptManager implements Disposable {
+    private disposables: Disposable[] = [];
+    private _jupyterInitScriptVersion: string | undefined | null = null;
+
+    constructor(
+        private readonly workspacePath: Uri,
+        private readonly extensionContext: ExtensionContext,
+        private readonly connectionManager: ConnectionManager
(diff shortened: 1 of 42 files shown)


packages/databricks-vscode/src/language/notebooks/NotebookInitScriptManager.ts
```diff
@@ -63,11 +139,182 @@ export class NotebookInitScriptManager implements Disposable {
...
+                                "Open Output Window"
+                            )
+                            .then((choice) => {
+                                if (choice === "Open Output Window") {
+                                    this.outputWindow.show();
+                                }
+                            });
+                        return false;
+                    }
+                    window.showInformationMessage(
```
fjakobs, line 305: IMHO When everything is fine we should be silent

kartikgupta-db: The init scripts take a few seconds to run. That can cause confusion. That's why I also added a progress bar to show that the script is being analysed. This just gives a nice ending to the progress bar.

Users might want to fix any errors (packages and such), and try rerunning the script. No feedback would be weird then. 

Although, the extension does do a bunch of prechecks, so there should not be many missing packages. Which makes this more of an edge case right now and future proofing mechanism. I am ok with removing both notifications (progress bar and success) if the above cases have been considered.

packages/databricks-vscode/src/language/notebooks/NotebookInitScriptManager.ts
```diff
@@ -76,6 +178,216 @@ export class NotebookInitScriptManager implements Disposable {
...
+                              );
+                          if (success) {
+                              window.showInformationMessage(
+                                  "Successfully verified Databricks notebook init scripts"
+                              );
+                          }
+                          return success;
+                      }
+                  )
+                : await this.executeAndVerifyInitScripts(executable);
```
kartikgupta-db, line 383: Now withProgress is called only when we call `fromCommand` otherwise we remain silent (until error).

fjakobs (approved):

PR #810, opened 2023-07-13
Update api to the latest version

fjakobs:
## Changes
- Matches GO SDK v 0.12.0
- Some breaking changes to match Go SDK


packages/databricks-sdk-js/src/apis/workspace/model.ts
```diff
@@ -187,7 +187,13 @@ export interface DeleteSecret {
...
 }
 
-export type ExportFormat = "DBC" | "HTML" | "JUPYTER" | "R_MARKDOWN" | "SOURCE";
+export type ExportFormat =
+    | "DBC"
+    | "HTML"
+    | "JUPYTER"
+    | "R_MARKDOWN"
+    | "SOURCE"
+    | "AUTO";
```
fjakobs, line 196: @kartikgupta-db we need to make this change upstream. Do you have context why we need `AUTO`?

kartikgupta-db: AUTO allows us to upload both a notebook and a file from a common interface. We do not have to parse #Databricks notebook source comment. The API figures it out.

fjakobs: This is actually for exporting. `import` does have the `AUTO` flag but `export` doesn't. From what I can see `export` is only being used in integ tests.

kartikgupta-db (approved):

PR #811, opened 2023-07-13
bump dependency versions

fjakobs:
## Changes
This also updates prettier to the next major version, which caused some files to be re-formatted.


packages/databricks-vscode/package.json
```diff
@@ -681,53 +681,53 @@
...
-        "prettier": "^2.8.8",
+        "prettier": "^3.0.0",
         "tmp-promise": "^3.0.3",
         "ts-mockito": "^2.6.1",
         "ts-node": "^10.9.1",
-        "typescript": "~5.0.0",
+        "typescript": "~5.1.6",
         "vsce": "^2.15.0",
-        "wdio-video-reporter": "^4.0.2",
-        "wdio-vscode-service": "patch:wdio-vscode-service@npm:5.1.0#.yarn/patches/wdio-vscode-service-npm-5.1.0-dcb34bb6e9.patch",
```
kartikgupta-db, line 729: We can remove this patch now

kartikgupta-db (approved):

PR #819, opened 2023-07-25
Code completion for DABs includes

fjakobs:
## Changes
This PR watches bundle config files and dynamically applies YAML code completion to all files included by the config.

https://github.com/databricks/databricks-vscode/assets/40952/a5e862a1-5889-4a25-aa7f-0973670b03db


packages/databricks-vscode/src/bundle/GenerateBundle.ts
```diff
@@ -1,16 +1,64 @@
...
     // URI scheme for DABs JSON schemas
     const dabsUriScheme = "dabs";
 
     // URI for bundle root config json schema
-    const rootConfigSchemaUri = `${dabsUriScheme}:///root.json`;
+    const rootConfigSchemaUri = `${dabsUriScheme}:///dabs.json`;
+
+    const folder = workspace.workspaceFolders?.[0];
+    const configFilePattern = new RelativePattern(
+        folder!,
```
kartikgupta-db, line 20: Return if no folder. Failing gracefully is better.

packages/databricks-vscode/src/bundle/GenerateBundle.ts
```diff
@@ -1,16 +1,64 @@
...
+        "{databricks,bundle}.{yml,yaml}"
+    );
+
+    // file watcher on all YAML files
+    const watcher = workspace.createFileSystemWatcher("**/*.{yml,yaml}");
+    watcher.onDidChange(async () => {
```
kartikgupta-db, line 26: Let's use onDidCreate and onDidDelete

kartikgupta-db: We also need a watcher on root bundle.yaml, so that we can update globs when the root is updated. 
IMO we ONLY need the root watcher.

packages/databricks-vscode/src/bundle/GenerateBundle.ts
```diff
@@ -21,16 +69,9 @@ export async function generateBundleSchema(cli: CliWrapper) {
...
-                    "bundle.yml",
-                    "bundle.yaml",
-                ];
-                for (const name of validFileNames) {
-                    if (path.basename(resource) === name) {
-                        return rootConfigSchemaUri;
-                    }
+                const resourceUri = Uri.parse(resource);
+                if (configFiles.has(resourceUri.path)) {
+                    return rootConfigSchemaUri;
```
kartikgupta-db, line 74: Alternatively, why can't we just call the updateFileGlobs here? It is not a very expensive function (for reasonable bundles) and less compute intensive than watchers.

packages/databricks-vscode/src/bundle/GenerateBundle.ts
fjakobs: Aren't we guaranteed to always have a workspace folder? 

https://github.com/databricks/databricks-vscode/blob/main/packages/databricks-vscode/src/extension.ts#L71-L84

Still, I agree that we should make this code more defensive.

kartikgupta-db: Yep. Just do not like the look of `!`

packages/databricks-vscode/src/bundle/GenerateBundle.ts
fjakobs: We need an `onDidCreate` watcher for `*.yml` and an `onDidChange` watcher for `{databricks|bundle}.{yaml|yml}`.

That would be a bit more efficient but I feel that this code is easier to read.

We need the watcher on z`onDidCreate` in order to cover include patterns like `*.yml`.

packages/databricks-vscode/src/bundle/GenerateBundle.ts
fjakobs: This code is executed whenever a file is opened. I'd like to keep that code path as fast as possible.

packages/databricks-vscode/src/bundle/GenerateBundle.ts
fjakobs: Agreed. I'll change it

packages/databricks-vscode/src/bundle/GenerateBundle.ts
kartikgupta-db: This need not be too expensive. All we need to do is match the `resourceUri` with all the include paths in the root bundle.yaml. 

That is mostly string matching and regex matching for globs. We shouldn't even need filesystem access if we have a single watcher on root bundle.yaml keeping the list of includes updated.

kartikgupta-db:
Apparently we can have root config file including another file (a.yml) which in turn includes another file (b.yml). b.yml need not be explicitly included in the base root config. 

Need to investigate, but putting this on hold for this release.

kartikgupta-db:
Closing in favour of https://github.com/databricks/databricks-vscode/pull/892

PR #829, opened 2023-08-08
Start using databricks.yaml to save project config

kartikgupta-db:


package.json
```diff
@@ -43,5 +43,8 @@
     },
     "resolutions": {
         "json5": "2.2.2"
+    },
+    "dependencies": {
+        "yaml": "^2.3.1"
```
fjakobs, line 48: why do we need this on the top level?

packages/databricks-vscode/src/configuration/auth/MetadataService.ts
```diff
@@ -19,6 +19,8 @@ export class MetadataService implements Disposable {
 
     private magic!: string;
     private _apiClient: ApiClient | undefined;
+    private onDidChangeMagicEvent = new EventEmitter<void>();
+    public onDidChangeMagic = this.onDidChangeMagicEvent.event;
```
fjakobs, line 23: What does `magic` refer to and how is it related to `databricks.yml`?

packages/databricks-vscode/src/file-managers/ProjectConfigFile.test.ts
```diff
@@ -1,48 +1,140 @@
...
+
+        const yamlData = YAML.parse(
+            await readFile(
+                DatabricksYamlFile.getFilePath(new LocalUri(tempDir)).path,
+                {encoding: "utf-8"}
+            )
+        );
+
+        assert.deepEqual(yamlData, {
+            environments: {
```
fjakobs, line 62: Why are we using environments? Shouldn't we just overwrite the top-level keys?

package.json
kartikgupta-db: Nope. I will move it.

packages/databricks-vscode/src/configuration/auth/MetadataService.ts
kartikgupta-db: The metdataservice has a nonce (`http:localhost:port/nonce`). We need to rewrite env vars whenever this nonce is regenerated. The env vars allow bundle to pickup auth from metadata-service.

packages/databricks-vscode/src/file-managers/ProjectConfigFile.test.ts
kartikgupta-db: Top level keys might be present in a user defined bundle.yaml. The only way to safely override them (while allowing users the ability to target other envs from terminal) is to use a vscode specific env to override the configs. This is the only env users can edit through the UI and is never committed to git.

packages/databricks-vscode/src/file-managers/DatabricksYamlFile.ts
```diff
@@ -0,0 +1,99 @@
...
+            mode: "development",
+        };
+        /* eslint-enable @typescript-eslint/naming-convention */
+
+        const doc = new YAML.Document({environments});
+        return doc.toString();
+    }
+
+    static getTargetName(workspaceState: WorkspaceStateManager) {
+        return `databricks-vscode-${workspaceState.fixedUUID.slice(0, 8)}`;
```
fjakobs, line 48: what is the consideration between adding a new environment or simply overriding values in the root of the config?

packages/databricks-vscode/src/file-managers/DatabricksYamlFile.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private static fromYaml(
+        yaml: string,
+        workspaceState: WorkspaceStateManager
+    ) {
+        const parsedConfig: DatabricksYaml = YAML.parse(yaml);
+        const target =
```
fjakobs, line 56: we should validate the schema here and make sure all required fields are defined.

packages/databricks-vscode/src/file-managers/ProjectConfigFile.test.ts
fjakobs: but it also means that users can't easily use it from the command line. Unless we make the VSCode env the default.

fjakobs (commented):
Let's discuss how we want to override environments. Otherwise, the code is looking good.

packages/databricks-vscode/src/file-managers/DatabricksYamlFile.ts
kartikgupta-db: There are no required fields. This is completely dynamic and users can have a broken env (eg with no host). They have to select proper configs from the UI.

PR #834, opened 2023-08-16
Support for %run magics and # MAGIC commands

kartikgupta-db:


packages/databricks-vscode/resources/python/00-databricks-init.py
```diff
@@ -131,10 +128,77 @@ def fs(self, line: str, local_ns):
...
+            return "Databricks notebook source" in f.readline()
+
+def strip_hash_magic(lines: List[str]) -> List[str]:
+    if len(lines) == 0:
+        return lines
+    if lines[0].startswith("# MAGIC"):
+        return [line.partition("# MAGIC")[2] for line in lines]
+    return lines
+
+def convert_databricks_notebook_to_ipynb(py_file: str):
```
fjakobs, line 143: this code is becoming pretty complex. Let's think about how we can add some unit tests.

fjakobs (commented):
could you add a description and ideally a small video how this behaves and works?

fjakobs (approved):

PR #852, opened 2023-08-31
Move sdk/services from the SDK to VS Code

fjakobs:
## Changes
 - Move sdk/services from the SDK to VS Code
 - Get rid of most `@databircks/databricks-sdk/dist/...` imports. Only `context` is still there, which can be cleaned up after the next SDK release.
 
 ## Testing
  - [x] make sure integ tests are run
  - [x] make sure unit tests are run


kartikgupta-db (approved):
Nit: Can we name the new folder, sdk-wrappers or sdk-extensions or something like that?

mgyucht (approved):
Makes sense, though maybe the name of the directory `sdk` is overloaded, what about `sdk-extensions` or something like that?

fjakobs:
I did the rename

PR #892, opened 2023-10-06
Bundle watcher

kartikgupta-db:
## Changes
* Autogenerate BundleSchema.d.ts file (from `databricks bundle schema` command) to provide typing for the parsed bundle object.
* Add a watcher for all DABs configuration files
* Add BundleFileSet object, to handle operations on the full bundle file tree. 
* Used these to add autocompletion for all files included in the root bundle.yaml.

**NOTE:** This PR currently parses yamls in code. We want to move that to databricks CLI. The implementation can is modular and can be swapped with the CLI calls once we have those features in the CLI. 

## Tests
* Unit tests
* Manual


packages/databricks-vscode/src/bundle/bundleAutocompleteProvider.ts
```diff
@@ -18,21 +24,22 @@ export async function generateBundleSchema(cli: CliWrapper) {
...
-                    }
+                if (
+                    bundleFileList.find(
+                        (i) => i.fsPath === Uri.parse(resource).fsPath
+                    ) !== undefined
+                ) {
+                    return rootConfigSchemaUri;
                 }
-                return undefined;
             },
```
pietern, line 43: @kartikgupta-db Is this registration updated as new bundle configuration files are added, or does this only trigger on extension reload? Not clear to me when this contributor callback is called.

packages/databricks-vscode/src/bundle/bundleAutocompleteProvider.ts
```diff
@@ -18,21 +24,22 @@ export async function generateBundleSchema(cli: CliWrapper) {
...
-                    "bundle.yml",
-                    "bundle.yaml",
-                ];
-                for (const name of validFileNames) {
-                    if (path.basename(resource) === name) {
-                        return rootConfigSchemaUri;
-                    }
+                if (
+                    bundleFileList.find(
+                        (i) => i.fsPath === Uri.parse(resource).fsPath
```
pietern, line 38: This can call `Uri.parse` outside of the lambda to save some cycles.

packages/databricks-vscode/src/bundle/BundleFileSet.ts
```diff
@@ -0,0 +1,133 @@
...
+            return (
+                await glob.glob(
+                    toGlobPath(
+                        path.join(this.workspaceRoot.fsPath, includedFilesGlob)
+                    ),
+                    {nocase: process.platform === "win32"}
+                )
+            ).map((i) => Uri.file(i));
+        }
+    }
```
pietern, line 79: Checking: this will soon use the `bundle validate` equivalent output to determine the same, correct?

packages/databricks-vscode/src/bundle/bundleAutocompleteProvider.ts
kartikgupta-db: Updated here as well https://github.com/databricks/databricks-vscode/pull/922

PR #913, opened 2023-10-23
Refactor `StateStore` to make keys more explicit at the point of use. 

kartikgupta-db:
## Changes
Advantages of this approach
* Adding new variables does not require creating an explicit setter and getter. Default setters and getters should work out of the box.
* Easier to see all the information in 1 location. 
* The keys are immutably exposed to point of use. It allows for potential programatic access to these functions. 
* Setters require calling the `set` function, which is async. This prevents us from having to create even more functions just to make setting a variable `await`able.

### DIFF
--- a/packages/databricks-vscode/src/vscode-objs/StateStorage.ts
+++ b/packages/databricks-vscode/src/vscode-objs/StateStorage.ts
@@ -68,6 +68,11 @@ const Keys = {
             return currentEnvs;
         },
     }),
+
+    "databricks.lastInstalledExtensionVersion": withType<string>()({
+        location: "workspace",
+        defaultValue: "0.0.0",
+    }),
 };
 
 type ValueType<K extends keyof typeof Keys> = (typeof Keys)[K]["_type"];
--- a/packages/databricks-vscode/src/vscode-objs/StateStorage.ts
+++ b/packages/databricks-vscode/src/vscode-objs/StateStorage.ts
@@ -1,73 +1,95 @@
 import {randomUUID} from "crypto";
 import {ExtensionContext} from "vscode";
 
+/* eslint-disable @typescript-eslint/naming-convention */
+type KeyInfo<V> = {
+    location: "global" | "workspace";
+    defaultValue?: V;
+    getter?: (storage: StateStorage, value: V | undefined) => V | undefined;
+    setter?: (storage: StateStorage, value: V | undefined) => V | undefined;
+};
+
+function withType<V>() {
+    return function <D extends KeyInfo<V>>(data: D) {
+        return data as typeof data & {_type: V};
+    };
+}
+
+const Keys = {
+    "databricks.clusterId": withType<string>()({
+        location: "workspace",
+    }),
+
+    "databricks.wsfs.skipSwitchToWorkspace": withType<boolean>()({
+        location: "workspace",
+        defaultValue: false,
+    }),
+
+    "databricks.autocompletion.skipConfigure": withType<boolean>()({
+        location: "workspace",
+        defaultValue: false,
+    }),
+
+    "databricks.fixedRandom": withType<number>()({
+        location: "global",
+        getter: (storage, value) => {
+            if (value === undefined) {
+                value = Math.random();
+                storage.set("databricks.fixedRandom", value);
+            }
+            return value;
+        },
+    }),
+
+    "databricks.fixedUUID": withType<string>()({
+        location: "workspace",
+        getter: (storage, value) => {
+            if (value === undefined) {
+                value = randomUUID();
+                storage.set("databricks.fixedUUID", value);
+            }
+            return value;
+        },
+    }),
+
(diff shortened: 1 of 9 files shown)


packages/databricks-vscode/src/vscode-objs/StateStorage.ts
```diff
@@ -1,73 +1,100 @@
...
-            randomNum = Math.random();
-            this.context.globalState.update(
-                "databricks.fixedRandom",
-                randomNum
-            );
-        }
-        return randomNum;
-    }
+function withType<V>() {
+    return function <D extends KeyInfo<V>>(data: D) {
```
ilia-db, line 13: Why nested functions? Can outer function already return data as `KeyInfoWithType<V>`? Later in the `storage.get` you won't need to use that `as` anymore.

packages/databricks-vscode/src/vscode-objs/StateStorage.ts
```diff
@@ -1,73 +1,100 @@
...
-        );
-    }
+    "databricks.lastInstalledExtensionVersion": withType<string>()({
+        location: "workspace",
+        defaultValue: "0.0.0",
+    }),
+};
+
+type ValueType<K extends keyof typeof Keys> = (typeof Keys)[K]["_type"];
+type GetterReturnType<D extends KeyInfo<any>> = D extends {getter: infer G}
```
ilia-db, line 79: Does it make sense for a getter type to be different from ValueType?

packages/databricks-vscode/src/vscode-objs/StateStorage.ts
```diff
@@ -1,73 +1,100 @@
...
+        ? ReturnType<G>
+        : undefined
+    : undefined;
+
+type DefaultValue<K extends keyof typeof Keys> =
+    "defaultValue" extends keyof (typeof Keys)[K]
+        ? ValueType<K>
+        : GetterReturnType<(typeof Keys)[K]>;
+
+type KeyInfoWithType<V> = KeyInfo<V> & {
```
ilia-db, line 90: What are the benefits of keeping it separate from the KeyInfo? Can those additional types just be defined on the KeyInfo, for the simplicity sake?

kartikgupta-db: KeyInfo holds actual concrete data. The type fields are only types. Defining them on keyinfo means someone can try setting them when adding a new fields to keys.

packages/databricks-vscode/src/vscode-objs/StateStorage.ts
```diff
@@ -1,73 +1,100 @@
...
+        location: "workspace",
+        defaultValue: "0.0.0",
+    }),
+};
+
+type ValueType<K extends keyof typeof Keys> = (typeof Keys)[K]["_type"];
+type GetterReturnType<D extends KeyInfo<any>> = D extends {getter: infer G}
+    ? G extends (...args: any[]) => any
+        ? ReturnType<G>
+        : undefined
```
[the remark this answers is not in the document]
kartikgupta-db: We want the final return type of the get function to be undefined if both getter and default value are undefined.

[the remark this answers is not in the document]
ilia-db: Yeah, but that's the `undefined` on the line below. This `undefined` will happen when a default value is undefined and a non-undefined getter field is not a function (e.g. doesn't match `G extends (...args: any[]) => any`) - that seems as something that should never happen. Might be misunderstanding something though

packages/databricks-vscode/src/vscode-objs/StateStorage.ts
kartikgupta-db: Getters can return `ValueType | undefined` or just `ValueType`. I wanted to account for both the cases, and have the autocompletion NOT complete return as undefined if the getter always returns a value or if the defaultValue is defined. 

This is why we also require the nested withType function. This allows us to have getters which extend the standard getter type in the KeyInfo and potentially NOT return undefined. If we don't do this, we will have to type data as any (makes defining new keys type unsafe) or we will lose out on this information about getters.

packages/databricks-vscode/src/vscode-objs/StateStorage.ts
kartikgupta-db: This is related to how I wanted getters to work. Replied [here](https://github.com/databricks/databricks-vscode/pull/913#discussion_r1368593091)

ilia-db (approved):

packages/databricks-vscode/src/vscode-objs/StateStorage.ts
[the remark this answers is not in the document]
kartikgupta-db: KeyInfo shape should prevent this from happening, so this line should be a noop either way. Semantically never does make more sense. I will fix it in another PR.

PR #922, opened 2023-10-30
Make DABs autocomplete detect new bundle files

kartikgupta-db:
## Changes
* Fixes a big in https://github.com/databricks/databricks-vscode/pull/892, which caused the bundle autocomplete to not recognise a new file. 
* Bundle autocomplete now updates monitored files when
  * there is a change to the bundle root
  * a bundle file is created
  * a bundle file is deleted

## Tests
* manual

### DIFF
--- a/packages/databricks-vscode/src/bundle/bundleAutocompleteProvider.ts
+++ b/packages/databricks-vscode/src/bundle/bundleAutocompleteProvider.ts
@@ -39,9 +39,10 @@ export async function registerBundleAutocompleteProvider(
         redHatYamlSchemaApi.registerContributor(
             "dabs",
             (resource: string) => {
+                const resourceUri = Uri.parse(resource);
                 if (
                     bundleFileList.find(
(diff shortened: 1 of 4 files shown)


pietern (approved):

packages/databricks-vscode/src/bundle/bundleAutocompleteProvider.ts
```diff
@@ -28,14 +28,21 @@ export async function registerBundleAutocompleteProvider(
         context.subscriptions.push(
             bundleWatcher.onDidChangeRootFile(async () => {
                 bundleFileList = await bundleFileSet.allFiles();
+            }),
+            bundleWatcher.onDidCreate(async (e) => {
+                bundleFileList.push(e);
+            }),
+            bundleWatcher.onDidDelete(async (e) => {
+                bundleFileList.push(e);
```
pietern, line 36: Shouldn't we remove the file from this list on delete?

I might be misunderstanding how this works.

PR #923, opened 2023-10-31
Use databricks CLI log-file option to capture the logs

ilia-db:
## Changes
The CLI can write its logs to a file and we don't have to manage this on the extension side. 

- The format of the logs is a bit different now
- In the verbose mode we were writing debug logs to the terminal task in the VSCode (as they were coming to the stderr). Right now such logs are only in the log file, and output has just the stdout (as CLI keeps stderr empty when you ask it to log to a file)
- The verbose mode option is gone. The output in the sync terminal is clean and always based on stdout, and the verbose logs are always written into a file

## Tests
Updated existing tests

### DIFF
--- a/packages/databricks-vscode/src/cli/CliWrapper.test.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.test.ts
@@ -12,6 +12,7 @@ import {writeFile} from "node:fs/promises";
 
 import {CliWrapper} from "./CliWrapper";
 import path from "node:path";
+import os from "node:os";
 import {Context} from "@databricks/databricks-sdk/dist/context";
 import {logging} from "@databricks/databricks-sdk";
 
@@ -25,10 +26,12 @@ describe(__filename, () => {
     });
 
     it("should create sync commands", async () => {
+        const logUri = Uri.file(os.tmpdir());
         const cli = new CliWrapper({
             asAbsolutePath(path: string) {
                 return path;
             },
+            logUri,
         } as any);
         const mapper = new SyncDestinationMapper(
             new LocalUri(
@@ -45,13 +48,20 @@ describe(__filename, () => {
         let {command, args} = cli.getSyncCommand(mapper, "incremental");
         assert.equal(
             [command, ...args].join(" "),
-            "./bin/databricks sync . /Repos/fabian.jakobs@databricks.com/notebook-best-practices --watch --output json"
+            [
+                "./bin/databricks sync . /Repos/fabian.jakobs@databricks.com/notebook-best-practices --watch --output json",
+                `--log-level error --log-file ${logUri.fsPath}/databricks-cli-logs.json --log-format json`,
+            ].join(" ")
         );
 
         ({command, args} = cli.getSyncCommand(mapper, "full"));
         assert.equal(
             [command, ...args].join(" "),
-            "./bin/databricks sync . /Repos/fabian.jakobs@databricks.com/notebook-best-practices --watch --output json --full"
+            [
+                "./bin/databricks sync . /Repos/fabian.jakobs@databricks.com/notebook-best-practices --watch --output json",
+                `--log-level error --log-file ${logUri.fsPath}/databricks-cli-logs.json --log-format json`,
+                "--full",
+            ].join(" ")
         );
     });
 
--- a/packages/databricks-vscode/src/cli/CliWrapper.test.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.test.ts
@@ -19,6 +19,7 @@ import {Context} from "@databricks/databricks-sdk/dist/context";
 import {logging} from "@databricks/databricks-sdk";
 
 const execFile = promisify(execFileCb);
+const cliPath = path.join(__dirname, "../../bin/databricks");
 
 function getTempLogFilePath() {
     return path.join(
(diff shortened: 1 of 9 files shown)


packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -47,6 +47,24 @@ export class CliWrapper {
...
     }
 
+    get loggingArguments(): string[] {
+        if (!workspaceConfigs.loggingEnabled) {
+            return [];
+        }
+        const logFileUri = Uri.joinPath(
+            this.extensionContext.logUri,
+            "databricks-cli-logs.json"
+        );
```
kartikgupta-db, line 57: Let's fetch this from the Logger module or the LoggerManager class. We want to hold this file and the sdk-extension log file in the same place.

packages/databricks-vscode/src/cli/DatabricksCliSyncParser.ts
```diff
@@ -96,28 +89,11 @@ export class DatabricksCliSyncParser {
...
-                    databricksLogLevelToSdk.get(typeMatch[1]) ??
-                    currentLogLevel;
-            }
-            logging.NamedLogger.getOrCreate(Loggers.CLI).log(
-                currentLogLevel,
-                line,
-                {
-                    outfile: "stderr",
-                }
-            );
```
kartikgupta-db, line 120: This makes me super happy :))))))

packages/databricks-vscode/src/cli/CliWrapper.test.ts
```diff
@@ -5,62 +5,100 @@ import {
...
+import {when, spy, reset} from "ts-mockito";
 import {CliWrapper} from "./CliWrapper";
 import path from "node:path";
+import os from "node:os";
+import crypto from "node:crypto";
 import {Context} from "@databricks/databricks-sdk/dist/context";
 import {logging} from "@databricks/databricks-sdk";
 
 const execFile = promisify(execFileCb);
+const cliPath = path.join(__dirname, "../../bin/databricks");
```
kartikgupta-db, line 22:
```suggestion
const cliPath = path.join(__dirname, "..", "..", "bin", "databricks");
```

kartikgupta-db: To handle windows.

packages/databricks-vscode/src/cli/CliWrapper.test.ts
```diff
@@ -5,62 +5,100 @@ import {
...
+        os.tmpdir(),
+        `databricks-cli-logs-${crypto.randomUUID()}.json`
+    );
+}
+
+function createCliWrapper(logFilePath?: string) {
+    return new CliWrapper(
+        {
+            asAbsolutePath(relativePath: string) {
+                return path.join(__dirname, "../..", relativePath);
```
kartikgupta-db, line 35:
```suggestion
                return path.join(__dirname, "..", "..", relativePath);
```

kartikgupta-db (commented):
LGTM. Thanks! Just some nits. 

We also need to inform @PaulCornellDB to get our docs updated to say we do not support this configuration going forward.

packages/databricks-vscode/src/cli/CliWrapper.test.ts
ilia-db: [`path.join` normalizes the path](https://nodejs.org/api/path.html#pathjoinpaths), so there's no need to keep all the parts separate. It doesn't really hurt, but I think it's more readable with the fewer amount of arguments

kartikgupta-db: Aah. Nice. Didn't know that. Thanks!

kartikgupta-db (approved):

packages/databricks-vscode/src/cli/CliWrapper.test.ts
```diff
@@ -5,62 +5,100 @@ import {
...
                 Uri.from({
                     scheme: "wsfs",
-                    path: "/Repos/fabian.jakobs@databricks.com/notebook-best-practices",
+                    path: "/Repos/user@databricks.com/project",
                 })
             )
         );
 
+        const syncCommand = `${cliPath} sync . /Repos/user@databricks.com/project --watch --output json`;
+        const loggingArgs = `--log-level debug --log-file ${logFilePath} --log-format json`;
```
pietern, line 80: You can set these via environment variables. Then you don't need to worry about paths with spaces, etc.

Here's the list: https://github.com/databricks/cli/blob/d70d7445c4f8f873959dd45ae68f8c42617d9353/cmd/root/logger.go#L17

ilia-db: I went with the arguments for the consistency sake, as we use them for all other CLI options (including paths for sync destination, which can also contain spaces). As I understand paths with spaces shouldn't be a concern, as nodejs handles normalization for us, plus we don't spawn the process in a shell, so normalising shouldn't be necessary. I haven't tested this on Windows though, will try to double check

pietern: Understood, I didn't see the array in `loggingArguments` and thought the args were built through string interpolation. Would be great to double check that everything works on Windows.

ilia-db: Checked on windows, no problems for paths with spaces
Screenshot 2023-11-03 at 11 59 14

PR #924, opened 2023-10-31
Use bundle for configuration and not project.json

kartikgupta-db:
## Changes
* Select bundle target from configuration menu. If not selected, a default target will be selected (based on mode and `default=true` flags). 
* Load the configuration from the bundle for the selected target.
* Users can override 
  * **auth type**:  triggers login flow with the selected auth type.
  * **cluster**: triggers cluster selector is logged in
  *  workspace folder

This does NOT use bundle to sync and run code. This is still handled by `databricks sync` and custom logic for running.


packages/databricks-vscode/src/extension.ts
```diff
@@ -585,6 +588,15 @@ export async function activate(
             );
         });
 
+    context.subscriptions.push(
+        commands.registerCommand("databricks.internal.clearOverrides", () => {
+            stateStorage.set("databricks.bundle.overrides", undefined);
+        })
+    );
+
```
kartikgupta-db, line 596: This is intended for resetting all the overrides for debugging. I intend to make it into a button in near future so users can reset all overrides.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -98,125 +153,53 @@ export class ConnectionManager {
...
-            NamedLogger.getOrCreate("Extension").error(message, e);
-            window.showErrorMessage(message);
 
-            this.updateState("DISCONNECTED");
-            await this.logout();
+        if (!(await authProvider.check())) {
+            // We return without any state changes. This ensures that
+            // if users move from a working auth type to an invalid
+            // auth, the old auth will continue working and they will not
+            // have to re authenticate even the old one.
```
ilia-db, line 171: Will they get a notification about incorrect auth? 

I might be misunderstanding this comment, but if I change auth params to some incorrect values, I would expect the extension to stop working. Or at least notify me about the problem.

Should we throw an error here instead of silently returning, so the config wizard can show an error notification? (`init` login can also be wrapped)

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -256,227 +239,81 @@ export class ConnectionManager {
             await this.waitForConnect();
         }
 
-        this._projectConfigFile = undefined;
         this._workspaceClient = undefined;
         this._databricksWorkspace = undefined;
-        this.updateCluster(undefined);
-        this.updateSyncDestination(undefined);
+        await this.setClusterManager();
+        await this.setSyncDestinationMapper();
```
ilia-db, line 245: Nitpick, but `update*` prefix for these methods now makes more sense, since they don't accept any arguments and so we aren't explicitly setting anything

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -60,10 +51,74 @@ export class ConnectionManager {
 
     public metadataServiceUrl?: string;
 
+    private async setSyncDestinationMapper() {
+        const workspacePath = await this.configModel.get("workspaceFsPath");
+        const remoteUri = workspacePath
```
ilia-db, line 56: I noticed that in the `login` method we are also validating the remote uri. Does it make sense to do it here too, since this logic can be triggered outside of the login flow?

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -98,125 +153,53 @@ export class ConnectionManager {
...
+            this.detachSyncDestination();
         }
 
-        await this.createWsFsRootDirectory(workspaceClient);
+        await this.createWsFsRootDirectory(this._workspaceClient);
+        await this.setSyncDestinationMapper();
+        await this.setClusterManager();
+        await this.configModel.set("authParams", authProvider.toJSON());
+
         this.updateState("CONNECTED");
```
ilia-db, line 203: If something fails before we reach this point (but after setting CONNECTING state) we might end up in a bad state (with hanging `waitForConnect` promises, not working `logout` method, and the UI not being updated properly). Was there a reason to remove the try catch fail-safe we had around `_login` before?

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -256,227 +239,81 @@ export class ConnectionManager {
...
-                        `Error while fetching permission for cluster ${
-                            (cluster as Cluster).id
-                        }`,
-                        e
-                    );
-                });
-
-            this.updateCluster(cluster);
-        } catch (e) {
+            this.login(config.authProvider);
```
ilia-db, line 262: Missing `await`?

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -256,227 +239,81 @@ export class ConnectionManager {
...
-        if (!this.cluster && this._projectConfigFile?.clusterId === undefined) {
+    @onError({
+        popup: {prefix: "Can't attach cluster: "},
+        log: {logger: Loggers.Extension},
+    })
+    async attachCluster(clusterId: string): Promise<void> {
+        if (this.cluster?.id === clusterId) {
             return;
         }
+        this.configModel.set("clusterId", clusterId);
```
ilia-db, line 285: Is `set` here returns a promise? We should return it or await, otherwise the error will be swallowed

Same for the detachCluster

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -256,227 +239,81 @@ export class ConnectionManager {
...
-        } catch (e) {
-            NamedLogger.getOrCreate("Extension").error(
-                "Attach Sync Destination error",
-                e
-            );
+    @onError({
+        popup: {prefix: "Can't attach sync destination: "},
+        log: {logger: Loggers.Extension},
+    })
+    async attachSyncDestination(remoteWorkspace: RemoteUri): Promise<void> {
```
ilia-db, line 300: Same as above

packages/databricks-vscode/src/extension.ts
```diff
@@ -89,6 +91,7 @@ export async function activate(
         return undefined;
     }
 
+    const workspaceUri = workspace.workspaceFolders[0].uri;
```
ilia-db, line 94: What if I add several folders to the workspace, all containing databricks bundle configs? Not sure how common that might be, but I personally like to add several folders to the same workspace instead of having multiple VSCode windows open.

I can imagine our side panel you will also need a project selector at the top or something. I wonder if this is something we even want to support in the first iteration.

packages/databricks-vscode/src/utils/onErrorDecorator.ts
```diff
@@ -0,0 +1,67 @@
...
+interface Props {
+    popup?:
+        | {
+              prefix?: string;
+          }
+        | true;
+    log?: {
+        logger: string;
+        prefix?: string;
+    };
```
ilia-db, line 17: Do we expect to supply different loggers here? Can we simplify for now and always use `Loggers.Extension`? We can always add the log parameter later if we need it

packages/databricks-vscode/package.json
```diff
@@ -211,9 +211,13 @@
             {
                 "command": "databricks.call",
```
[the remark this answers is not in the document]
kartikgupta-db: Aah, I wasn't aware that you can just do it programatically. In that case this is exactly what we should do. Can you make a PR for that?

packages/databricks-vscode/src/cluster/ClusterManager.ts
```diff
@@ -16,8 +16,13 @@ export class ClusterManager implements Disposable {
 
     private setInterval() {
         this.refreshTimer = setInterval(async () => {
+            const oldState = this.cluster.state;
             await this.cluster.refresh();
-            this.onChange(this.cluster.state);
+            if (
+                JSON.stringify(oldState) !== JSON.stringify(this.cluster.state)
```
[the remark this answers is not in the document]
kartikgupta-db: Yep.

packages/databricks-vscode/src/cluster/ClusterManager.ts
```diff
@@ -16,8 +16,13 @@ export class ClusterManager implements Disposable {
 
     private setInterval() {
         this.refreshTimer = setInterval(async () => {
+            const oldState = this.cluster.state;
             await this.cluster.refresh();
```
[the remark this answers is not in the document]
kartikgupta-db: The setInterval should not wait for the promise to resolve (from what I understand). So it's event loop should continue sending the requests at the correct interval.

packages/databricks-vscode/src/configuration/ConfigurationDataProvider.test.ts
```diff
@@ -122,16 +133,17 @@ describe(__filename, () => {
...
-            instance(mock(Telemetry))
+            instance(mock(Telemetry)),
+            mockConfigModel
         );
         disposables.push(provider);
 
         const children = await resolveProviderResult(provider.getChildren());
         assert(children);
-        assert.equal(children.length, 0);
+        assert.equal(children.length, 4);
```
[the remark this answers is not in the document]
kartikgupta-db: Yep. I will do that in an even later PR. This UI has completely changed now, so the tests are no longer relevant. I fixed this here to make sure no unintentional regression happened.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
kartikgupta-db: The errors and retries are handled by each of the "check" functions. https://github.com/databricks/databricks-vscode/blob/bundle-integ/packages/databricks-vscode/src/configuration/auth/AuthProvider.ts#L53-L57

packages/databricks-vscode/src/configuration/ConnectionManager.ts
kartikgupta-db: I missed adding that to the `configureWorkspace` method later. The idea was that login is supposed to login and nothing else and fail in a weird state if that happens (because it never did reach any other state). And then the calling functions handle retry or logout.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
kartikgupta-db: missed it.

packages/databricks-vscode/src/extension.ts
kartikgupta-db: Right now, no. We do not support multiple workspaces. We might look into it later. The option is mostly open to us in this design.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -60,10 +51,74 @@ export class ConnectionManager {
...
+                this
+            ),
+            this.configModel.onDidChange(
+                "clusterId",
+                this.setClusterManager,
+                this
+            )
+            // We DO NOT react to auth parameter changes, because ideally all
+            // auth parameter changes MUST pass through this class and are only
+            // set after a successful login. We do not want to relogin.
```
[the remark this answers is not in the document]
kartikgupta-db: That is a good point. We should bypass configModel and directly listen to bundle changes here.

packages/databricks-vscode/src/configuration/ConfigurationDataProvider.ts
```diff
@@ -82,28 +87,65 @@ export class ConfigurationDataProvider
         element?: ConfigurationTreeItem | undefined
     ): Promise<Array<ConfigurationTreeItem>> {
         switch (this.connectionManager.state) {
+            case "DISCONNECTED":
             case "CONNECTED":
                 break;
             case "CONNECTING":
                 await this.connectionManager.waitForConnect();
```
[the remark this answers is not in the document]
kartikgupta-db: We do want to wait here. This allows the UI to show a small loading indicator on top of the configuration view indicating that something is happening. We already have a task to add progress bar for loading. Maybe we can take it out then.

packages/databricks-vscode/src/utils/onErrorDecorator.ts
kartikgupta-db: Updated

packages/databricks-vscode/src/configuration/ConnectionManager.ts
kartikgupta-db: Yep. Added.

packages/databricks-vscode/src/cluster/ClusterManager.ts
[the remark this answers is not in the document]
ilia-db: > The setInterval should not wait for the promise to resolve
But that's exactly what I'm saying - it's possible that we will have one long refresh call (spanning longer that 10s due to network glitch), then set interval will initiate another one, which will resolve successfully, then the old one will resolve with a failure (or bad cluster state, which is already outdated) and we will end up in a detached state. 
It's a really artificial edge case though

packages/databricks-vscode/src/configuration/ConfigurationDataProvider.ts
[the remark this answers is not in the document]
ilia-db: Returning nodes with connecting spinners in the CONNECTED state is more obvious loading indicator, but that's subjective. We can handle it in the existing task indeed

packages/databricks-vscode/src/extension.ts
ilia-db: The only concern is how much code we will have to re-write. I still don't have a full grasp on all the modules, but it doesn't seem that big of a deal. I guess data providers will change the most, but the rest of the modules will just be instantiated multiple times. Do we have some "global" singleton dependencies that everyone depend on?

kartikgupta-db: We do not have any global singleton dependencies except a `WorkspaceConfigs` object which handles vscode configs. It can remain global since it doesn't have any state.

packages/databricks-vscode/src/cluster/ClusterManager.ts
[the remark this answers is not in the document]
kartikgupta-db: Right. We can queue these requests. But I would be ok with having momentary incorrect data but eventual consistency.

packages/databricks-vscode/src/utils/onErrorDecorator.ts
```diff
@@ -0,0 +1,75 @@
...
+    return function showErrorDecorator(
+        target: any,
+        propertyKey: string,
+        descriptor: PropertyDescriptor
+    ) {
+        let contextParamIndex: number = -1;
+        // Find the @context if it exists. We want to use it for logging whenever possible.
+        if (props.log !== undefined) {
+            try {
+                logging.withLogContext(Loggers.Extension)(
```
ilia-db, line 31: Seeing go-like contexts in js/ts is weird. Do we ever use them for something other than loggers? Just a random question :-)

kartikgupta-db: We use them for passing down cancellation tokens to sdk operations. We had much bigger plans for these, but never went anywhere.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -98,126 +172,66 @@ export class ConnectionManager {
...
-                    const host = await workspaceClient.apiClient.host;
-                    await env.openExternal(
-                        Uri.parse(
-                            host.toString() +
-                                "#setting/accounts/workspace-settings"
-                        )
-                    );
-                }
             }
+            await this.logout();
```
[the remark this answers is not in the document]
kartikgupta-db: Yep. Good catch again. I have removed the dependence on waitForConnect and, infact, state in logout. 

Now there is essentially 1 login/logout mutex. That makes login and logout mutually exclusive.

ilia-db (approved):

PR #925, opened 2023-11-02
Make configuration wizard sticky

ilia-db:
## Changes
Add `ignoreFocusOut` to both steps of the configuration wizard. They will not disappear when focusing away from them, but can still be closed with an escape button.

## Tests
Manual


kartikgupta-db (approved):
Let's also create tickets for other quickpicks?

PR #926, opened 2023-11-02
Inject databricks globals into python debugger

kartikgupta-db:
## Changes
* Consolidate databricks run modes into a separate menu. 
image

* Inject globals when running with debugger or running the file locally.

### DIFF
--- a/packages/databricks-vscode/src/extension.ts
+++ b/packages/databricks-vscode/src/extension.ts
@@ -5,6 +5,10 @@ import {
     extensions,
     window,
     workspace,
+    DebugConfigurationProvider,
+    DebugConfiguration,
+    ProviderResult,
+    WorkspaceFolder,
 } from "vscode";
 import {CliWrapper} from "./cli/CliWrapper";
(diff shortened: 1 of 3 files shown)


packages/databricks-vscode/src/extension.ts
```diff
@@ -574,14 +578,51 @@ export async function activate(
...
+    //             request: "launch",
+    //         },
+    //     ];
+    // }
+
+    resolveDebugConfigurationWithSubstitutedVariables(
+        folder: WorkspaceFolder | undefined,
+        debugConfiguration: DebugConfiguration
+    ): ProviderResult<DebugConfiguration> {
+        if (debugConfiguration.type !== "python") {
```
ilia-db, line 612: [VSCode makes sure](https://github.com/microsoft/vscode/blob/87fd6d3320d4ce70ba896bbf273b960a3f7c200f/src/vs/workbench/contrib/debug/browser/debugConfigurationManager.ts#L150) that you won't get a config for a type you haven't registered this config provider, so this check is not required

PR #931, opened 2023-11-07
Delete yaml from files watched by bundle autocomplete, when the yaml is deleted

kartikgupta-db:
## Changes
Fixes https://github.com/databricks/databricks-vscode/pull/922#discussion_r1382076227


pietern (approved):
Don't know enough about TS, but it looks good.

Would adding test coverage here be doable?

kartikgupta-db:
Testing is difficult here in general. We need a more comprehensive stubbing fw. I will do that in a later PR for most of dabs functionality.

PR #937, opened 2023-11-11
Add config level change events in `ConfigModel`

kartikgupta-db:
## Changes
* Add config level change events in config model. Whenever a config changes, the corresponding event is fired. 

* Also, allow writing auth parameters as an override.


packages/databricks-vscode/src/configuration/BundleConfigReaderWriter.ts
```diff
@@ -68,9 +69,10 @@ export class BundleConfigReaderWriter {
         return target;
     }
 
-    public async getWorkspaceFsPath(target?: BundleTarget) {
-        const filePath = target?.workspace?.file_path;
-        return filePath ? new RemoteUri(filePath) : undefined;
+    public async getWorkspaceFsPath(
```
ilia-db, line 72: Why async, is `target?.workspace?.file_path` a promise? Should we add `Promise<... | undefined>` to the return type?

packages/databricks-vscode/src/configuration/BundleConfigReaderWriter.ts
```diff
@@ -168,19 +170,22 @@ export class BundleConfigReaderWriter {
...
         }
         const data = await parseBundleYaml(file);
         const targetData = data.targets?.[target];
         if (targetData === undefined) {
-            return false;
+            throw new Error(`No target '${target}' for writing '${key}.`);
         }
 
         const newTargetData = this.writerMapping[key](targetData, value);
+        if (JSON.stringify(newTargetData) === JSON.stringify(targetData)) {
```
ilia-db, line 184: Can we move to using deep equality util here? (and in other places)
At least in the past stringify comparison was sensitive to the order of the fields in object (at the creation time).

packages/databricks-vscode/src/configuration/ConfigModel.ts
```diff
@@ -38,22 +38,43 @@ export class ConfigModel implements Disposable {
...
+            for (const key of <(keyof DatabricksConfigs)[]>(
+                Object.keys(newValue)
+            )) {
+                if (
+                    oldValue === null ||
+                    JSON.stringify(oldValue[key]) !==
+                        JSON.stringify(newValue[key])
+                ) {
+                    this.changeEmitters.get(key)?.emitter.fire();
+                    this.onDidChangeAnyEmitter.fire();
```
ilia-db, line 61: Might be useful to move this out of the for loop, to not blast the same event many times (so emit it only once if at least something has changed)

kartikgupta-db: This is fixed in one of the later PRs https://github.com/databricks/databricks-vscode/pull/941/files

packages/databricks-vscode/src/configuration/BundleConfigReaderWriter.ts
kartikgupta-db: The idea was to have all the getter functions to have the same interface so that we can easily loop over them and await all.

packages/databricks-vscode/src/configuration/BundleConfigReaderWriter.ts
kartikgupta-db: That is a great point. I will make a separate PR to replace all instances of json.stringify. I was thinking of using https://www.npmjs.com/package/deep-equal. Do you have any other library in mind or should we implement something on our own?

ilia-db: `lodash.isequal` might be better. We only have to care about primitive values, objects, and arrays, and can technically just create our own utility and don't pull any other additional dependencies, but I think we already have lodash in the bundle, since it's a transitive dependency, might as well make it a direct dependency

ilia-db (approved):

PR #940, opened 2023-11-11
 Refactor configuration UI to better support override and bundle configuration model

kartikgupta-db:
## Changes
* Split `ConfigurationDataProvider` into smaller `Component`s. This enables a composable UI where each configuration is represented by a separate config. 
* Register `ConnectionManager` change listeners during the `init` call only after a login with saved data has been attempted. This race conditions where some changes to the connections manager, such as re-attaching a sync destination due to changes in `ConfigModel` during init, can cause relogin, even before `loginWithSavedAuth` is attempted. 

**Note**: The UI is a placeholder. We still need to iterate on it to make it both accessible and more inline with our brand design.

https://github.com/databricks/databricks-vscode/assets/88345179/354cf62f-28d1-4706-8425-7fa2f00c51c5

## Tests
* manually


packages/databricks-vscode/src/extension.ts
```diff
@@ -594,9 +596,7 @@ export async function activate(
         })
     );
 
-    connectionManager.init().catch((e) => {
-        window.showErrorMessage(e);
-    });
+    await confgiModel.init();
```
ilia-db, line 599: I like the more explicit `connectionManager.init` better. 

Here we rely on target change event for the actual initialisation, which is not obvious. Technically target may not even change between restarts, it's just an internal implementation detail of the ConfigModel/CachedValue logic that we emit those change events on every startup.

ilia-db: I think we already have problems because of emitting change events on every startup - for example connection manager calls `updateSyncDestinationMapper` when `workspaceFsPath` changes in the model. But since it "changes" on the startup (and before the connection manager has a chase to do the login), the initial call to `updateSyncDestinationMapper` fails with user facing "Invalid sync destination" error here:
https://github.com/databricks/databricks-vscode/blob/bundle-integ/packages/databricks-vscode/src/sync/SyncDestination.ts#L62

It might be better to avoid emitting change events on startup and do more explicit initialisation.

packages/databricks-vscode/src/configuration/ui/ConfigurationDataProvider.ts
```diff
@@ -0,0 +1,124 @@
...
+                this.connectionManager,
+                this.configModel
+            )
+        );
+
+        this.disposables.push(
+            ...this.components,
+            ...this.components.map((c) =>
+                c.onDidChange(() => {
+                    this._onDidChangeTreeData.fire();
```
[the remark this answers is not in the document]
kartikgupta-db: It does sort of debounce in the sense that the UI does not show any change if there is no change produced by the data provider. Also, the different updates are synchronised so only 1 update runs at a time.

packages/databricks-vscode/src/extension.ts
kartikgupta-db: How about we move the listeners to init function? That way, the listeners are registered after login has been attempted once.

ilia-db (approved):

PR #941, opened 2023-11-13
Return config source from `ConfigModel`

kartikgupta-db:
## Changes
* We need config source to display the overrides clearly in the UI. This PR allows config source to be passed through `ConfigModel`.
## Tests
Manual

### DIFF
--- a/packages/databricks-vscode/src/configuration/types.ts
+++ b/packages/databricks-vscode/src/configuration/types.ts
@@ -9,6 +9,10 @@ export type DatabricksConfigs = {
     workspaceFsPath?: string;
 };
 
+export type DatabricksConfigSource = {
+    [key in keyof DatabricksConfigs]: "bundle" | "override";
+};
+
 export const OVERRIDEABLE_CONFIGS = [
     "clusterId",
     "authParams",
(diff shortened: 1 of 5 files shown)


packages/databricks-vscode/src/configuration/types.ts
```diff
@@ -9,6 +9,10 @@ export type DatabricksConfigs = {
     workspaceFsPath?: string;
 };
 
+export type DatabricksConfigSource = {
```
ilia-db, line 12: Nitpick, but it's a bit confusing how we have `DatabricksConfigs` as a plural thing, even though it's just one object with key value pairs. And in some cases we refer to it in a singular form, but in other cases we refer to its keys as configs. And this source name uses a singular notation now. 

I would expect DatabricksConfigs to be DatabricksConfig (without s), and then many other types and methods that operate on its keys renamed to have a key in their names, not config

kartikgupta-db: Makes sense mostly. Not sure I understand the second part. Does it mean that we add the "key" suffix?

ilia-db: Yes, for example this renames:
BUNDLE_CONFIGS -> BUNDLE_KEYS
OVERRIDEABLE_CONFIGS -> OVERRIDEABLE_KEYS
OverrideableConfigs -> OverrideableConfig
isOverrideableConfig -> isOverrideableKey
isBundleConfig -> isBundleKey

kartikgupta-db: Makes sense. I will take a stab at it. I will push it into the same PR if the changes are not too big.

ilia-db (approved):

PR #943, opened 2023-11-18
Replace `JSON.stringify` based object comparisions with `lodash.isEqual`

kartikgupta-db:


packages/databricks-vscode/src/cluster/ClusterManager.ts
```diff
@@ -1,7 +1,7 @@
 import {compute, Time, TimeUnits} from "@databricks/databricks-sdk";
 import {Cluster} from "../sdk-extensions";
 import {CancellationTokenSource, Disposable} from "vscode";
-
+import _ from "lodash";
```
ilia-db, line 4: Let's add `lodash` explicitly to our package.json, otherwise builds can break when we update dependencies. We can also "optimise" and add only a dependency on `lodash.isEqual`

PR #944, opened 2023-11-21
Send environment type (tests, prod) in telemetry events

ilia-db:
## Changes
Set the type based on a presence of TEST_DEFAULT_CLUSTER_ID env var, which is only set in e2e tests.

Also don't sent telemetry to a prod endpoint when we run unit tests.p

## Tests
Manually and with new unit tests


kartikgupta-db (approved):

kartikgupta-db:
Can you also cherry pick this into the bundle branch? We are running tests there as well.

PR #953, opened 2023-11-28
Inject version using correct syntax

ilia-db:


ilia-db:
Inject version using correct syntax

PR #972, opened 2023-12-13
Dissallow all overrides except cluster ID for dev mode and Auth Params for all modes

kartikgupta-db:
## Changes
* Keeping in line with the intended CUJs we only want to allow overriding cluster id and that too only for dev mode. 
* This also hides cluster ID in other modes since it is irrelevant there.


packages/databricks-vscode/src/configuration/ui/ClusterComponent.ts
```diff
@@ -157,6 +133,14 @@ export class ClusterComponent extends BaseComponent {
     public async getChildren(
         parent?: TreeItem
     ): Promise<ConfigurationTreeItem[]> {
+        // Only show cluster details when we are connected and in dev mode. In other modes
+        // there is no concept of a global interactive cluster
+        if (
+            this.connectionManager.state !== "CONNECTED" ||
+            (await this.configModel.get("mode")) !== "dev"
```
ilia-db, line 140: Should we show some `Cluster management is only enabled in "dev" mode` child here? Or maybe change the root item to indicate this "non-dev" state?

Same for the sync destination component

ilia-db (approved):

packages/databricks-vscode/src/configuration/ui/ClusterComponent.ts
kartikgupta-db: Created a ticket.

PR #978, opened 2023-12-15
Refactor config loaders

kartikgupta-db:
## Changes
This changes the `*ConfigReaderWriter` classes into `*ConfigModel` classes. We introduce the following architecture:
* Configs storage is handled by "Storage Medium" such as `BundleFileSet` and `StateStorage`.  They emit event if there is a change. If we can't monitor the underlying medium for changes, we must emit event from the "set" method (see `StateStorage`). 
* Models are responsible for extracting their respective domain of configs from whatever storage medium they choose. They are also responsible for emitting and event if there is a change in any of the configs in their domain. 
  * Models might need some information such as target and auth params, to automatically refresh the configs in their domain. There are multiple setters which are exposed by each loader for taking this information. We want to have multiple setters and not a single one because multiple pieces of information can come at different times. Eg target -> load auth info and login -> pass auth params to loader
* Models are responsible for correctly writing data to the storage medium.


packages/databricks-vscode/src/configuration/ConfigModel.ts
```diff
@@ -97,28 +79,39 @@ export class ConfigModel implements Disposable {
...
 
     constructor(
-        public readonly overrideReaderWriter: ConfigOverrideReaderWriter,
-        public readonly bundleConfigReaderWriter: BundleConfigReaderWriter,
-        private readonly stateStorage: StateStorage,
-        private readonly bundleWatcher: BundleWatcher
+        private readonly overrideableConfigLoader: OverrideableConfigLoader,
+        private readonly overrideableConfigWriter: OverrideableConfigWriter,
+        public readonly bundleFileConfigLoader: BundleFileConfigLoader,
+        private readonly bundleFileConfigWriter: BundleFileConfigWriter,
```
ilia-db, line 90: I don't fully understand the reason behind splitting things into loaders and writers. They all have similar methods and operate on the same data. If the underlying data shape changes or we want to add some new methods, then both classes have to be changed at the same time, indicating that they might need to be the same thing after all. Fewer abstractions and moving parts is usually a good thing. But it feels like I'm missing some benefits here.

On top of that we have chains of CachedValues - the configCache here depends on the bundleFileConfigLoader (plus overrides cache), which depends on the bundleFileSet cache. It's hard to say if the current logic is correct and if all the caches are invalidated at the right times. 

Makes sense to cache the IO responses, but I'm not sure about the intermediate cache from the FileConfigLoader for example. Can we try to simplify? Again, I might be missing some important edge cases, but in my view something like this can be a bit simpler and easier to maintain and evolve:
* `FSConfigModel` - the class that reads and writes bundle configuration from to the file system. Uses the file watcher internally and exposes change events. Can cache the values. BundleFileSet, BundleFileConfigLoader, and BunleFileConfigWriter are removed.
* `OverridesConfigModel` - the class that reads and writes bundle configuration from and to the vscode storage, doesn't really need a cache. OverrideableConfigLoader and OverrideConfigWriter are removed.
* Those models will be used by the main `ConfigModel` and no one else
* The current target name is saved only in the main `ConfigModel`, the internal models accept the current target name name as an argument to their methods, where necessary
* The main `ConfigModel` also doesn't need a cache, but we can keep it if it makes the change events easier to implement

kartikgupta-db: I wanted to split into the loaders because this way we can have multiple loaders for different lifecycle steps for bundles (before auth and post auth), but common writer. You are right that we do not really need this for override. I will revert that to a common one.

kartikgupta-db: As for cache chains, this is essentially an abstraction for limiting change events. 

DataIo classes like BundleFileset cache changes to files and react when there are actual changes.
Loader cache changes to only the relevant configs. And only react when these specific ones change. 
ConfigModel is a consolidated view and that only changes when the specific configs change. 

We have such caches hidden throughout the code. This just makes it more abstracted. We will even need it for override model in the future when we move to storing stuff in project.json again. The general feeling in the room was that people were not happy with vs code internal storage.

packages/databricks-vscode/src/locking/CachedValue.ts
```diff
@@ -12,7 +19,14 @@ export class CachedValue<T> {
             return this.mutex
                 .wait()
```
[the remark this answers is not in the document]
kartikgupta-db: Yes. That is good catch. I will update it here and also make a separate PR to update it everywhere in the code base.

packages/databricks-vscode/src/configuration/ConfigModel.ts
ilia-db: > we can have multiple loaders for different lifecycle steps for bundles (before auth and post auth), but common writer

Hmm, I guess that's a question for the other PR, but why do we want to write to the bundle file something that was produced from the `bundle validate` command? We need to get the config with interpolated variables from the CLI, but that's not something we want to write back to the config file, right?

And another question is why do we need BundleFileConfigLoader at all if we don't do variables interpolation ourselves? Seems like we can't use any of the values from it, as anything can be `${someVar}` string instead of a real value.

packages/databricks-vscode/src/configuration/writers/BundleFileConfigWriter.ts
```diff
@@ -0,0 +1,87 @@
...
+import {BundleFileSet, parseBundleYaml, writeBundleYaml} from "../../bundle";
+import {BundleTarget} from "../../bundle/types";
+import {Mutex} from "../../locking";
+import {BundleFileConfig} from "../types";
+import lodash from "lodash";
+/**
+ * Reads and writes bundle configs. This class does not notify when the configs change.
+ * We use the BundleWatcher to notify when the configs change.
+ */
+export class BundleFileConfigWriter {
```
ilia-db, line 11: Based on the in-person discussion - if we won't need it, let's remove it then.

packages/databricks-vscode/src/configuration/loaders/OverrideableConfigLoaderWriter.ts
```diff
@@ -0,0 +1,73 @@
+import {StateStorage} from "../../vscode-objs/StateStorage";
+import {OverrideableConfig} from "../types";
+import {CachedValue} from "../../locking/CachedValue";
+import {Disposable} from "vscode";
+import {Mutex} from "../../locking";
+
+export class OverrideableConfigLoaderWriter implements Disposable {
```
ilia-db, line 7: `loaders/OverrideableConfigLoaderWriter.ts` - nitpick, but should the "writer" be in the loader directory? Can we move things from here to one level higher and just name things as `*Model` (or without "model", pretty much the same as we had before)?

kartikgupta-db: I agree. I refactored this. Also refactored a bunch of other code to be more streamlined. Ig we can make the next PR a bit more streamlined as well! Thanks for the patience here :)

ilia-db (approved):

PR #979, opened 2023-12-15
Use `bundle validate` to load interpolated view of configs after login

kartikgupta-db:
## Changes
* Add a `ConfigModel` for loading configs - using `bundle validate` - **after** auth is completed. This also interpolates bundle variables and params. Right now we use it for workspace path and clusterId. In the future, we would also want to use it to pull a pre deploy view of the resources. 

**Note**: We still need more discussions on how to handle params. We can have a separate PR for that.


packages/databricks-vscode/src/bundle/AuthenticatedBundleModel.ts
```diff
@@ -0,0 +1,81 @@
...
+import {AuthProvider} from "../configuration/auth/AuthProvider";
+import {Mutex} from "../locking";
+import {CliWrapper} from "../cli/CliWrapper";
+import {BundleTarget} from "./types";
+import {CachedValue} from "../locking/CachedValue";
+import {onError} from "../utils/onErrorDecorator";
+import lodash from "lodash";
+import {workspaceConfigs} from "../vscode-objs/WorkspaceConfigs";
+
+export class AuthenticatedBundleModel implements Disposable {
```
ilia-db, line 12: It's sort of make sense for the AuthenticatedBundleConfigLoader to be named that way (although I prefer `configuration/AuthenticatedBundleConfigModel`), since it explicitly filters out all keys that are not related to the auth. But this one doesn't filter anything, should we just name it `BundleModel`? We can think about better names when (of if) we add more bundle-related models (e.g. pre auth or post deploy that you mentioned).

packages/databricks-vscode/src/configuration/types.ts
```diff
@@ -29,8 +29,23 @@ export type BundleFileConfig = Pick<
     (typeof BUNDLE_FILE_CONFIG_KEYS)[number]
 >;
 
-export const DATABRICKS_CONFIG_KEYS = Array.from(
-    new Set([...OVERRIDEABLE_CONFIG_KEYS, ...BUNDLE_FILE_CONFIG_KEYS])
+export const AUTHENTICATED_BUNDLE_CONFIG = [
```
ilia-db, line 32: Should we add `*_KEYS` for consistency with the other names?

packages/databricks-vscode/src/configuration/types.ts
```diff
@@ -43,7 +58,8 @@ export function isBundleConfigKey(key: any): key is keyof BundleFileConfig {
     return BUNDLE_FILE_CONFIG_KEYS.includes(key);
 }
 
-export interface ConfigReaderWriter<T extends keyof DatabricksConfig> {
-    readAll(target: string): Promise<DatabricksConfig | undefined>;
-    write(key: T, target: string, value?: DatabricksConfig[T]): Promise<void>;
+export function isAuthenticatedBundleConfigKey(
+    key: any
+): key is keyof AuthenticatedBundleConfig {
```
ilia-db, line 63: But the AuthenticatedBundleConfig includes the keys from the DatabricksConfig, so this condition is less strict that our actual logic below. Is that intentional?

kartikgupta-db: Yes. The idea is to be able to quickly disambiguate if a key belongs to one of the models, and can hence be safely passed around.

packages/databricks-vscode/src/bundle/AuthenticatedBundleModel.ts
kartikgupta-db: I believe you will be happy with the new changes :)

ilia-db (approved):

PR #989, opened 2024-01-08
Rollback telemetry package

ilia-db:
There's a problem with the newer version of the telemetry package. Temporary rolling back to the old one.


kartikgupta-db (approved):
Verified with @mgyucht.

PR #995, opened 2024-01-10
Automatically read auth info from override, bundle and .databrickscfg, in order

kartikgupta-db:
## Changes
* Read auth profile from overrides.
* If not found, read the auth information hard coded in bundle (the actual reading is NOT implemented here)
* If not found, find a unique profile matching the host


packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -175,16 +162,54 @@ export class ConnectionManager implements Disposable {
...
+        const host = await this.configModel.get("host");
+        const target = this.configModel.target;
+        if (host === undefined || target === undefined) {
+            return;
+        }
+
+        // Try to load a profile user had previously selected for this target
+        const savedProfile = (
+            await this.configModel.overrideableConfigModel.load(["authProfile"])
+        )?.authProfile;
```
ilia-db, line 175: Seems simpler to just load everything, since we are selecting authProfile from the object in any case.

Same for the authParams later. Or are there some hidden reason to do such partial loading? As I understand we are doing the full loading internally anyway

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -211,10 +236,12 @@ export class ConnectionManager implements Disposable {
...
                 );
 
-                await this.createWsFsRootDirectory(this._workspaceClient);
                 await this.updateSyncDestinationMapper();
                 await this.updateClusterManager();
-                await this.configModel.set("authParams", authProvider.toJSON());
+                await this.configModel.set(
+                    "authProfile",
+                    authProvider.toJSON().profile as string | undefined
+                );
```
ilia-db, line 244: If we always save it to the override here, then potential profile changes in the bundle.yml won't take any effect, or do I misunderstand the logic here?

Should we only save the profile to overrides when it doesn't exist in the bundle config? Although I'm not sure if we even currently loading the profile name from the bundle models

packages/databricks-vscode/src/configuration/ConnectionManager.ts
kartikgupta-db: Good catch. Left over from some previous commits.

PR #1002, opened 2024-01-12
Add a view for DABs resource viewer

kartikgupta-db:
## Changes
* Add a view for DABs resource viewer. 
* The view only appears when the user has selected a valid target
* Fix some callbacks so target is updated properly when bundle file changes.

## Tests
* manual

### DIFF
--- a/packages/databricks-vscode/src/configuration/ui/BundleTargetComponent.ts
+++ b/packages/databricks-vscode/src/configuration/ui/BundleTargetComponent.ts
@@ -22,6 +22,9 @@ export class BundleTargetComponent extends BaseComponent {
         this.disposables.push(
             this.configModel.onDidChangeTarget(() => {
                 this.onDidChangeEmitter.fire();
+            }),
+            this.configModel.bundlePreValidateModel.onDidChange(() => {
+                this.onDidChangeEmitter.fire();
             })
         );
     }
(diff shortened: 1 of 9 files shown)


packages/databricks-vscode/src/configuration/ui/BundleTargetComponent.ts
```diff
@@ -22,6 +22,9 @@ export class BundleTargetComponent extends BaseComponent {
         this.disposables.push(
             this.configModel.onDidChangeTarget(() => {
                 this.onDidChangeEmitter.fire();
+            }),
+            this.configModel.bundlePreValidateModel.onDidChange(() => {
```
ilia-db, line 26: Can we avoid exposing sub-models from `configModel`? From the maintenance perspective it might make our life easier if we were to keep them as an implementation detail. We can expose some more specific change events instead, or just listen to specific keys?

kartikgupta-db: I agree to not expose from configModel. But I would rather import and depend on the submodels explicitly. 

Some of our Ui elements need access to specific data from each of the bundle life cycle models (pre validate, post validate, remote tf state). If we add specific change events to configModel, it will devolve into events highly coupled with these submodels anyway. 

Having said that, we don't really need it here 😅

ilia-db (approved):

PR #1003, opened 2024-01-12
Add a method to wrap event callbacks (lambdas) with onError functionality

kartikgupta-db:

### DIFF
--- a/packages/databricks-vscode/src/configuration/models/ConfigModel.ts
+++ b/packages/databricks-vscode/src/configuration/models/ConfigModel.ts
@@ -2,7 +2,7 @@ import {Disposable, EventEmitter, Uri, Event} from "vscode";
 import {Mutex} from "../../locking";
 import {CachedValue} from "../../locking/CachedValue";
 import {StateStorage} from "../../vscode-objs/StateStorage";
-import {onError} from "../../utils/onErrorDecorator";
+import {onError, onErrorLambda} from "../../utils/onErrorDecorator";
 import {AuthProvider} from "../auth/AuthProvider";
 import {
     OverrideableConfigModel,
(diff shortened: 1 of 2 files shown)


packages/databricks-vscode/src/bundle/models/BundleValidateModel.ts
```diff
@@ -25,9 +26,14 @@ export class BundleValidateModel extends BaseModelWithStateCache<BundleValidateS
...
         this.disposables.push(
-            this.bundleWatcher.onDidChange(async () => {
-                await this.stateCache.refresh();
-            })
+            this.bundleWatcher.onDidChange(
+                withOnErrorHandler(
+                    async () => {
+                        await this.stateCache.refresh();
+                    },
+                    {log: true, throw: false}
```
ilia-db, line 34: Aren't those the defaults?

packages/databricks-vscode/src/configuration/models/ConfigModel.ts
```diff
@@ -125,10 +125,12 @@ export class ConfigModel implements Disposable {
...
-            this.overrideableConfigModel.onDidChange(async () => {
-                //refresh cache to trigger onDidChange event
-                await this.configCache.refresh();
-            }),
+            this.overrideableConfigModel.onDidChange(
+                withOnErrorHandler(async () => {
+                    //refresh cache to trigger onDidChange event
+                    await this.configCache.refresh();
+                })
+            ),
```
ilia-db, line 133: Why do we wrap the refresh here, but not in the other listeners below?

kartikgupta-db: I missed it. I will wrap all of these in a separate PR. This is mostly to fix the errors we are seeing rn.

packages/databricks-vscode/src/bundle/models/BundleValidateModel.ts
kartikgupta-db: Yes, I just felt it is easier to keep track of what thing does what and what doesn't this way.

packages/databricks-vscode/src/configuration/models/ConfigModel.ts
ilia-db: Can we add something like `cachedValue.onDidRefrechError` instead?

kartikgupta-db: Added a wrapper for event emitter that handles error handling with some defaults for cached value. I can do a pass to move all our events to this.

packages/databricks-vscode/src/configuration/ui/ConfigurationDataProvider.ts
```diff
@@ -42,7 +42,7 @@ export class ConfigurationDataProvider
         private readonly configModel: ConfigModel
     ) {
         this.disposables.push(
-            this.bundleProjectManager.onDidChangeStatus(() => {
+            this.bundleProjectManager.onDidChangeStatus(async () => {
```
kartikgupta-db, line 45: Need this because it is easier to handle always async functions in the event wrapper.

ilia-db (approved):

PR #1011, opened 2024-01-22
Add bundle project manager

ilia-db:
"Bundle project manager" is being initialised before configModel or connectionManager.
If necessary, it does automatic migration from the legacy projects to the bundles (will be implemented in the follow-up).
If it detects multiple bundle projects and offers to choose which one to open. If it detects a bundle config in the root of the workspace it kicks off configModel and connectionManager.
If it detects no bundle config in the root, it offers to create a project (spawning the Databricks CLI in the terminal)

I've played around with displaying notifications for things like "hey, chose one project to open", or "there's no config in your workspace, would you like to init a new project?", but they don't flow well with existing quick-pick/input wizards and are pretty easy to miss. Currently I'm using welcome views to show such information:

The default view is still the same:
Screenshot 2024-01-22 at 10 11 20

If we don't detect bundle config in the root, but there are subfolders with bundles, we show an additional text and a button:
Screenshot 2024-01-22 at 10 11 58

It's still possible to login in this state, after which we will also show "init" button:

Screenshot 2024-01-22 at 10 13 27

If there are no sub-projects, then we just show login button alone in the view:
Screenshot 2024-01-22 at 10 14 54

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -233,7 +233,7 @@
             },
             {
                 "command": "databricks.bundle.initNewProject",
-                "title": "Initialize new Databricks project",
+                "title": "Initialize new project",
                 "enablement": "databricks.context.activated",
                 "category": "Databricks"
             }
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -57,9 +57,9 @@
                 "enablement": "databricks.context.activated && databricks.context.loggedIn"
             },
             {
-                "command": "databricks.connection.configureWorkspace",
+                "command": "databricks.connection.configureLogin",
                 "icon": "$(gear)",
-                "title": "Configure workspace",
+                "title": "Sign in to Databricks workspace",
                 "enablement": "databricks.context.activated",
                 "category": "Databricks"
             },
@@ -278,7 +278,7 @@
             },
             {
                 "view": "configurationView",
-                "contents": "Configure your Databricks workspace:\n[Configure Databricks](command:databricks.connection.configureWorkspace)\n[Show Quickstart](command:databricks.quickstart.open)\nTo learn more about how to use Databricks with VS Code [read our docs](https://docs.databricks.com/dev-tools/vscode-ext.html).",
+                "contents": "Sign in to your Databricks workspace:\n[Sign in to Databricks](command:databricks.connection.configureLogin)\n[Show Quickstart](command:databricks.quickstart.open)\nTo learn more about how to use Databricks with VS Code [read our docs](https://docs.databricks.com/dev-tools/vscode-ext.html).",
                 "when": "workspaceFolderCount > 0 && !databricks.context.loggedIn"
             },
             {
@@ -290,7 +290,7 @@
         "menus": {
             "view/title": [
                 {
-                    "command": "databricks.connection.configureWorkspace",
+                    "command": "databricks.connection.configureLogin",
                     "when": "view == configurationView",
                     "group": "navigation@3"
                 },
@@ -347,7 +347,7 @@
                     "group": "inline@2"
                 },
                 {
-                    "command": "databricks.connection.configureWorkspace",
+                    "command": "databricks.connection.configureLogin",
                     "when": "view == configurationView && viewItem =~ /^databricks.configuration.authType.*$/",
                     "group": "inline@2"
                 },
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -230,6 +230,12 @@
                 "title": "Select a Databricks Asset Bundle target.",
                 "enablement": "databricks.context.activated",
                 "category": "Databricks"
+            },
+            {
+                "command": "databricks.bundle.initNewProject",
+                "title": "Initialize new Databricks project",
+                "enablement": "databricks.context.activated",
+                "category": "Databricks"
             }
         ],
         "viewsContainers": {
@@ -291,6 +297,10 @@
                 },
                 {
                     "command": "databricks.quickstart.open",
+                    "when": "view == configurationView"
+                },
+                {
+                    "command": "databricks.bundle.initNewProject",
                     "when": "view == configurationView"
                 },
                 {
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -273,13 +273,8 @@
             },
             {
                 "view": "configurationView",
-                "contents": "Current workspace has no Databricks configuration at the root level, do you want to initialize a new Databricks project?\n[Initialize New Project](command:databricks.bundle.initNewProject)\n",
-                "when": "workspaceFolderCount > 0 && databricks.context.loggedIn"
-            },
-            {
-                "view": "configurationView",
-                "contents": "Sign in to your Databricks workspace:\n[Sign in to Databricks](command:databricks.connection.configureLogin)\n[Show Quickstart](command:databricks.quickstart.open)\nTo learn more about how to use Databricks with VS Code [read our docs](https://docs.databricks.com/dev-tools/vscode-ext.html).",
-                "when": "workspaceFolderCount > 0 && !databricks.context.loggedIn"
+                "contents": "Current workspace has no Databricks configuration at the root level, do you want to initialize a new Databricks project?\n[Initialize New Project](command:databricks.bundle.initNewProject)\n[Show Quickstart](command:databricks.quickstart.open)\nTo learn more about how to use Databricks with VS Code [read our docs](https://docs.databricks.com/dev-tools/vscode-ext.html).",
+                "when": "workspaceFolderCount > 0"
             },
             {
                 "view": "configurationView",
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -268,8 +268,18 @@
         "viewsWelcome": [
             {
                 "view": "configurationView",
-                "contents": "In order to connect to a cluster you first have to configure your Databricks workspace:\n[Configure Databricks](command:databricks.connection.configureWorkspace)\n[Show Quickstart](command:databricks.quickstart.open)\nTo learn more about how to use Databricks with VS Code [read our docs](https://docs.databricks.com/dev-tools/vscode-ext.html).",
-                "when": "workspaceFolderCount > 0"
+                "contents": "There are multiple Databricks projects in the workspace, please chose which one to open:\n[Open Existing Project](command:databricks.bundle.openSubProject)\n",
+                "when": "workspaceFolderCount > 0 && databricks.context.subProjectsAvailable"
+            },
+            {
+                "view": "configurationView",
+                "contents": "Current workspace has no Databricks configuration at the root level, do you want to initialize a new Databricks project?\n[Initialize New Project](command:databricks.bundle.initNewProject)\n",
+                "when": "workspaceFolderCount > 0 && databricks.context.loggedIn"
+            },
+            {
+                "view": "configurationView",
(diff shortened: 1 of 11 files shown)


packages/databricks-vscode/package.json
```diff
@@ -268,8 +268,18 @@
...
+                "when": "workspaceFolderCount > 0 && databricks.context.subProjectsAvailable"
+            },
+            {
+                "view": "configurationView",
+                "contents": "Current workspace has no Databricks configuration at the root level, do you want to initialize a new Databricks project?\n[Initialize New Project](command:databricks.bundle.initNewProject)\n",
+                "when": "workspaceFolderCount > 0 && databricks.context.loggedIn"
+            },
+            {
+                "view": "configurationView",
+                "contents": "Configure your Databricks workspace:\n[Configure Databricks](command:databricks.connection.configureWorkspace)\n[Show Quickstart](command:databricks.quickstart.open)\nTo learn more about how to use Databricks with VS Code [read our docs](https://docs.databricks.com/dev-tools/vscode-ext.html).",
```
kartikgupta-db, line 281: We can probably take out the "configure" button now. I don't think it makes sense with the more guided flow we have with the bundles

packages/databricks-vscode/src/bundle/BundleFileSet.ts
```diff
@@ -54,6 +56,22 @@ export class BundleFileSet {
...
 
+    async getSubProjects(): Promise<{relative: Uri; absolute: Uri}[]> {
+        const subProjectRoots = await glob.glob(
+            toGlobPath(this.getAbsolutePath(this.subProjectFilePattern).fsPath),
+            {nocase: process.platform === "win32"}
+        );
+        const delimiter = process.platform === "win32" ? "\\" : "/";
+        return subProjectRoots.map((rootFile) => {
+            const partsWithoutFileName = rootFile.split(delimiter).slice(0, -1);
+            const absolute = Uri.file(partsWithoutFileName.join(delimiter));
```
kartikgupta-db, line 67: This can be simplified by [path.dirname](https://nodejs.org/api/path.html#pathdirnamepath)

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    private _subProjects = new CachedValue<{absolute: Uri; relative: Uri}[]>(
+        async () => {
+            const subProjects = await this.bundleFileSet.getSubProjects();
+            this.customWhenContext.setSubProjectsAvailable(
+                subProjects?.length > 0
+            );
+            return subProjects;
+        }
+    );
```
kartikgupta-db, line 41: We should refresh these caches when bundle file set changes.

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
```diff
@@ -0,0 +1,143 @@
...
+export class BundleProjectManager {
+    private logger = logging.NamedLogger.getOrCreate(Loggers.Extension);
+    private disposables: Disposable[] = [];
+
+    private _isBundleProject = new CachedValue<boolean>(async () => {
+        const rootBundleFile = await this.bundleFileSet.getRootFile();
+        return rootBundleFile !== undefined;
+    });
+
+    public onDidChangeStatus = this._isBundleProject.onDidChange;
```
kartikgupta-db, line 26: Let's not expose this. This class is too meta for other classes to take a dependency on this. I would want it to drive all downstream classes (as it is doing now), but nothing else should take a dependency.

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
```diff
@@ -0,0 +1,143 @@
...
+        private configModel: ConfigModel,
+        private bundleFileSet: BundleFileSet,
+        private workspaceUri: Uri
+    ) {}
+
+    public async isBundleProject(): Promise<boolean> {
+        return await this._isBundleProject.value;
+    }
+
+    public async configureWorkspace(): Promise<void> {
```
kartikgupta-db, line 55: We should maybe try to reconfigure when _isBundleProject changes?

packages/databricks-vscode/src/configuration/ui/ConfigurationDataProvider.ts
```diff
@@ -48,6 +50,9 @@ export class ConfigurationDataProvider
         );
 
         this.disposables.push(
+            this.bundleProjectManager.onDidChangeStatus(() => {
+                this._onDidChangeTreeData.fire();
+            }),
```
kartikgupta-db, line 55: I see why this event is needed now, but is there a way to maybe dispose this class from the BundleProjectManager and then recreate it when we have a project? I am still uncomfortable with exporting this event. 

Not the highest priority immediately, but definitely would be careful of taking anymore dependencies on the init class.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -247,11 +246,16 @@ export class ConnectionManager implements Disposable {
 
                 await this.updateSyncDestinationMapper();
                 await this.updateClusterManager();
-                await this.configModel.set(
-                    "authProfile",
-                    authProvider.toJSON().profile as string | undefined
-                );
+                // Target may not exist when we login in the workspace without any bundle configuration
+                // (e.g. for the sake of bundle initialization wizard)
+                if (this.configModel.target) {
```
kartikgupta-db, line 251: In that case, we can avoid using this function entirely. 
1. configureLogin gives you a configured auth provider
2. We should add a authProvider.check to configureLogin. This will mean that configureLogin always gives a valid authProvider
3. Use the authProvider to initialise a new sdk client and set that as the client for a new service. 
4. Spawn a dedicated terminal for bundle init. 

This is only for the case where users do choose to make a new auth profile. When they don't we can use the existing infra from connectionManager directly. 

This also ensures any code related to bundle init remains separate from stuff that runs after init.

packages/databricks-vscode/src/configuration/ui/ConfigurationDataProvider.ts
ilia-db: Currently bundle project manager is not really an "init" class. For that we need to move a lot of initialisation and creation of classes into it, plus add disposal of them when we detect that the bundle config has disappeared. I would prefer to do it in one of the follow ups that focuses only on such restructuring though

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
ilia-db: Makes sense, for that we need to move a lot of initialisation logic into the project manager so we can properly dispose it during "reconfiguration" process. Let me do it in a follow up after we have the init wizard in a terminal

packages/databricks-vscode/src/configuration/ConnectionManager.ts
ilia-db: My plan was to continue using the same login flow whether we are in a bundle project or not.`configureLogin` command can be called in both situations. If users then follow up with bundle-init the plan is to do logout/login sequence to pick up new bundle auth config. In most cases though bundle-init will result in us prompting to open a new window with a folder into which the project was generated.

It seems simpler if we only have one flow, but I might be missing some of your reasoning

packages/databricks-vscode/package.json
ilia-db: Removed the button in the new revision

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
ilia-db: Improved a bit the initialization flow in the new revision

packages/databricks-vscode/src/configuration/ConnectionManager.ts
ilia-db: Improved this in the new revision

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
```diff
@@ -0,0 +1,254 @@
...
+        // We have a configurationView that shows "open project" button if the context value is true.
+        await this._subProjects.refresh();
+
+        const isLegacyProject = await this._isLegacyProject.value;
+        if (isLegacyProject) {
+            this.logger.debug(
+                "Detected a legacy project.json, starting automatic migration"
+            );
+            await this.migrateProjectJsonToBundle();
+        }
```
kartikgupta-db, line 101: What if it is both a legacy project and a bundleProject at the same time? Can both flows trigger at the same time for that case?

kartikgupta-db: Maybe inverting the order of operations here helps. We return if it is a legacy project and then refresh other caches. 

Ideally we should be able to not depend on refreshes for critical flows.

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    private async migrateProjectJsonToBundle() {
+        // TODO
+    }
+
+    public async initNewProject() {
+        const authProvider = await LoginWizard.run(this.cli, this.configModel);
+        if (!authProvider || !(await authProvider.check())) {
+            return;
+        }
```
kartikgupta-db, line 177: Can we show users a prompt if they want to use the already connected authentication method? In that case, we can use auth provider from connection manager.

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
ilia-db: It depends on how we want to do the migration. It might be safer not to remove existing project.json even after we created bundle config based on it. In that case the project is both "legacy" and "bundle".

I think here it makes sense to return after the `this._isBundleProject.refresh()` if we are in a bundle project.

kartikgupta-db: We might not want to remove the project.json for sometime, in case users want to revert to older extension.

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
ilia-db: Currently the command to init new project is only exposed in non-logged-in state. But let me expose it in the command palette to as "Initialise new project using existing auth" and update the logic

packages/databricks-vscode/package.json
```diff
@@ -293,6 +299,10 @@
                     "command": "databricks.quickstart.open",
                     "when": "view == configurationView"
                 },
+                {
+                    "command": "databricks.bundle.initNewProject",
+                    "when": "view == configurationView"
+                },
```
kartikgupta-db, line 305: The command has no icon. How does it show up in the bar?

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
```diff
@@ -171,8 +178,15 @@ export class BundleProjectManager {
     }
 
     public async initNewProject() {
-        const authProvider = await LoginWizard.run(this.cli, this.configModel);
+        let authProvider =
+            this.connectionManager.databricksWorkspace?.authProvider;
+        if (!authProvider) {
+            authProvider = await LoginWizard.run(this.cli, this.configModel);
+        }
```
kartikgupta-db, line 185: No strong opinions here, but might be nice to ask users if they want to continue using the same auth?

kartikgupta-db (approved):
Minor nits. With this, we should already have a pretty good devloop!

packages/databricks-vscode/package.json
ilia-db: It's grouped with the quick Quickstart

Screenshot 2024-01-24 at 11 17 35

PR #1012, opened 2024-01-24
Deploy and run resource

kartikgupta-db:
## Changes
* show deploy and run button in the UI
* Open a new terminal that shows the live run of the deployment. Terminals are keyed by `resource-key` and `target`. This allows us to reuse existing terminals for a new run. 
* 
* Running state is preserved across target changes. 
* The run button changes to cancel button when running. Also show a spinner for the relevant resources. **Note**: The cancel button ONLY kills the cli process. It does not cancel the online run.

### DIFF
--- a/packages/databricks-vscode/src/bundle/BundleRunManager.ts
+++ b/packages/databricks-vscode/src/bundle/BundleRunManager.ts
@@ -1,8 +1,18 @@
-import {CancellationTokenSource, Disposable, Terminal, window} from "vscode";
+import {
+    CancellationTokenSource,
+    Disposable,
+    EventEmitter,
+    Terminal,
+    window,
+} from "vscode";
 import {BundleRemoteStateModel} from "./models/BundleRemoteStateModel";
-import {CustomOutputTerminal} from "./tasks/CustomOutputTerminal";
+import {CustomOutputTerminal} from "./CustomOutputTerminal";
+import {Mutex} from "../locking";
 
 export class BundleRunManager implements Disposable {
+    private onDidChangeEmitter = new EventEmitter<void>();
+    readonly onDidChange = this.onDidChangeEmitter.event;
+
     private disposables: Disposable[] = [];
     private terminalDetails: Map<
         string,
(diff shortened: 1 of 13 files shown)


packages/databricks-vscode/src/bundle/BundleRunManager.ts
```diff
@@ -0,0 +1,184 @@
...
+                                    new Error(
+                                        `Process exited with code ${exitCode}`
+                                    )
+                                );
+                            }
+                        },
+                        window.onDidCloseTerminal((e) => {
+                            e.name === terminal.terminal.name &&
+                                reject(new Error("Terminal closed"));
+                        })
```
ilia-db, line 128: I think the intent is to add it to `disposables`, but we are adding it as a second arg to the `onDidCloseProcess`

packages/databricks-vscode/src/bundle/BundleRunManager.ts
```diff
@@ -0,0 +1,184 @@
...
+                            } else {
+                                reject(
+                                    new Error(
+                                        `Process exited with code ${exitCode}`
+                                    )
+                                );
+                            }
+                        },
+                        window.onDidCloseTerminal((e) => {
+                            e.name === terminal.terminal.name &&
```
ilia-db, line 126: Should we just compare refs? E.g. `e === terminal.terminal`? Shouldn't change much, but since we are re-using terminal names we might end up with some bugs were it will be relevant

packages/databricks-vscode/src/bundle/BundleRunManager.ts
```diff
@@ -0,0 +1,184 @@
...
+            }
+            if (terminal.pty.process !== undefined) {
+                // There is already a process running. Raise error
+                throw new Error(
+                    `Process already running. Pid: ${terminal.pty.process.pid}`
+                );
+            }
+            this.cancellationTokenSources.get(terminalName)?.cancel();
+            this.cancellationTokenSources.get(terminalName)?.dispose();
+            this.cancellationTokenSources.delete(terminalName);
```
ilia-db, line 82: Is that even possible to have the cancellation token at this point in time? Would that mean that a process is already running and we should throw an error instead, as we do several lines above?

packages/databricks-vscode/src/bundle/BundleRunManager.ts
```diff
@@ -0,0 +1,184 @@
...
+                    terminal?.pty.close();
+                    //Dispose self on cancellation
+                    onCancellationEvent.dispose();
+                });
+
+            const cmd = this.bundleRemoteStateModel.getRunCommand(resourceKey);
+
+            await this.isRunningMutex.synchronise(async () => {
+                this._isRunning.set(terminalName, true);
+                this.onDidChangeEmitter.fire();
```
ilia-db, line 100: Hmm, what are we synchronising here? The code doesn't look async

ilia-db (approved):
Left some comments, but feel free to address them in follow up PRs, so we can merge this one earlier

packages/databricks-vscode/src/bundle/BundleRunManager.ts
kartikgupta-db: We want to have the same terminal name as well. I was keying terminal names with it. I think there was some issue with using references. They are not the same somehow.

PR #1013, opened 2024-01-25
Migrate from project.json to databricks.yml

ilia-db:
Migrate legacy project.json to databricks.yml bundle config

So far the migration-template lives in the extension dir. It also doesn't have the right amount of comments and explainations, we should follow up on that before GA


packages/databricks-vscode/src/bundle/migration-template/template/databricks.yml.tmpl
```diff
@@ -0,0 +1,52 @@
...
+      {{- if .root_path}}
+      root_path: {{.root_path}}
+      {{- end}}
+
+  ## Optionally, there could be a 'staging' target here.
+  ## (See Databricks docs on CI/CD at https://docs.databricks.com/dev-tools/bundles/index.html.)
+  #
+  # staging:
+  #   workspace:
+  #     host: {{workspace_host}}
```
kartikgupta-db, line 34:
```suggestion
  ## Optionally, there could be a 'prod' target here.
  ## (See Databricks docs on CI/CD at https://docs.databricks.com/dev-tools/bundles/index.html.)
  #
  # prod:
  #   workspace:
  #     host: {{workspace_host}}
```

packages/databricks-vscode/src/bundle/migration-template/databricks_template_schema.json
```diff
@@ -0,0 +1,19 @@
+{
```
kartikgupta-db, line 1: Why do we have this file?

packages/databricks-vscode/src/file-managers/ProjectConfigFile.ts
```diff
@@ -0,0 +1,121 @@
+import path from "node:path";
```
kartikgupta-db, line 1: Looking at this file always scares me X:

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -284,6 +284,40 @@ export class CliWrapper {
...
+                "init",
+                templateDirPath,
+                "--output-dir",
+                outputDirPath,
+                "--config-file",
+                initConfigFilePath,
+            ],
+            {env: this.getBundleInitEnvVars(authProvider)}
+        );
+    }
```
kartikgupta-db, line 319: Maybe let's launch it in the same terminal as the normal bundle init? Just easier to debug when it fails, even for the user. And they have some visual feedback for automatic bundle init. Rn it seems like it is completely behind the scenes.

packages/databricks-vscode/src/bundle/migration-template/databricks_template_schema.json
ilia-db: It's required for all templates. That's how CLI knows what template variables to expect, and in the interactive mode this schema is used to ask questions.

packages/databricks-vscode/src/cli/CliWrapper.ts
ilia-db: We will show an error to the users eventually (when they go to the manual migration flow), but I think it makes sense to add error notification in the automatic migration flow (right now there's only error log).

Opening terminal makes sense for transparency, but it goes against the desire of not bothering existing project.json users.  Plus we will want to monitor command completion, which means custom terminal with pty, which adds a bunch of complexity

kartikgupta-db (approved):

PR #1016, opened 2024-01-26
Use sdk with a new version of fetch

ilia-db:
The new sdk uses undici instead of fetch


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@databricks/databricks-vscode",
-    "version": "1.2.5",
+    "version": "1.2.6",
```
kartikgupta-db, line 3: Please revert the version changes

kartikgupta-db (approved):

PR #1017, opened 2024-01-28
Show run status for jobs and pipeline

kartikgupta-db:
## Changes
* Show a run status treeItem for job, pipelines and each of the tasks. 
* The cancel button now also cancels the remote job / pipeline.


packages/databricks-vscode/src/bundle/run/JobRunStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+        const match = output.match(/.*\/run\/(\d*).*/);
+        if (match === null) {
+            return;
+        }
+        this.runId = match[1];
+
+        this.startPolling();
+    }
+
+    @onError({popup: {prefix: "Error starting polling."}})
```
ilia-db, line 33: For a user facing popup more descriptive name would be better - something like "Failed to check the run status of the job". Similar comment for pipelines

packages/databricks-vscode/src/bundle/run/JobRunStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+            ).wait({
+                onProgress: async (progress) => {
+                    this.data = progress;
+                    this.onDidChangeEmitter.fire();
+                },
+            });
+        } catch (e) {
+            if (e instanceof JobsError || e instanceof JobsRetriableError) {
+                // On progress is already fired for these status updates. We let the listeners handle it.
+                return;
```
ilia-db, line 57: Are we fine with being stuck in the `running` state here?

ilia-db (approved):

packages/databricks-vscode/src/bundle/run/JobRunStatus.ts
kartikgupta-db: Default is 10 mins

kartikgupta-db: Actually, makes sense. We should change the running state here. The running state here is mostly for polling being running or not, which need not coincide with the actual run happening or not. But no harm in keeping this state consistent.

PR #1020, opened 2024-01-30
Show deploy and run commands in terminal

kartikgupta-db:
## Changes
* show commands for deploying a bundle and running it in the terminal
* synchronise deploys so multiple deploys do not occur simultaneously


packages/databricks-vscode/src/bundle/BundleCommands.ts
```diff
@@ -61,6 +65,9 @@ export class BundleCommands implements Disposable {
         await this.refreshRemoteState();
     }
 
+    private deployMutex = new Mutex();
+
+    @Mutex.synchronise("deployMutex")
```
ilia-db, line 70: If someone clicks on the deploy button 10 times, will we just deploy the bundle 10 times one after another? Should we throttle here instead? Similar questions for the other command above

ilia-db (approved):

PR #1027, opened 2024-02-01
List profiles with all auth methods

fjakobs:
## Changes

JS SDK now supports all auth methods so we no longer have to filter out unsupported auth methods.
With the bump of the SDK this PR also adds support for Databricks OAuth U2M on Azure and GCP.

## Tests

Manually tests


packages/databricks-vscode/src/configuration/configureWorkspaceWizard.ts
```diff
@@ -209,15 +209,7 @@ async function listProfiles(cliWrapper: CliWrapper) {
...
-    return profiles.filter((profile) => {
-        return [
-            "pat",
-            "basic",
-            "azure-cli",
-            "oauth-m2m",
-            "azure-client-secret",
-        ].includes(profile.authType);
-    });
+    return profiles;
```
ilia-db, line 212: But will we be able to login with any of the new profiles? Later on we limit the selection to the 3 auth types, if I understand correctly:

https://github.com/databricks/databricks-vscode/blob/main/packages/databricks-vscode/src/configuration/configureWorkspaceWizard.ts#L168

And then we create AuthProviders for the same limited set of auth types

https://github.com/databricks/databricks-vscode/blob/main/packages/databricks-vscode/src/configuration/auth/AuthProvider.ts#L73-L92

It would make sense to add AuthProviders for all the newly added auth types

fjakobs: These are two separate things. The filter I removed was only for the `profile` auth type. Some of the auth methods need to be manually configured. I would leave this out of scope for now.

ilia-db (approved):

PR #1029, opened 2024-02-01
Use a tree representation to structure resource explorer UI components

kartikgupta-db:
## Changes
* Use a recursive tree representation for bundle resource expliorer UI components. This allows us to better structure UI components than the renderer based approach. 
* Add deeplinks for all bundle resources and run statuses.


packages/databricks-vscode/src/bundle/models/BundleRemoteStateModel.ts
```diff
@@ -3,32 +3,31 @@ import {CliWrapper} from "../../cli/CliWrapper";
...
 import {Loggers} from "../../logger";
 import {Context, context} from "@databricks/databricks-sdk";
+import {BundleValidateModel} from "./BundleValidateModel";
 
 /* eslint-disable @typescript-eslint/naming-convention */
-type Resources = Required<BundleTarget>["resources"];
-export type Resource<K extends keyof Required<Resources>> =
-    Required<Resources>[K];
-
+export type BundleResourceModifiedStatus = "CREATED" | "DELETED" | "UPDATED";
```
ilia-db, line 16: CLI returns it in lowercase though

packages/databricks-vscode/src/ui/bundle-resource-explorer/BundleCommands.ts
```diff
@@ -90,25 +80,27 @@ export class BundleCommands implements Disposable {
...
+        if (!isRunnable(type)) {
+            throw new Error(`Cannot run resource of type ${type}`);
         }
         //TODO: Don't deploy if there is no diff between local and remote state
         await this.deploy();
 
         await this.bundleRunStatusManager.run(
-            treeNode.resourceKey,
-            treeNode.type
+            (treeNode as any).resourceKey,
```
ilia-db, line 91: Can we avoid `any` here? Why do we not have `resourceKey` on `BundleResourceExplorerTreeNode` anymore if we still want to use it here?

packages/databricks-vscode/src/ui/bundle-resource-explorer/PipelineRunStatusTreeNode.ts
```diff
@@ -0,0 +1,135 @@
...
+import {ConnectionManager} from "../../configuration/ConnectionManager";
+
+function getSimplifiedUpdateState(
+    update?: GetUpdateResponse["update"]
+): SimplifiedRunState {
+    if (update?.state === undefined) {
+        return "Unknown";
+    }
+
+    switch (update.state) {
```
ilia-db, line 20: We could either add a default case here with the unknown state, or a default case to the getThemeIconForStatus, with question mark icon

packages/databricks-vscode/src/ui/bundle-resource-explorer/JobTreeNode.ts
```diff
@@ -0,0 +1,115 @@
...
+            return [];
+        }
+
+        const children: BundleResourceExplorerTreeNode[] = [];
+        const runMonitor = this.bundleRunStatusManager.runStatuses.get(
+            this.resourceKey
+        ) as JobRunStatus | undefined;
+
+        if (runMonitor?.runId !== undefined) {
+            children.push(new JobRunStatusTreeNode(runMonitor));
```
ilia-db, line 70: I guess we should add `this` as a "parent" argument to the JobRunStatusTreeNode constructor.
We are using the `parent` constructor argument inconsistently, might be better to change the constructor types of all the elements that should have parents to have non optional `parent` argument.

On the other hand we can remove all the "parent" handling, since we aren't using the reveal API and don't have to implement `getParent` method in our tree provider.

packages/databricks-vscode/src/ui/bundle-resource-explorer/BundleCommands.ts
kartikgupta-db: We only have it on some nodes that represent a runnable node. The nodes are being filtered out above. I will look for some workaround for this any.

packages/databricks-vscode/src/ui/bundle-resource-explorer/JobTreeNode.ts
kartikgupta-db: >On the other hand we can remove all the "parent" handling, since we aren't using the reveal API and don't have to implement getParent method in our tree provider.

I want to use it in the future to jump to say a pipeline from a pipeline task etc.

ilia-db (approved):

PR #1034, opened 2024-02-02
Expose bundle-init wizard in empty workspace

ilia-db:
- Moved out bundle init logic out of BundleProjectManager into BundleInitWizard
- Moved some logic to "pure" functions so they can be called without initializing a bunch of dependencies
- Welcome view in the empty workspace now has Initialize Project button
- Improve initial terminal prompt of the bundle-init wizard


kartikgupta-db (approved):
Love this!

PR #1051, opened 2024-02-07
Initialise -> Initialize

ilia-db:
Initialise -> Initialize


kartikgupta-db (approved):
Thanks! That was incredibly quick!

PR #1057, opened 2024-02-08
Listen to changes to "host" in databricks.yml to retrigger login

kartikgupta-db:
## Changes
Auth resolution does not happen again when pre validate configs change (host, mode etc). We listen to specific changes to host to retrigger login. 

Also, check if the current host is not the same as the one in the saved profile (in case someone changes host in the bundle.yml).


packages/databricks-vscode/src/configuration/models/ConfigModel.ts
```diff
@@ -131,6 +131,9 @@ export class ConfigModel implements Disposable {
             }),
             this.bundlePreValidateModel.onDidChange(async () => {
                 await this.readTarget();
+                // We fire the target change emitter because auth info might have changed,
+                // which we handle exactly the same way as a target change.
+                this.onDidChangeTargetEmitter.fire();
```
ilia-db, line 136: Seems a bit hacky - we will either emit an event without target actually changing, or emit the event twice. Both cases can lead to some bugs down the road, when we add more listeners for this event. 

Can we just listen to `configModel.jonDidChangeKey("host")` on the connection manager side? And eventually listen to more keys, like "authParams", when we support them. We even have a TODO over there about it

kartikgupta-db: Yes. That actually works better. Will update.

packages/databricks-vscode/src/configuration/ConfigureWorkspaceWizard.ts
```diff
@@ -194,10 +194,7 @@ export class ConfigureWorkspaceWizard {
         if (pick.profile !== undefined) {
             // TODO: If the profile has an auth type with deeper support in the Extension, (azure-cli, databricks-cli),
```
fjakobs, line 195: We should track this TODO and at least cover Azure CLI and Databricks CLI auth

kartikgupta-db: Might as well fix it :). 
https://github.com/databricks/databricks-vscode/pull/1061

packages/databricks-vscode/src/configuration/auth/AuthProvider.ts
```diff
@@ -125,6 +125,14 @@ export abstract class AuthProvider {
 }
 
 export class ProfileAuthProvider extends AuthProvider {
+    static async from(profile: string, checked = false) {
+        const host = await new WorkspaceClient(
+            ProfileAuthProvider._getSdkConfig(profile)
+        ).apiClient.config.getHost();
```
ilia-db, line 131: `workspaceClient.apiClient.config` is the config we pass to the workspace client in the first place (created by `_getSdkConfig`). Is there a hidden reason behind creating workspace client? Can we document it here then?

kartikgupta-db: Good catch. No reason. Removed.

PR #1059, opened 2024-02-08
Read data from global workspace block in databaricks.yml

kartikgupta-db:
## Changes
* Bundles support loading workspace data (host, sync destination) from a global "workspace" block, which is overriden by value of the optional workspace block in each target block.
* This PR adds support for this behaviour.

### DIFF
--- a/packages/databricks-vscode/src/bundle/models/BundlePreValidateModel.ts
+++ b/packages/databricks-vscode/src/bundle/models/BundlePreValidateModel.ts
@@ -33,9 +33,19 @@ export class BundlePreValidateModel extends BaseModelWithStateCache<BundlePreVal
     }
 
     get targets() {
-        return this.bundleFileSet.bundleDataCache.value.then(
-            (data) => data?.targets
-        );
+        return (async () => {
+            const targets = Object.assign(
+                {},
+                (await this.bundleFileSet.bundleDataCache.value).targets ?? {}
+            );
+
+            await Promise.all(
+                Object.keys(targets ?? {}).map(async (key) => {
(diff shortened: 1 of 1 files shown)


packages/databricks-vscode/src/bundle/models/BundlePreValidateModel.ts
```diff
@@ -67,16 +78,30 @@ export class BundlePreValidateModel extends BaseModelWithStateCache<BundlePreVal
             : undefined;
     }
 
+    private async getRawTargetData(target: string) {
+        const bundle = await this.bundleFileSet.bundleDataCache.value;
```
ilia-db, line 82: Can we pass bundle data as an argument and make this function synchronous? I think it will make the other code a bit more readable

packages/databricks-vscode/src/bundle/models/BundlePreValidateModel.ts
```diff
@@ -32,9 +33,17 @@ export class BundlePreValidateModel extends BaseModelWithStateCache<BundlePreVal
...
+        return (async () => {
+            const bundle = await this.bundleFileSet.bundleDataCache.value;
+            const targets = Object.assign({}, bundle.targets ?? {});
+
+            await Promise.all(
+                Object.keys(targets ?? {}).map(async (key) => {
+                    targets[key] = this.getRawTargetData(bundle, key);
+                })
+            );
+            return targets;
```
ilia-db, line 45: `await Promise.all` is no longer necessary here

ilia-db (approved):

PR #1060, opened 2024-02-08
Append profile to databrickscfg instead of writing the full file

kartikgupta-db:
## Changes
* Writing the full .databrickscfg leads to unintended consequences such as deletion of comments and writing invalid profiles. 
* We should append to the file instead.

### DIFF
--- a/packages/databricks-vscode/src/bundle/models/BundlePreValidateModel.ts
+++ b/packages/databricks-vscode/src/bundle/models/BundlePreValidateModel.ts
@@ -1,9 +1,10 @@
 import {Uri} from "vscode";
 import {BundleFileSet, BundleWatcher} from "..";
-import {BundleTarget} from "../types";
+import {BundleSchema, BundleTarget} from "../types";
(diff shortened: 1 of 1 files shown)


packages/databricks-vscode/src/configuration/LoginWizard.ts
```diff
@@ -370,13 +371,17 @@ export async function saveNewProfile(
...
     );
-    iniFile[profileName] = Object.fromEntries(
+
+    const profile: any = {};
+    profile[profileName] = Object.fromEntries(
         Object.entries(iniData).filter((kv) => kv[1] !== undefined)
     );
-
+    const iniStr = ini.stringify(profile);
+    const finalStr = `${os.EOL};This profile is autogenerate by the Databricks Extension for VS Code${os.EOL}${iniStr}`;
```
ilia-db, line 384:
```suggestion
    const finalStr = `${os.EOL};This profile is autogenerated by the Databricks Extension for VS Code${os.EOL}${iniStr}`;
```

ilia-db (approved):

PR #1062, opened 2024-02-09
Hide context buttons during deployment

kartikgupta-db:
## Changes
* Hide buttons such as "open external url", "start cluster", "run resource" etc when deploying because these might be in inconsistent state. 
* Throw a better error from openExternal command when url is not found.


packages/databricks-vscode/package.json
```diff
@@ -394,47 +394,47 @@
...
                 {
                     "command": "databricks.bundle.deployAndRun",
-                    "when": "view == dabsResourceExplorerView && viewItem =~ /^databricks.bundle.*.runnable.*$/",
+                    "when": "view == dabsResourceExplorerView && viewItem =~ /^databricks.bundle.*.runnable.*$/ && databricks.context.bundle.deploymentState == idle",
                     "group": "inline@0"
                 },
                 {
                     "command": "databricks.bundle.cancelRun",
-                    "when": "view == dabsResourceExplorerView && viewItem =~ /^databricks.bundle.*.cancellable.*$/",
+                    "when": "view == dabsResourceExplorerView && viewItem =~ /^databricks.bundle.*.cancellable.*$/ && databricks.context.bundle.deploymentState == running",
```
ilia-db, line 437: `deploying`

 Maybe we can just have `bundle.deploying` flag? Or do we expect to have more than two deployment states?

kartikgupta-db: Possibly error for enabling force?

fjakobs (approved):

packages/databricks-vscode/package.json
ilia-db: True. Also, the first part of my comment was about incorrect `running` state, is should be `deploying`, right?

kartikgupta-db: Aah. Copilot 😅

kartikgupta-db: https://github.com/databricks/databricks-vscode/pull/1064

PR #1063, opened 2024-02-09
Redirect summarise and validate errors to bundle output panel

kartikgupta-db:


ilia-db (commented):
We already have "Databricks Logs" output channel attached to the logger. Why don't we just use the logger for the DABs channel too? Create another logger namespace in the LoggerManager and use it in all the necessary places directly. Should remove the need for wiring all the events and listeners. Additional benefit is that the DABs logger can also have the file transport too (plus others), which might help with debugging customer cases

kartikgupta-db:
That's a good suggestion. I will redo this PR.

kartikgupta-db:
Closed in favour of https://github.com/databricks/databricks-vscode/pull/1076

PR #1066, opened 2024-02-11
Move all dbconnect run options to Databricks Run Icon

kartikgupta-db:
## Changes
* The run icon has new options to run the current python file and debug it.
* The run and debug options run with a bootstrap script (`dbconnect_bootstrap.py`) that 
  * Sets up the env vars by recursing up the file tree and searching for .databricks.env
  * Sets up the Databricks globals


packages/databricks-vscode/package.json
```diff
@@ -255,6 +255,18 @@
                 "enablement": "databricks.context.activated",
                 "category": "Databricks"
             },
+            {
+                "command": "databricks.run.dbconnect.debug",
+                "title": "Databricks Connect: Debug current file",
+                "enablement": "databricks.context.activated && databricks.context.loggedIn && databricks.context.dbconnect.enabled",
```
ilia-db, line 261: Can we have these commands / menu-items available even if dbconnect is not enabled? And then make the commands prompt to enable dbconnect if they detect that it's not installed. It should save users a couple of clicks and should be less confusing than seeing a disabled menu item and knot knowing why it is disabled

ilia-db (commented):
Unit tests are failing

packages/databricks-vscode/src/file-managers/DatabricksEnvFileManager.ts
```diff
@@ -143,23 +110,29 @@ export class DatabricksEnvFileManager implements Disposable {
...
-    @logging.withLogContext(Loggers.Extension)
-    async writeFile(@context ctx?: Context) {
-        await this.connectionManager.waitForConnect();
-
-        await this.mutex.wait();
-        try {
-            const data = Object.entries({
+    async getEnv() {
+        return Object.fromEntries(
+            Object.entries({
```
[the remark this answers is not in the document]
kartikgupta-db: It is easier to directly use the env object everywhere in downstream processes.

packages/databricks-vscode/package.json
kartikgupta-db: It just doesn't appear when dbconnect is disabled. But yeah makes sense. I will add that.

ilia-db (approved):

PR #1067, opened 2024-02-12
Move openDatabricksConfigFile to utils

ilia-db:
Call it from relevant places directly, since the command might not be registered

### DIFF
--- a/packages/databricks-vscode/src/utils/fileUtils.ts
+++ b/packages/databricks-vscode/src/utils/fileUtils.ts
@@ -1,7 +1,8 @@
 import {TextDecoder} from "util";
-import {Uri, workspace} from "vscode";
+import {Uri, workspace, FileSystemError, window} from "vscode";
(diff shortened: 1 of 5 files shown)


packages/databricks-vscode/src/utils/fileUtils.ts
```diff
@@ -48,3 +49,28 @@ export async function waitForDatabricksProject(
 export function getHomedir() {
     return process.env.HOME ?? homedir();
 }
+
+export async function openDatabricksConfigFile() {
+    const homeDir = getHomedir();
+    let filePath =
+        workspaceConfigs.databrickscfgLocation ??
+        process.env.DATABRICKS_CONFIG_FILE ??
+        path.join(homeDir, ".databrickscfg");
```
kartikgupta-db, line 58:
```suggestion
        path.join(getHomedir(), ".databrickscfg");
```

kartikgupta-db (approved):

PR #1071, opened 2024-02-13
Update login wizard and show cancel button for running resources

kartikgupta-db:
* Login wizard
  * In host selection menu, don't show same host multiple times if there are multiple profiles. Only show it once with all the profiles listed in the description.
  * Show existing profiles first in menu for creating/selecting auth type.
 
* Dabs Explorer
  * Show the name of resource in the tooltip for created/deleted resources
  * Show the cancel button for running resources (removed due to a previous PR).

* Actually show the project init icon in the configuration view header
* The URL object is not compared properly by lodash. This meant, we were a host change event whenever anything unrelated in the bundle changed. Which lead to reconnection on every edit to the bundle. Also fix this.

### DIFF
--- a/packages/databricks-vscode/src/locking/CachedValue.ts
+++ b/packages/databricks-vscode/src/locking/CachedValue.ts
@@ -38,13 +38,24 @@ export class CachedValue<T> implements Disposable {
                     newValue = {} as T;
                 }
 
+                const customizer: lodash.IsEqualCustomizer = (value, other) => {
+                    if (value instanceof URL && other instanceof URL) {
+                        return value.toString() === other.toString();
+                    }
+                    return undefined;
+                };
+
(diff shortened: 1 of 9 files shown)


packages/databricks-vscode/src/locking/CachedValue.ts
```diff
@@ -38,13 +38,24 @@ export class CachedValue<T> implements Disposable {
                     newValue = {} as T;
                 }
 
+                const customizer: lodash.IsEqualCustomizer = (value, other) => {
+                    if (value instanceof URL && other instanceof URL) {
+                        return value.toString() === other.toString();
+                    }
```
kartikgupta-db, line 44: If we don't do this, URLs are never compared equal by default.

ilia-db (approved):

PR #1072, opened 2024-02-14
Prevent running resources when cli already launched but no run status

kartikgupta-db:
## Changes
* When we launch a run, there is a lag between the CLI process execution and when the IDE is able to poll for the first status update from the API. We should not allow a second run of the resource during this time. This PR blocks second runs by considering the "unknown" status also as a running state for the viewitem context. This allows vscode to hide the run button, even when we, we do not have a status update. Errors and cancellations are still handled gracefully, because the status reaches completed state. 

* Also, clear the run statuses map on target change.

### DIFF
--- a/packages/databricks-vscode/src/bundle/models/BundleRemoteStateModel.ts
+++ b/packages/databricks-vscode/src/bundle/models/BundleRemoteStateModel.ts
@@ -39,7 +39,7 @@ export function getResource(
             return undefined;
         }
         return prev[k];
-    }, resources);
+    }, resources ?? {});
 }
 
 export class BundleRemoteStateModel extends BaseModelWithStateCache<BundleRemoteState> {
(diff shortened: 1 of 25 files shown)


packages/databricks-vscode/src/bundle/models/BundleRemoteStateModel.ts
```diff
@@ -30,6 +30,17 @@ export type BundleRemoteState = BundleTarget & {
...
+export function getResource(
+    key: string,
+    resources?: BundleRemoteState["resources"]
+) {
+    return key.split(".").reduce((prev: any, k) => {
+        if (prev === undefined) {
+            return undefined;
+        }
+        return prev[k];
+    }, resources);
```
ilia-db, line 42:
```suggestion
    }, resources || {});
```

ilia-db (approved):

PR #1073, opened 2024-02-14
Add e2e tests for bundle init flow

ilia-db:
Add smoke tests for bundle init and auth flows.

Also fixes these problems:
- DATABRICKS_CONFIG_FILE env var was not always respected (for example when creating a new profile)
- bundle file watcher was broken on windows: we were using raw fsPath with backslashes, which doesn't work for globs
- saveNewProfile was broken on windows: `toISOString` returns values with characters that aren't allowed in windows paths, resulting in `ENOENT` errors from `copyFile`


packages/databricks-vscode/src/test/e2e/init.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+        assert(section);
+        const welcome = await section.findWelcomeContent();
+        assert(welcome);
+        try {
+            await getTabByTitle("DATABRICKS.quickstart.md");
+        } catch (error) {
+            console.log(
+                "Failed to find quickstart tab, searching for a changelog instead"
+            );
+            await getTabByTitle("Preview CHANGELOG.md");
```
kartikgupta-db, line 61: We might show "What's new" as well. Do we really need to wait for this?

packages/databricks-vscode/src/test/e2e/init.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+        }
+    });
+
+    it("should open databricks panel and login", async function () {
+        await browser.executeWorkbench((vscode) => {
+            vscode.commands.executeCommand("databricks.bundle.initNewProject");
+        });
+
+        let input = await waitForInput();
```
kartikgupta-db, line 70: Can you add comments about what this input field is for? I think this one is for selecting host?

packages/databricks-vscode/src/test/e2e/init.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+        await sleep(1000);
+
+        input = await waitForInput();
+        await input.selectQuickPick("DEFAULT");
+    });
+
+    it("should initialize new project", async function () {
+        const input = await waitForInput();
+
+        // Clicking on the first pick manually, as selectQuickPick partially deletes default input value
```
kartikgupta-db, line 82: Curious. Is that a bug or a feature?

packages/databricks-vscode/src/test/e2e/init.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+        await initTab.select();
+        // Answer on all wizard questions with defaults until the tab is closed
+        while (true) {
+            const activeTab = await editorView.getActiveTab();
+            if ((await activeTab?.getTitle()) !== title) {
+                break;
+            }
+            await browser.keys([Key.Enter]);
+            await sleep(2000);
+        }
```
kartikgupta-db, line 101:
```suggestion
        browser.waitUntil(async () => {
            const activeTab = await editorView.getActiveTab();
            if ((await activeTab?.getTitle()) !== title) {
                return true;
            }
            await browser.keys([Key.Enter]);
        }, {timeout: 20_000, interval: 2_000})
```

packages/databricks-vscode/src/test/e2e/init.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        const projectInput = await waitForInput();
+        await projectInput.selectQuickPick("/my_project");
+
+        // TODO: wdio-vscode-service needs to be improved for this to work properly
+        // Wait until vscode is re-opened with the new workspace root
```
kartikgupta-db, line 107: Let's create a ticket for them. They are normally extremely responsive. But we might have to build it ourselves with their help.

packages/databricks-vscode/src/test/e2e/utils.ts
```diff
@@ -204,3 +205,38 @@ export async function startSyncIfStopped() {
...
+        async () => {
+            input = await new InputBox(workbench.locatorMap).wait();
+            return input !== undefined;
+        },
+        {timeout: 5000, interval: 500}
+    );
+    assert(input !== undefined);
+    while (await input.hasProgress()) {
+        await sleep(500);
+    }
```
kartikgupta-db, line 240: IMO also use timeout? Easier to debug it from logs. Otherwise the calling function might timeout and then we have to repro to find the line that timedout. 

Not sure what the best practice is here though. What is the best granularity to have timeouts in tests?

packages/databricks-vscode/src/test/e2e/init.e2e.ts
ilia-db: After we've moved to persistent login quick picks we don't have to wait for the changelog tab anymore, as it shouldn't close the quickpick. Right now it's here just to check the actual fact that we show those things. But yeah, I better change it, since we can show other things too

packages/databricks-vscode/src/test/e2e/init.e2e.ts
ilia-db: https://github.com/webdriverio-community/wdio-vscode-service/pull/106

kartikgupta-db: Oh wow. Thank!!

packages/databricks-vscode/src/test/e2e/init.e2e.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        const profileSelectionInput = await waitForInput();
+        await profileSelectionInput.selectQuickPick("DEFAULT");
+    });
+
+    it("should initialize new project", async function () {
+        const parentFolderInput = await waitForInput();
+
+        // Clicking on the first pick manually, as selectQuickPick partially deletes default input value
+        const picks = await parentFolderInput.getQuickPicks();
```
kartikgupta-db, line 71: Can we use tmpdir here instead?

packages/databricks-vscode/src/test/e2e/wdio.conf.ts
```diff
@@ -194,7 +194,11 @@ export const config: Options.Testrunner = {
     // Services take over a specific job you don't want to take care of. They enhance
     // your test setup with almost no effort. Unlike plugins, they don't add new
     // commands. Instead, they hook themselves up into the test process.
-    services: [["vscode", {cachePath: "/tmp/wdio-vscode-service"}]],
+    services: ["vscode"],
+    // TODO: enable the cache again after updating the wdio service
+    // services: [["vscode", {cachePath: "/tmp/wdio-vscode-service"}]],
+
+    cacheDir: "/tmp/wdio-vscode-service",
```
kartikgupta-db, line 201: What happened here?

packages/databricks-vscode/src/test/e2e/auth.e2e.ts
```diff
@@ -0,0 +1,133 @@
...
+        const authMethodInput = await waitForInput();
+        const newProfilePick =
+            await authMethodInput.findQuickPick("NEW_PROFILE");
+        assert(
+            newProfilePick,
+            "NEW_PROFILE is absent in the quick pick selection"
+        );
+        await newProfilePick.select();
+        await waitForLogin("NEW_PROFILE");
+    });
```
kartikgupta-db, line 132: Can we add 1 more test to reload the window and see if the NEW_PROFILE is picked up this time.

packages/databricks-vscode/src/test/e2e/wdio.conf.ts
```diff
@@ -117,7 +117,7 @@ export const config: Options.Testrunner = {
         return [
             {
                 "browserName": "vscode",
-                "browserVersion": engines.vscode.replace("^", ""),
+                "browserVersion": "stable",
```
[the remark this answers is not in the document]
ilia-db: Due to chromedriver issues we can't actually test the minimal supported version while using the `6.0.0` version of `wdio-vscode-service`. We can work around `5.*` issues and don't update, but I'm not sure it's worth it. 

Testing on stable makes sense since that's what our users use. But I agree that it's useful to test the minimal supported version. I'll add a test-matrix on the github actions level for that in a separate PR, but I think I'll just use the minimal vscode version that works with the new wdio service (1.86, instead of our 1.83)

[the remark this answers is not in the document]
kartikgupta-db: Let's bump our minimum supported version to 1.86 as well. We are anyway pitching this as the v2. Might as well take it to the highest version.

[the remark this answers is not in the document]
ilia-db: Yeah, that works too, although it still makes sense to run test under the moving "stable" target.

kartikgupta-db (approved):

PR #1076, opened 2024-02-19
Redirect summarise and validate errors to bundle output panel

kartikgupta-db:
## Changes
image

Uploading Screen Recording 2024-02-19 at 14.05.03.mov…


packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -250,8 +250,21 @@ export class ConnectionManager implements Disposable {
...
+                        if (choice === "Show Logs") {
+                            commands.executeCommand(
+                                "databricks.bundle.showLogs"
+                            );
+                        }
+                    });
+            } else if (e instanceof Error) {
+                window.showErrorMessage(
                     `Can't connect to the workspace: "${e.message}"."`
                 );
```
ilia-db, line 269: We can also offer to show the logs here, pointing to the full "Databricks Logs" output channel. 

Related: might be useful to create a wrapper for the `showErrorMessageWithLogs` logic, since this is very common pattern in the code-base

packages/databricks-vscode/src/logger/LoggerManager.ts
```diff
@@ -25,40 +32,49 @@ export class LoggerManager {
...
 
     async initLoggers() {
-        const outputChannel = window.createOutputChannel("Databricks Logs");
-        outputChannel.clear();
-
-        const sdkAndExtensionLogfile =
-            await this.getLogFile("sdk-and-extension");
+        const commonLogTransports = [
+            ...LOG_OUTPUT_CHANNEL_LEVELS.map(
+                (level) =>
```
ilia-db, line 49: Why do we create different transports for each level? Might be good to add a comment with an explanation here

ilia-db (approved):

packages/databricks-vscode/src/logger/LoggerManager.ts
kartikgupta-db: We need this because the output console requires different functions to print different levels of log. Since it is called from the stream, which does not get the log level as input ever, we need to specify the log level of the stream at object initialisation time.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
kartikgupta-db: I will take this in a separate PR?

PR #1105, opened 2024-03-04
Add telemetry for the v2

ilia-db:
- PROJECT_INITIALIZATION with "dabs" | "legacy" | "unknown" types
- LOGIN_WITH_SAVED_AUTH with "init" | "hostChange" | "targetChange" sources

### DIFF
--- a/packages/databricks-vscode/src/bundle/BundleProjectManager.ts
+++ b/packages/databricks-vscode/src/bundle/BundleProjectManager.ts
@@ -17,6 +17,7 @@ import {ProjectConfigFile} from "../file-managers/ProjectConfigFile";
 import {randomUUID} from "crypto";
 import {onError} from "../utils/onErrorDecorator";
 import {BundleInitWizard, promptToOpenSubProjects} from "./BundleInitWizard";
+import {Events, Telemetry} from "../telemetry";
 
 export class BundleProjectManager {
     private logger = logging.NamedLogger.getOrCreate(Loggers.Extension);
@@ -42,7 +43,8 @@ export class BundleProjectManager {
         private connectionManager: ConnectionManager,
         private configModel: ConfigModel,
         private bundleFileSet: BundleFileSet,
-        private workspaceUri: Uri
+        private workspaceUri: Uri,
+        private telemetry: Telemetry
     ) {
         this.disposables.push(
             this.bundleFileSet.bundleDataCache.onDidChange(async () => {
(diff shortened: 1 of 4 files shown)


packages/databricks-vscode/src/telemetry/constants.ts
```diff
@@ -10,13 +10,21 @@ export const DEV_APP_INSIGHTS_CONFIGURATION_STRING =
...
+    MANUAL_MIGRATION = "manualMigration",
+    BUNDLE_RUN = "bundleRun",
     CONNECTION_STATE_CHANGED = "connectionStateChanged",
-    SYNC_DESTINATION = "syncDestination",
-    SWITCH_TO_WORKSPACE_PROMPT = "switchToWorkspacePrompt",
 }
 /* eslint-enable @typescript-eslint/naming-convention */
 
+export type AutoLoginSource = "init" | "hostChange" | "targetChange";
+export type ManualLoginSource = "authTypeSwitch" | "authTypeLogin" | "command";
```
kartikgupta-db, line 25: Bundle init is another source

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
```diff
@@ -126,9 +120,18 @@ export class BundleProjectManager {
...
+        );
+        try {
+            await this.configModel.init();
+            await this.connectionManager.init();
+            this.projectServicesReady = true;
+            recordEvent({success: true, type: "dabs"});
+        } catch (e) {
+            recordEvent({success: false, type: "dabs"});
+            throw e;
+        }
```
kartikgupta-db, line 134: This seems like a common pattern and something that can be abstracted and better wrapped in a try-catch-finally block. We can do this in a later PR.

packages/databricks-vscode/src/bundle/BundleProjectManager.ts
```diff
@@ -154,14 +157,20 @@ export class BundleProjectManager {
     private async detectLegacyProjectConfig() {
         this.legacyProjectConfig = await this.loadLegacyProjectConfig();
         if (!this.legacyProjectConfig) {
+            this.customWhenContext.setPendingManualMigration(true);
```
kartikgupta-db, line 160: Nice catch!

kartikgupta-db (approved):

packages/databricks-vscode/src/telemetry/constants.ts
ilia-db: Or manual migration. The manual login actions there are different though - they don't lead to logged-in extension state (and don't capture `ConnectionManager.connect` flow), that's why I didn't include them. We can add lower level event for LoginWizardExecution with more sources later.

But that reminds me that I forgot to add telemetry for the bundle-init flow at all, gonna add it

PR #1112, opened 2024-03-05
Add DATABRICKS_CLI_UPSTREAM env var to CLI calls

ilia-db:
Also add DATABRICKS_CLI_UPSTREAM_VERSION

### DIFF
--- a/packages/databricks-vscode/src/cli/CliWrapper.test.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.test.ts
@@ -18,9 +18,14 @@ import crypto from "node:crypto";
 import {Context} from "@databricks/databricks-sdk/dist/context";
 import {logging} from "@databricks/databricks-sdk";
 import {LoggerManager} from "../logger";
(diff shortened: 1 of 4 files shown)


packages/databricks-vscode/src/cli/CliWrapper.test.ts
```diff
@@ -204,4 +209,50 @@ nothing = true
...
+                shell: true,
+            },
+        };
+        try {
+            assert.ok(isMatch(runCmd, expected));
+        } catch (e) {
+            // Run this in the "catch" case to show better error messages
+            assert.deepStrictEqual(runCmd, expected);
+            throw e;
+        }
```
kartikgupta-db, line 256: why use `isMatch` at all?

ilia-db: I just don't want the tests to fail if we add more env vars, doing soft isMatch will prevent that. But yeah, the benefits are questionable to be honest

PR #1114, opened 2024-03-05
Pass CLI path to all auth providers and login only once during bundle init

kartikgupta-db:
## Changes
* Store overrides in a per target file instead of state store
* Have a root overrides file for cases when we are not sure of the target. 
* Whenever a target is selected, and it does not already have an override file, it will MOVE the root override file to initialise it's own state json file.

### DIFF
--- a/packages/databricks-vscode/src/bundle/BundleInitWizard.ts
+++ b/packages/databricks-vscode/src/bundle/BundleInitWizard.ts
@@ -14,9 +14,13 @@ import {CliWrapper} from "../cli/CliWrapper";
 import {getSubProjects} from "./BundleFileSet";
 import {tmpdir} from "os";
 import {ShellUtils} from "../utils";
+import {OverrideableConfigModel} from "../configuration/models/OverrideableConfigModel";
+import {writeFile, mkdir} from "fs/promises";
+import path from "path";
 
 export async function promptToOpenSubProjects(
-    projects: {absolute: Uri; relative: Uri}[]
+    projects: {absolute: Uri; relative: Uri}[],
+    authProvider?: AuthProvider
 ) {
     type OpenProjectItem = QuickPickItem & {uri?: Uri};
     const items: OpenProjectItem[] = projects.map((project) => {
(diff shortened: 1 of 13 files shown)


packages/databricks-vscode/src/bundle/BundleInitWizard.ts
```diff
@@ -34,9 +38,21 @@ export async function promptToOpenSubProjects(
...
+        const rootOverrideFilePath =
+            OverrideableConfigModel.getRootOverrideFile(item.uri);
+        await mkdir(path.dirname(rootOverrideFilePath.fsPath), {
+            recursive: true,
+        });
+        await writeFile(
+            rootOverrideFilePath.fsPath,
+            JSON.stringify({authProfile: authProvider.toJSON().profile})
+        );
+    }
```
ilia-db, line 55: Maybe move this logic to the OverrideableModel?

Although I'm not even sure we need a concept of the "root" override file (at least not yet). In the "normal" scenario with only one profile per host we won't really need this override, right? After opening that folder we will log-in automatically. Having all this logic around root override seems like a lot of effort for not much benefit, unless we already plan to use it for a bunch of other things.

packages/databricks-vscode/src/configuration/auth/AuthProvider.ts
```diff
@@ -151,33 +149,27 @@ export abstract class AuthProvider {
...
             case "databricks-cli":
-                if (!config.databricksCliPath) {
-                    throw new Error("Missing path for databricks-cli");
-                }
-
-                return new DatabricksCliAuthProvider(
-                    host,
-                    config.databricksCliPath
-                );
+                return new DatabricksCliAuthProvider(host, cli);
```
ilia-db, line 152: With so many changes around passing cli to auth providers I'm not sure I can actually understand what exactly we are improving, can you explain it a bit?

kartikgupta-db: If the profile does not have a `databricksclipath` variable, we would error out, https://github.com/databricks/databricks-vscode/pull/1114/files#diff-3e8d1317ab3ec8502ceebeae3f641a4e8f7388e442dc8804cdfdb5ea2db283b8L154-L156. 

But, we can be using the internal databricks cli instead of relying on the profile to have the path. It is quiet a common profile pattern to not hard code the cli path. In fact, we should not be writing the cli path either because the profile will fail when we update the extension (hence breaking the path).

kartikgupta-db: This was breaking oauth for users with existing profiles.

packages/databricks-vscode/src/bundle/BundleInitWizard.ts
kartikgupta-db: We don't have a way to enforce 1 profile per host (and should not because we can have multiple auth methods). A lot of bricksfooders ended up creating multiple profiles for the same host when initializing multiple projects. The extension would then be in a login state until they manually reselected their already created profiles. 

This allows us to establish some communication method between the pre bundle init extension instance and post bundle init extension instance. Right now it is just this, but in future with env management, I can imagine having to communicate that information, or cluster information as well.

ilia-db (approved):

PR #1125, opened 2024-03-07
Cache list profiles output based on last modified time of .databrickscfg

kartikgupta-db:
## Changes
* Also show  'Multiple login profiles available. Click to select a profile.' instead of "login to Databricks" when multiple profiles are available.
* Hide the login button when no target is selected.

### DIFF
--- a/packages/databricks-vscode/src/cli/CliWrapper.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.ts
@@ -17,6 +17,7 @@ import {AuthProvider} from "../configuration/auth/AuthProvider";
 import {removeUndefinedKeys} from "../utils/envVarGenerators";
 import {quote} from "shell-quote";
 import {stat} from "fs/promises";
+import {Stats} from "fs";
 
 const withLogContext = logging.withLogContext;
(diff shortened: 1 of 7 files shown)


packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -169,19 +174,37 @@ export class CliWrapper {
...
+        configfilePath: string,
         @context ctx?: Context
-    ): Promise<Array<ConfigEntry>> {
+    ): Promise<Array<ProfileDetails>> {
+        let stats: Stats;
+        try {
+            stats = await stat(configfilePath);
+            if (
+                stats.mtimeMs <=
+                (this.databricksCfgCache.get(configfilePath)?.lastModified ?? 0)
```
ilia-db, line 185: Do `mv` or `cp` commands change modification timestamps of the files? And what about windows counterparts?

kartikgupta-db: Mv and cp do change the mtime. 

For windows, not sure, but this gets us some performance increase. I will do a more thorough ananlysis in a later PR.

PR #1147, opened 2024-03-19
Bundle cli dependencies

ilia-db:
Bundle terraform binaries with the extension and expose them with env vars to the CLI and the terminal

Depends on https://github.com/databricks/cli/pull/1294


packages/databricks-vscode/src/extension.ts
```diff
@@ -165,11 +166,19 @@ export async function activate(
     // Add the databricks binary to the PATH environment variable in terminals
     context.environmentVariableCollection.clear();
     context.environmentVariableCollection.persistent = false;
-    context.environmentVariableCollection.append(
+    context.environmentVariableCollection.prepend(
         "PATH",
-        `${path.delimiter}${context.asAbsolutePath("./bin")}`
+        `${context.asAbsolutePath("./bin")}${path.delimiter}`
```
kartikgupta-db, line 171: But append seems to have started to work for me now :)

kartikgupta-db (approved):

packages/databricks-vscode/src/extension.ts
```diff
@@ -170,6 +179,23 @@ export async function activate(
         `${path.delimiter}${context.asAbsolutePath("./bin")}`
     );
 
+    const cliDeps = getCLIDependenciesEnvVars(context);
+    for (const [key, value] of Object.entries(cliDeps)) {
```
kartikgupta-db, line 183: The idea is to always use the packaged tf binaries? Can you add a comment for this?

ilia-db: Yes, we always use bundled dependencies, this approach seems to be the simples one. I've added the comment here

PR #1151, opened 2024-03-20
Ignore stderr for validate and summary commands

ilia-db:
We still log stderr, but don't pass it down to the callers, as it will mess up json parsing.

If the command fails with non zero status we also fail, but the presence of stderr is not an indicator of failure in this case.


lennartkats-db (approved):
👍

PR #1157, opened 2024-03-25
Add a interface for specifying bundle variables

kartikgupta-db:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -163,7 +163,7 @@ jobs:
         working-directory: packages/databricks-vscode
 
       - name: Upload test logs
-        if: always()
+        if: failure()
         continue-on-error: true
         uses: actions/upload-artifact@v3
         with:


packages/databricks-vscode/src/bundle/models/BundleVariableModel.ts
```diff
@@ -0,0 +1,236 @@
...
+    })
+    async deleteBundleVariableFile() {
+        if (this.bundleVariableFilePath === undefined) {
+            return;
+        }
+
+        try {
+            await workspace.fs.delete(this.bundleVariableFilePath);
+        } catch (e: any) {
+            if (e.code !== "ENOENT") {
```
ilia-db, line 231: `code` can also be `FileNotFound`

packages/databricks-vscode/src/ui/bundle-variables/VariableTreeNode.ts
```diff
@@ -0,0 +1,33 @@
...
+        return {
+            label: this.key,
+            description:
+                this.value.vscodeOverrideValue ?? this.value.valueInTarget,
+            collapsibleState: TreeItemCollapsibleState.Collapsed,
+        };
+    }
+
+    getChildren(): BundleVariableTreeNode[] {
+        if (this.value.valueInTarget !== undefined) {
```
ilia-db, line 23: It's actually possible to define global variables without descriptions, just with `var: value` as you do in targets. In that case we will get an expandable node, but it won't expand to anything

kartikgupta-db: The value is treaded as the default value in that case ig?

ilia-db: Ouch, I commented under the wrong line. I mean that there are cases when we should use `TreeItemCollapsibleState.None`, as there's nothing to expand to

ilia-db: But also if you define the global variable as simple `var: value`, we won't show the `value` in the tree item at all

packages/databricks-vscode/package.json
```diff
@@ -298,6 +310,12 @@
                     "visibility": "visible",
                     "when": "databricks.context.bundle.isTargetSet"
                 },
+                {
+                    "id": "dabsVariableView",
+                    "name": "Bundle Variable View",
+                    "visibility": "visible",
+                    "when": "databricks.context.bundle.isTargetSet"
```
ilia-db, line 317: Empty variable view is confusing right now, especially since it lets you "set bundle variables", but nothing will actually show up untill you define the actual variables in the `databricks.yml`. Let's hide it when there's not variables?

packages/databricks-vscode/package.json
```diff
@@ -298,6 +310,12 @@
                     "visibility": "visible",
                     "when": "databricks.context.bundle.isTargetSet"
                 },
+                {
+                    "id": "dabsVariableView",
+                    "name": "Bundle Variable View",
```
ilia-db, line 315:
```suggestion
                    "name": "Bundle Variables View",
```

packages/databricks-vscode/package.json
```diff
@@ -275,6 +275,18 @@
...
+            },
+            {
+                "command": "databricks.bundle.variable.openFile",
+                "title": "Set bundle variables",
+                "enablement": "databricks.context.activated && databricks.context.bundle.isTargetSet",
+                "category": "Databricks"
+            },
+            {
+                "command": "databricks.bundle.variable.reset",
+                "title": "Reset to default value",
```
ilia-db, line 287:
```suggestion
                "title": "Reset to default values",
```

packages/databricks-vscode/package.json
```diff
@@ -275,6 +275,18 @@
                 "title": "Copy",
                 "enablement": "databricks.context.activated",
                 "category": "Databricks"
+            },
+            {
+                "command": "databricks.bundle.variable.openFile",
+                "title": "Set bundle variables",
```
ilia-db, line 281: The name can be a bit confusing, I would expect to "set" the varialbes in the `databricks.yml`. Should it be "override"?

packages/databricks-vscode/src/bundle/models/BundleVariableModel.ts
```diff
@@ -0,0 +1,236 @@
...
+        if (target === undefined) {
+            return undefined;
+        }
+
+        return Uri.joinPath(
+            this.workspaceRoot,
+            ".databricks",
+            "bundle",
+            target,
+            "vscode.bundlevars"
```
ilia-db, line 91: Have we considered using `json`? Much easier to validate and show proper errors to the users when the format is incorrect. Right now we rely on try/catch block in the `getVariableOverrides`, but when the format is incorrect the error will not be user friendly

packages/databricks-vscode/src/bundle/models/BundleVariableModel.ts
```diff
@@ -0,0 +1,236 @@
...
+            const isRequired =
+                (inTargetVariable ??
+                    definition.lookup ??
+                    definition.default) === undefined;
+
+            variables[key] = {
+                ...definition,
+                lookup:
+                    typeof inTargetVariable === "object"
+                        ? inTargetVariable
```
ilia-db, line 156: Should it be `inTargetVariable.lookup`?

packages/databricks-vscode/src/ui/bundle-variables/LookupVariableTreeNode.ts
```diff
@@ -0,0 +1,37 @@
...
+export class LookupVariableTreeNode implements BundleVariableTreeNode {
+    constructor(
+        public readonly key: string,
+        public readonly value: BundleVariable,
+        public readonly parent?: BundleVariableTreeNode
+    ) {}
+
+    getTreeItem(): BundleVariableTreeItem {
+        const lookup = this.value.lookup as any;
+        const lookupValue = Object.values(lookup).at(0) as string | undefined;
```
ilia-db, line 15: Why are we not checking the first non undefined value instead? Similar to how we check the key in the `getChildren`.

packages/databricks-vscode/src/bundle/models/BundleVariableModel.ts
kartikgupta-db: No, because when defining new lookups inside a target, users don't have to specify the lookup field.

ilia-db (approved):

PR #1159, opened 2024-03-26
Add currently selected python environment to path for wheel builds

kartikgupta-db:
## Changes
* Deploying jobs with wheels fails because we are not propagating the selected virtual env to bundle commands. 
* This PR fixes that by adding the path to virtual environment python binary to `PATH` env var for bundle deploy.

### DIFF
--- a/packages/databricks-vscode/src/cli/CliWrapper.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.ts
@@ -415,7 +415,8 @@ export class CliWrapper {
 
         // Add python executable to PATH
         const executable = await pythonExtension.getPythonExecutable();
-        let shellPath = process.env.PATH;
+        const cliEnvVars = EnvVarGenerators.getEnvVarsForCli(configfilePath);
+        let shellPath = cliEnvVars.PATH;
         if (executable) {
             shellPath = `${path.dirname(executable)}${path.delimiter}${
                 process.env.PATH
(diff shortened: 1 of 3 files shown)


packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -392,6 +394,7 @@ export class CliWrapper {
         target: string,
         authProvider: AuthProvider,
         workspaceFolder: Uri,
+        pythonExtension: MsPythonExtensionWrapper,
```
ilia-db, line 397: Is the problem with the `deploy` only? Should we propagate the correct PATH to other bundle commands (like `run`)?

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -410,15 +413,25 @@ export class CliWrapper {
             bundleOpName: "deploy",
         });
 
+        // Add python executable to PATH
+        const executable = await pythonExtension.getPythonExecutable();
+        let shellPath = process.env.PATH;
+        if (executable) {
+            shellPath = `${path.dirname(executable)}${path.delimiter}${
+                process.env.PATH
```
ilia-db, line 421: Might be better to use the PATH from the getEnvVarsForCli. Currently it's the same, but if we ever update it that function, this logic will override the changes

packages/databricks-vscode/src/cli/CliWrapper.ts
kartikgupta-db: Yeah, it is deploy only for now.

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -410,15 +413,26 @@ export class CliWrapper {
             bundleOpName: "deploy",
         });
 
+        // Add python executable to PATH
+        const executable = await pythonExtension.getPythonExecutable();
+        const cliEnvVars = EnvVarGenerators.getEnvVarsForCli(configfilePath);
+        let shellPath = cliEnvVars.PATH;
+        if (executable) {
+            shellPath = `${path.dirname(executable)}${path.delimiter}${
+                process.env.PATH
```
ilia-db, line 422:
```suggestion
                shellPath
```

ilia-db (approved):

PR #1160, opened 2024-03-26
Fixes to initialisation UI

kartikgupta-db:
### Welcome view changes
* Always show button to create a databricks project.
* Don't automatically migrate from old project.json if there are subprojects present. 
* Reword welcome view buttons and text
image

### Login Changes
* Don't logout if refreshing one of the models fails. This means users can see the error, fix it and have their fixes automatically be picked up and tried out. 
* Show "Multiple login profiles available. Click to select a profile." when multiple profiles found in .databrickscfg.

### DIFF
--- a/packages/databricks-vscode/src/configuration/ConnectionManager.ts
+++ b/packages/databricks-vscode/src/configuration/ConnectionManager.ts
@@ -293,7 +293,6 @@ export class ConnectionManager implements Disposable {
                     `Can't connect to the workspace: "${e.message}"."`
                 );
             }
-            await this.logout();
         }
     }
 
@@ -309,11 +308,15 @@ export class ConnectionManager implements Disposable {
             "authProfile",
             authProvider.toJSON().profile as string | undefined
         );
-        await this.configModel.setAuthProvider(authProvider);
+
         await this.updateSyncDestinationMapper();
         await this.updateClusterManager();
         await this._metadataService.setApiClient(this.apiClient);
-        this.updateState("CONNECTED");
+        try {
+            await this.configModel.setAuthProvider(authProvider);
+        } finally {
+            this.updateState("CONNECTED");
+        }
     }
 
     @Mutex.synchronise("loginLogoutMutex")
--- a/packages/databricks-vscode/src/configuration/ConnectionManager.ts
+++ b/packages/databricks-vscode/src/configuration/ConnectionManager.ts
@@ -5,8 +5,8 @@ import {
     AuthType as SdkAuthType,
 } from "@databricks/databricks-sdk";
 import {Cluster} from "../sdk-extensions";
-import {EventEmitter, Uri, window, Disposable, commands} from "vscode";
-import {CliWrapper, ProcessError} from "../cli/CliWrapper";
+import {EventEmitter, Uri, window, Disposable} from "vscode";
+import {CliWrapper} from "../cli/CliWrapper";
 import {
     SyncDestinationMapper,
     RemoteUri,
@@ -275,20 +275,7 @@ export class ConnectionManager implements Disposable {
                 `Can't connect to the workspace`,
                 e
             );
-            if (e instanceof ProcessError) {
-                window
-                    .showErrorMessage(
-                        `Can't connect to the workspace. Error executing Databricks CLI command.`,
-                        "Show Logs"
-                    )
-                    .then((choice) => {
-                        if (choice === "Show Logs") {
-                            commands.executeCommand(
-                                "databricks.bundle.showLogs"
-                            );
-                        }
-                    });
-            } else if (e instanceof Error) {
+            if (e instanceof Error) {
                 window.showErrorMessage(
                     `Can't connect to the workspace: "${e.message}"."`
                 );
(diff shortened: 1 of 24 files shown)


packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -275,25 +273,11 @@ export class ConnectionManager implements Disposable {
...
-                    )
-                    .then((choice) => {
-                        if (choice === "Show Logs") {
-                            commands.executeCommand(
-                                "databricks.bundle.showLogs"
-                            );
-                        }
-                    });
-            } else if (e instanceof Error) {
+            if (e instanceof Error) {
```
ilia-db, line 276: With `allSettled` we are swallowing normal errors right now, we should re-throw them in setTarget/setAuthProvider methods.

But I'm not sure why should we move error handling over there? Seems easier to just have `Promise.all` and catch everything here. Plus we need to add error handling to `ConnectionCommands.selectTarget` too.

packages/databricks-vscode/src/configuration/ui/ClusterComponent.ts
```diff
@@ -105,6 +105,10 @@ export class ClusterComponent extends BaseComponent {
                         new ThemeColor("notificationsErrorIcon.foreground")
                     ),
                     id: TREE_ICON_ID,
+                    command: {
```
ilia-db, line 108: Let's remove the command from the `package.json->view/item/context` then?

packages/databricks-vscode/src/configuration/ui/AuthTypeComponent.ts
```diff
@@ -41,10 +48,20 @@ export class AuthTypeComponent extends BaseComponent {
         }
 
         if (authProvider === undefined) {
-            const label = "Login to Databricks";
+            const host = await this.configModel.get("host");
+            if (host === undefined) {
+                return [];
```
ilia-db, line 53: If the bundle doesn't have a host defined we show a "failed to initialize target" notification right now, and then end up in a broken state when you can't really switch target anymore or even trigger the error again. 

At least here it would make sense to show a message about the absent host and prompt users to open and update the dabs config

packages/databricks-vscode/src/configuration/models/ConfigModel.ts
```diff
@@ -214,12 +215,32 @@ export class ConfigModel implements Disposable {
             // We want to wait for all the configs to be loaded before we emit any change events from the
             // configStateCache.
             await this.readStateMutex.synchronise(async () => {
-                await Promise.all([
+                await Promise.allSettled([
```
ilia-db, line 218: Is it intentional that we never fail here anymore and always proceed to the `setAuthProvider`? `setAuthProvider` doesn't fail too with `allSettled`, so we are swallowing non-ProcessError errors now, breaking some of the higher-level error handling (in the ConnectionManager and ConnectionCommands)

kartikgupta-db: Yes.  The idea is to not block login based on errors here, and also actually run all the promises parallely. I agree that we should show error messages for other errors as well. 

Right now, when any of the models fail to refresh, we completely logout, and user has to log back in. With this change, the user can make changes and fixes to their bundle, which will automatically refresh the models.

packages/databricks-vscode/src/configuration/ui/AuthTypeComponent.ts
kartikgupta-db: Fixed this state as well. Now we show a message saying "host is misconfigured". And once user fixes it, changes are picked up and processed automatically.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
kartikgupta-db: Removed all settled. I split setting target/auth provider and doing state refresh into 2 separate steps. Now, auth provider and target will still be updated in all models, even if the login fails.

packages/databricks-vscode/src/configuration/models/ConfigModel.ts
```diff
@@ -208,35 +208,42 @@ export class ConfigModel implements Disposable {
...
+            await this.configsMutex.synchronise(async () => {
+                this._target = target;
+                await this.stateStorage.set("databricks.bundle.target", target);
+                // We want to wait for all the configs to be loaded before we emit any change events from the
+                // configStateCache.
+                this.bundlePreValidateModel.setTarget(target);
+                this.bundleValidateModel.setTarget(target);
+                this.overrideableConfigModel.setTarget(target);
+                this.bundleRemoteStateModel.setTarget(target);
+                this.onDidChangeTargetEmitter.fire();
```
ilia-db, line 222: I wonder if we now have a race condition because you emit the target change before refreshing the models. During the refresh the connection manager tries to login with saved auth and might use stale values from the configModel (since its refresh it tied to the refreshes of other models, which might not happen in time)

ilia-db: Should we emit the target change at the very end of the method?

PR #1171, opened 2024-04-05
Allow opting out of cluster override and make it more visible in UI

kartikgupta-db:


packages/databricks-vscode/package.json
```diff
@@ -6,7 +6,7 @@
     "license": "LicenseRef-LICENSE",
     "version": "2.0.2",
     "engines": {
-        "vscode": "^1.86.0"
+        "vscode": "^1.88.0"
```
ilia-db, line 9: Yesterday we've got a feedback from a customer that vscode version requirements can be confusing. When you try to install extension using usupported vscode version it doesn't really show any errors or messages - it seems like the extension is installed, but just doesn't work.

One way to minimize the confusion is to have the requirement clearly stated in docs, but another one is to relax the requirements here. Can we do that, or we actually need `1.88.0` version?

packages/databricks-vscode/src/ui/configuration-view/ConfigurationTreeViewManager.ts
```diff
@@ -0,0 +1,38 @@
...
+        readonly treeView: TreeView<ConfigurationTreeItem>,
+        readonly configModel: ConfigModel
+    ) {
+        this.disposables.push(
+            treeView.onDidChangeCheckboxState(
+                async (e: TreeCheckboxChangeEvent<ConfigurationTreeItem>) => {
+                    await Promise.all(
+                        e.items.map(async ([item, state]) => {
+                            if (item.id === CLUSTER_OVERRIDE_CHECKBOX_ID) {
+                                await this.configModel.set(
```
ilia-db, line 23: Why can't we do it in the ClusterComponent directly and not rely on treeView-level change events? E.g. execute a command each time the checkbox is changed

kartikgupta-db: The only way to listen to changes to builtin check box is through tree view. We will have to implement our own check boxes otherwise.

packages/databricks-vscode/package.json
kartikgupta-db: We need this for supporting checkbox change events.

ilia-db (approved):

PR #1250, opened 2024-06-19
Add support for multiple vscode workspaces

kartikgupta-db:

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -544,7 +544,7 @@
                 {
                     "submenu": "databricks.run",
                     "group": "navigation@0",
-                    "when": "resourceLangId == python || resourceLangId == scala || resourceLangId == r || resourceLangId == sql || resourceExtname == .ipynb"
+                    "when": "databricks.context.isActiveFileInActiveWorkspace && resourceLangId =~ /^(python|scala|r|sql)$/ || databricks.context.isActiveFileInActiveWorkspace && resourceExtname == .ipynb"
                 }
             ],
             "databricks.run": [
@@ -881,7 +881,7 @@
         "useYarn": false
     },
     "cli": {
-        "version": "0.219.0"
+        "version": "0.221.0"
     },
     "scripts": {
         "vscode:prepublish": "rm -rf out && yarn run package:compile && yarn run package:wrappers:write && yarn run package:jupyter-init-script:write && yarn run package:copy-webview-toolkit && yarn run generate-telemetry",
--- a/packages/databricks-vscode/src/configuration/models/OverrideableConfigModel.ts
+++ b/packages/databricks-vscode/src/configuration/models/OverrideableConfigModel.ts
@@ -66,7 +66,6 @@ export class OverrideableConfigModel extends BaseModelWithStateCache<Overrideabl
         private readonly workspaceFolderManager: WorkspaceFolderManager
     ) {
         super();
-        this.disposables.push();
     }
 
     public setTarget(target: string | undefined) {
--- a/packages/databricks-vscode/src/configuration/models/OverrideableConfigModel.ts
+++ b/packages/databricks-vscode/src/configuration/models/OverrideableConfigModel.ts
@@ -4,6 +4,7 @@ import {BaseModelWithStateCache} from "./BaseModelWithStateCache";
 import {Uri} from "vscode";
 import path from "path";
 import {existsSync} from "fs";
+import {WorkspaceFolderManager} from "../../vscode-objs/WorkspaceFolderManager";
 
 export type OverrideableConfigState = {
     authProfile?: string;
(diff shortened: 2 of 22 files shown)


packages/databricks-vscode/package.json
```diff
@@ -544,7 +544,7 @@
                 {
                     "submenu": "databricks.run",
                     "group": "navigation@0",
-                    "when": "resourceLangId == python || resourceLangId == scala || resourceLangId == r || resourceLangId == sql || resourceExtname == .ipynb"
+                    "when": "databricks.context.isActiveFileInActiveWorkspace && resourceLangId =~ /^(python|scala|r|sql)$/ || databricks.context.isActiveFileInActiveWorkspace && resourceExtname == .ipynb"
```
ilia-db, line 547:
```suggestion
                    "when": "databricks.context.isActiveFileInActiveWorkspace && (resourceLangId =~ /^(python|scala|r|sql)$/ || resourceExtname == .ipynb)"
```

packages/databricks-vscode/src/extension.ts
```diff
@@ -847,19 +868,25 @@ export async function activate(
...
+            .catch((e) => {
+                recordInitializationEvent({success: false});
+                logging.NamedLogger.getOrCreate(Loggers.Extension).error(
+                    "Failed to configure workspace",
+                    e
+                );
+                window.showErrorMessage(e);
+            })
+            .finally(() => {
+                customWhenContext.setInitialized();
```
ilia-db, line 884: We should reset the context to false when we start initialising in a new folder - to show `initializing...` welcome view

ilia-db: Or it won't get shown anyway as we don't reset the tree view providers?

packages/databricks-vscode/src/configuration/models/OverrideableConfigModel.ts
```diff
@@ -57,8 +58,15 @@ export class OverrideableConfigModel extends BaseModelWithStateCache<Overrideabl
...
-    constructor(private readonly workspaceRoot: Uri) {
+    get workspaceRoot() {
+        return this.workspaceFolderManager.activeWorkspaceFolder.uri;
+    }
+
+    constructor(
+        private readonly workspaceFolderManager: WorkspaceFolderManager
+    ) {
         super();
+        this.disposables.push();
```
ilia-db, line 69: :suspect:

packages/databricks-vscode/src/extension.ts
```diff
@@ -173,7 +174,17 @@ export async function activate(
...
     }
 
-    const workspaceUri = workspace.workspaceFolders[0].uri;
+    const workspaceFolderManager = new WorkspaceFolderManager(
+        customWhenContext
+    );
+
+    context.subscriptions.push(
+        telemetry.registerCommand(
+            "databricks.selectWorkspaceFolder",
```
ilia-db, line 183: We can also add it to package.json and expose it in the command palette UI

packages/databricks-vscode/package.json
kartikgupta-db: Brackets don't seem to work :/

packages/databricks-vscode/src/extension.ts
kartikgupta-db: Yeah. It will not show. Also, seems like we forget the target when we switch from a bad project (without databricks yaml for example), to a good one. Not fixing that edge case.

packages/databricks-vscode/src/language/ConfigureAutocomplete.ts
```diff
@@ -43,7 +44,7 @@ export class ConfigureAutocomplete implements Disposable {
     constructor(
         private readonly context: ExtensionContext,
         private readonly stateStorage: StateStorage,
-        private readonly workspaceFolder: string,
+        private readonly workspaceFolderManager: WorkspaceFolderManager,
```
ilia-db, line 47: Should we listen to changes here an re-run `configure`?

kartikgupta-db: Configure autocomplete has always been run on demand (by running the command). I do not want to change that behaviour in this PR.

packages/databricks-vscode/src/configuration/models/OverrideableConfigModel.ts
kartikgupta-db: Removed

packages/databricks-vscode/src/extension.ts
kartikgupta-db: I want to keep it hidden so that users don't mess up their projects.

packages/databricks-vscode/src/language/ConfigureAutocomplete.ts
ilia-db: Hmm, that `configure` call in the constructor is confusing, but I see that it doesn't do anything

ilia-db (approved):

PR #1277, opened 2024-07-08
Show correct default value for variables in bundle variable UI

### LINKED ISSUE #1272: [BUG] Can't reset variable overrides
**Describe the bug**
As soon as you edit a `bundlevars` json, both the current value and the default value are set to the override, with the `reset` action doing nothing.

**To Reproduce**
- Create an example python dabs project
- Add this var to the config:
```
variables:
  test:
    description: Test
    default: hello
```
- In the variables explorer click on the cog wheel
- Change the value from `hello` to `123`
- Observe that the Default value also changed  to `123`
Screenshot 2024-07-03 at 13 55 51

- Click on `reset` icon and observe that nothing has actually been reseted.

**System information:**

1. MacOS
2. Databricks Extension Version: v2.2.2

**Additional context**
Seems like the override ends up in the `preValidateConfig`, which we don't reset when we delete the bundlevars json

kartikgupta-db:
## Changes
Fixes #1272

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
@@ -16,13 +16,13 @@ import {BundleSchema} from "../../bundle/types.ts";
 describe("Bundle Variables", async function () {
     let workbench: Workbench;
     let vscodeWorkspaceRoot: string;
-    let clusterId: string;
     let schemaDef: BundleSchema;
 
     this.timeout(3 * 60 * 1000);
 
     async function createProjectWithJob() {
         const projectName = getUniqueResourceName("bundle_variables");
+        /* eslint-disable @typescript-eslint/naming-convention */
         schemaDef = getBasicBundleConfig({
             bundle: {
                 name: projectName,
@@ -48,10 +48,6 @@ describe("Bundle Variables", async function () {
 
     before(async function () {
         assert(
-            process.env.TEST_DEFAULT_CLUSTER_ID,
-            "TEST_DEFAULT_CLUSTER_ID env var doesn't exist"
-        );
-        assert(
             process.env.WORKSPACE_PATH,
             "WORKSPACE_PATH env var doesn't exist"
         );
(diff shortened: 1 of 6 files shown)


ilia-db (commented):
Can we add a few tests that cover variables? Can be e2e, or just unit tests that cover important edge cases (initial load, updating the view when config changes, updating the view when overrides change, updating the view when we reset overrides).

packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
```diff
@@ -0,0 +1,217 @@
...
+    ) {
+        await section.expand();
+        const variableTreeItem = await section.findItem(variableName);
+        assert(variableTreeItem);
+        console.log(
+            await variableTreeItem.getDescription(),
+            expected.value,
+            (await variableTreeItem.getDescription()) === expected.value
+        );
+        assert((await variableTreeItem.getDescription()) === expected.value);
```
ilia-db, line 93: Better to use `assert.strictEqual` here (and in similar cases below), instead of simple assert with `===` comparison inside. The `strictEqual` will print all the arguments in case it fails

packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    it("should wait for connection", async () => {
+        await waitForLogin("DEFAULT");
+        await dismissNotifications();
+    });
+
+    async function assertVariableValue(
+        section: CustomTreeSection,
+        variableName: string,
+        expected: {value?: string; defaultValue?: string; lookup?: string}
```
ilia-db, line 83: Do we use `lookup` anywhere?

kartikgupta-db: Nope. I intended to write tests for this, but got too tedious. Will remove for now.

ilia-db (approved):

PR #1278, opened 2024-07-08
Bring back sync functionality

kartikgupta-db:
## Changes
* Bring back sync functionality. Users can choose to manually start sync from the UI and not rely on deploy before file runs. This avoids breaking some existing user flows. 
* The files are still run only after a deploy.
* There is possibility of race conditions between sync and deploy but the possibility is very small. Also, syncing is blocked for prod and staging targets.

### DIFF
--- a/packages/databricks-vscode/src/configuration/models/ConfigModel.ts
+++ b/packages/databricks-vscode/src/configuration/models/ConfigModel.ts
@@ -147,6 +147,11 @@ export class ConfigModel implements Disposable {
             ),
             this.bundleRemoteStateModel.onDidChange(async () => {
                 await this.configCache.refresh();
+            }),
+            this.onDidChangeKey("mode")(async () => {
+                this.vscodeWhenContext.isDevTarget(
+                    (await this.configCache.value).mode === "development"
+                );
             })
         );
     }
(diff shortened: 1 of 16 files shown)


packages/databricks-vscode/src/configuration/models/ConfigModel.ts
```diff
@@ -154,6 +154,13 @@ export class ConfigModel implements Disposable {
     @onError({popup: true})
     public async init() {
         await this.readTarget();
+        this.disposables.push(
```
ilia-db, line 157: Why not in the constructor were we register other event handlers?

packages/databricks-vscode/src/sync/CodeSynchronizer.ts
```diff
@@ -0,0 +1,145 @@
...
+    get isRunning() {
+        return (
+            this.state === "IN_PROGRESS" ||
+            this.state === "WATCHING_FOR_CHANGES"
+        );
+    }
+    // This function waits for sync to reach WATCHING_FOR_CHANGES which is a
+    // necessary condition to execute local code on databricks. This state denotes
+    // all local changes have been synced to remote workspace
+    async waitForSyncComplete(): Promise<void> {
```
ilia-db, line 124: Do we want to use somewhere? Might as well delete it, if not

ilia-db (commented):
Let's also bring back e2e tests for the sync logic? We had implicit tests for it in the `run*` e2e suites, but right now we can add something more explicit

ilia-db (approved):

PR #1292, opened 2024-07-19
Cleanup redundant commands and fix errors when resetting bundle variables

kartikgupta-db:


fjakobs (approved):

packages/databricks-vscode/src/extension.ts
```diff
@@ -804,6 +815,20 @@ export async function activate(
...
             runCommands
         ),
+        window.onDidChangeActiveTextEditor(async (e) => {
+            const uri = e?.document.uri;
+
+            if (
+                uri &&
+                uri.scheme === "file" &&
+                ((await FileUtils.isNotebook(new LocalUri(uri))) ||
+                    uri.fsPath.endsWith(".py"))
```
ilia-db, line 825: Might be better to check `document.languageId`, will work with any docs that have relevant language mode enabled. We can still check if it's a real file though

PR #1295, opened 2024-07-23
Allow CLI path to have spaces

kartikgupta-db:


packages/databricks-vscode/src/configuration/ConnectionCommands.ts
```diff
@@ -119,7 +119,7 @@ export class ConnectionCommands implements Disposable {
     }
 
     attachClusterQuickPickCommand() {
-        return async (title?: string) => {
+        return async () => {
```
ilia-db, line 122: We still use the title in the `EnvironmentDependenciesVerifier->checkCluster`. We can remove the title from there too, but might as well leave it. Or is the `title` check here is not good enough and you are fixing a bug here by removing the argument?

packages/databricks-vscode/src/configuration/LoginWizard.ts
```diff
@@ -384,25 +385,36 @@ export async function saveNewProfile(
         throw new Error("Can't save empty auth provider to a profile");
     }
 
-    const {path: configFilePath} = await loadConfigFile(
-        workspaceConfigs.databrickscfgLocation
-    );
+    const configFilePath: string = getDatabricksConfigFilePath().fsPath;
+    let shouldBackup = true;
+    try {
+        await loadConfigFile(configFilePath);
```
ilia-db, line 391: Is the bug here in that we are failing to save the profile if the cfg file doesn't exist?

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -19,9 +19,42 @@ import {quote} from "shell-quote";
...
-const execFile = promisify(execFileCb);
+function getEscapedCommandAndAgrs(
+    cmd: string,
+    args: string[],
+    options: SpawnOptionsWithoutStdio
+) {
+    if (process.platform === "win32") {
+        args = [
+            "/d",
+            "/c",
```
ilia-db, line 33: What do those magic args do? Can you add a link to some docs to the code?

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -19,9 +19,42 @@ import {quote} from "shell-quote";
...
+            `""${cmd}" ${args.map((a) => `"${a}"`).join(" ")}"`,
+        ];
+        cmd = "cmd.exe";
```
ilia-db, line 36: Why are we changing the `cmd` here? In the `args` we will capture the old `cmd`, right? So it seems like this assignment should have any effects?

ilia-db: Oh wait, we return it later on, makes sense

packages/databricks-vscode/src/configuration/LoginWizard.ts
kartikgupta-db: Yep

packages/databricks-vscode/src/cli/CliWrapper.ts
kartikgupta-db: We want to use cmd.exe to run this new command. Sort of like `bash -c "command"`

packages/databricks-vscode/src/configuration/ConnectionCommands.ts
kartikgupta-db: I messed up when fixing merge conflicts.

packages/databricks-vscode/src/bundle/BundleInitWizard.ts
```diff
@@ -188,12 +188,12 @@ export class BundleInitWizard {
             "bundle",
             "init",
             "--output-dir",
-            this.cli.escapePathArgument(parentFolder.fsPath),
+            `"${parentFolder.fsPath}"`,
```
ilia-db, line 191: I think mac and linux can technically have `"` in the dir names, so escaping the path can still be useful

ilia-db (approved):

PR #1315, opened 2024-08-19
Allow cancelling login, fix spaces in path for dbconnect and minor UI fixes

kartikgupta-db:
## Changes
* Allow cancelling login
* Show confirmation prompt for dabs deploy in prod.
* Fix spaces in path name for dbconnect paths. 
* Rename "workspace" in UI elements such as button "Migrate current workspace to Databricks Project" to "folder" to make it easier to understand.

### DIFF
--- a/packages/databricks-vscode/src/configuration/auth/AuthProvider.ts
+++ b/packages/databricks-vscode/src/configuration/auth/AuthProvider.ts
@@ -5,7 +5,7 @@ import {
     WorkspaceClient,
     logging,
 } from "@databricks/databricks-sdk";
-import {ProgressLocation, window} from "vscode";
+import {CancellationToken, ProgressLocation, window} from "vscode";
 import {normalizeHost} from "../../utils/urlUtils";
 import {workspaceConfigs} from "../../vscode-objs/WorkspaceConfigs";
 
@@ -64,9 +64,13 @@ export abstract class AuthProvider {
      * This function should not throw an error and each implementing class must
      * handle it's own error messages and retry loops.
      */
-    protected abstract _check(): Promise<boolean>;
+    protected abstract _check(token?: CancellationToken): Promise<boolean>;
 
-    public async check(force = false, showProgress = true): Promise<boolean> {
+    public async check(
+        force = false,
+        showProgress = true,
+        cancellationToken?: CancellationToken
+    ): Promise<boolean> {
         if (force) {
             this.checked = false;
         }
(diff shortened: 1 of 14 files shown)


packages/databricks-vscode/src/configuration/auth/AuthProvider.ts
```diff
@@ -64,32 +64,62 @@ export abstract class AuthProvider {
...
+            await window.withProgress(
+                {
+                    location: ProgressLocation.Notification,
+                    title: `Databricks: Cancelling login using ${this.describe()}`,
+                },
+                async () => {
+                    await task;
+                }
+            );
+            window.showErrorMessage("Databricks: Login cancelled");
```
ilia-db, line 119: The progress will already show a notification, do we need another one here?

packages/databricks-vscode/src/configuration/auth/AzureCliCheck.ts
```diff
@@ -321,6 +372,10 @@ export class AzureCliCheck implements Disposable {
                 }; echo "Press any key to close the terminal and continue ..."; ${ShellUtils.readCmd()}; exit`
             );
 
+            cancellationToken?.onCancellationRequested(() =>
+                terminal.dispose()
```
ilia-db, line 376: I guess that will trigger `window.onDidCloseTerminal`?

packages/databricks-vscode/src/configuration/auth/DatabricksCliCheck.ts
```diff
@@ -82,22 +92,24 @@ export class DatabricksCliCheck implements Disposable {
         );
 
         try {
-            await workspaceClient.currentUser.me();
+            await workspaceClient.currentUser.me(
```
ilia-db, line 95: Just for my understanding: the workspace client will throw some "cancelled" error if the token is cancelled, right? (so we will hit `return false` case)

packages/databricks-vscode/src/configuration/auth/orchestrate.ts
```diff
@@ -31,17 +29,43 @@ export async function orchestrate<S, KEYS extends string>(
...
+
+        let result: StepResult<S, KEYS> | undefined = undefined;
+        const task: Promise<StepResult<S, KEYS> | undefined> = (async () => {
+            return await steps[step!]();
+        })().catch((e) => {
+            if (!isCancelled()) {
+                throw e;
+            }
+            return undefined;
+        });
```
ilia-db, line 55: A bit hard to read this part, might be simpler to just set the task here (`const task: Promise<StepResult<S, KEYS> | undefined> = steps[step]();`), and do the try/catch around `Promise.race`. What do you think?

packages/databricks-vscode/src/configuration/auth/AuthProvider.ts
kartikgupta-db: Progress just disappears. We need a final notification showing that something has cancelled.

packages/databricks-vscode/src/configuration/auth/DatabricksCliCheck.ts
kartikgupta-db: yep

packages/databricks-vscode/src/ui/bundle-resource-explorer/BundleCommands.ts
```diff
@@ -88,6 +89,20 @@ export class BundleCommands implements Disposable {
...
+            if (mode !== "development") {
+                const choice = await window.showInformationMessage(
+                    `Deploying bundle to ${humaniseMode(
+                        mode
+                    )} target "${target}".`,
+                    {modal: true},
+                    "Continue"
+                );
+                if (choice !== "Continue") {
+                    throw new Error("Deployment cancelled.");
```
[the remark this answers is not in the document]
kartikgupta-db: Yeah, that makes sense. Will update it.

ilia-db (approved):

packages/databricks-vscode/src/configuration/auth/orchestrate.ts
kartikgupta-db: I want to have a listener on the task in case it it is preempted. Otherwise, unhandled promises cause errors to show up in the extension host logs.

ilia-db (approved):

PR #1363, opened 2024-09-18
DB Connect Progress: Make sure we always end up at 100%

fjakobs:
## Changes
DB Connect Progress: Make sure we always end up at 100%

### DIFF
--- a/packages/databricks-vscode/resources/python/00-databricks-init.py
+++ b/packages/databricks-vscode/resources/python/00-databricks-init.py
@@ -379,11 +379,7 @@ def register_spark_progress(spark, show_progress: bool):
         SI_BYTE_SUFFIXES = ("EiB", "PiB", "TiB", "GiB", "MiB", "KiB", "B")
 
         def __init__(
-<<<<<<< HEAD
             self
-=======
-            self,
->>>>>>> b1ceb52 (Inline code from pyspark)
         ) -> None:
             self._ticks = None
             self._tick = None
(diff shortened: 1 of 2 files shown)


packages/databricks-vscode/resources/python/00-databricks-init.py
```diff
@@ -403,18 +405,24 @@ def init_ui(self):
         def update_ticks(
             self,
             stages,
-            inflight_tasks: int
+            inflight_tasks: int,
+            done: bool
         ) -> None:
-            total_tasks = sum(map(lambda x: x.num_tasks, stages))
-            completed_tasks = sum(map(lambda x: x.num_completed_tasks, stages))
+            total_tasks = builtinSum(map(lambda x: x.num_tasks, stages))
```
ilia-db, line 411: Just for general understanding - what was wrong with the old approach here?

packages/databricks-vscode/resources/python/00-databricks-init.py
```diff
@@ -403,18 +405,24 @@ def init_ui(self):
...
+
+                if done:
+                    self._tick = self._ticks
+                    self._running = 0
+
                 if self._tick is not None and self._tick >= 0:
                     self.output()            
                 self._running = inflight_tasks
-
+
```
ilia-db, line 425: There's a couple of dirty lines

ilia-db (approved):

packages/databricks-vscode/resources/python/00-databricks-init.py
fjakobs: I ran into a case where sum imported from pyskaprk.sql.functions was shadowing the builtin. I've added a comment.

ilia-db (approved):

PR #1364, opened 2024-09-18
Add venv or conda to gitignore before deployments

ilia-db:
## Changes
This only covers top level .venv or .conda environments (the only two you can create with the python extension flow that we trigger).

We ask the user permission to do that, which can be permanently ignored for a specific project

## Tests
manually


kartikgupta-db (approved):

ilia-db:
Cancelling in favour of using the CLI with the new `--verbose` flag: https://github.com/databricks/cli/pull/1774/files

We will still need to update the CLI, and add an ability to cancel deployments

PR #1373, opened 2024-09-24
Update js sdk to 0.6.2

ilia-db:
## Changes
0.6.2 reverted back to using fetch polyfill, which better handles http proxies. 

## Tests
Manually and existing tests


kartikgupta-db (commented):
Didn't we switch it in the first place, because undici supports proxies better?

fjakobs (approved):

PR #1394, opened 2024-10-15
Wait for the cluster to start in e2e tests

ilia-db:
Prevous logic was not awaiting properly if the cluster was in the PENDING state, which can happen when another suite just started it. This was causing run-* tests to fail.


ilia-db:
Closing for now, maybe in the future I'll be able to make it work, but for now it's a rabbit hole of different errors

PR #1397, opened 2024-10-18
Improve run button UX

ilia-db:
## Changes
- Add login command to the run context menu if the extension is logged out
- Focus config view when the login command is executed
- Fix a bug where the run button isn't visible if the vscode is opened with a python file already active

Screenshot 2024-10-18 at 11 13 24

 The Sign-In item is not visible in the logged in state

## Tests
Manually and existing tests


fjakobs (approved):
This UX is much better

PR #1405, opened 2024-10-24
[Internal] Automatically trigger integration tests on PR

hectorcast-db:
## Changes
Automatically trigger integration tests when a PR is opened or updated

## Tests
Workflows below.


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+                  app-id: ${{ secrets.DECO_WORKFLOW_TRIGGER_APP_ID }}
+                  private-key: ${{ secrets.DECO_WORKFLOW_TRIGGER_PRIVATE_KEY }}
+                  owner: ${{ secrets.ORG_NAME }}
+                  repositories: ${{secrets.REPO_NAME}}
+
+            - name: Trigger Workflow in Another Repo
+              env:
+                  GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+              run: |
+                  gh workflow run vscode-isolated-pr.yml -R ${{ secrets.ORG_NAME }}/${{secrets.REPO_NAME}} \
```
ilia-db, line 32: When we change the main branch back to main I'll need to update the vscode-isolated-pr.yml in the private repo, right?

ilia-db: Oh, and let's merge this into bundle-integ branch, not main

ilia-db:
It seems like we now run all integration tests in one workflow. This is not ideal for us. Currently we run each integ test in a separate workflow. This makes them orders of magnitude faster, but also properly isolates one test from the other. Can we use the same approach here? We can implement proper isolation within one workflow, but that's not a quick change

hectorcast-db:
Moved to here
https://github.com/databricks/databricks-vscode/pull/1406

PR #1413, opened 2024-10-29
Simplify UI for selecting virtual environments

ilia-db:
## Changes
Create our own environment selector quick pick:
- Explicit "Select Python Environment" title, not "Select Interpreter" that the python extension uses
- Show only environments, not all interpreters as the python extension does
- Still show the option to use python extension selector for power users

New selector:
Screenshot 2024-10-29 at 10 53 21

Old selector, or when you click on the "Use Python Extension to setup environments":
Screenshot 2024-10-29 at 10 53 30

As before, when there's no detected environments we show "create" dialog directly:
Screenshot 2024-10-29 at 11 06 04

## Tests
Manually

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -277,6 +277,13 @@
                 "category": "Databricks"
             },
             {
+                "command": "databricks.environment.setup",
+                "title": "Setup python environment",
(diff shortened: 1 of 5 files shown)


fjakobs (approved):
This is great. The old dialog was always a bit confusing.

Issue #1424, opened 2024-10-30
[BUG] Incorrect magics handling


ilia-db:
**Describe the bug**
Magics that start with the `# MAGIC` comment aren't handled unless they appear at the start of the cell. There's also an unexpected behaviour when executing the magic at the top of the cell - the cell itself is ignored. 

**To Reproduce**

**Screenshots**

**System information:**

1. Paste the output ot the `Help: About` command (CMD-Shift-P).
2. Databricks Extension Version

**Databricks Extension Logs**
Please attach the [databricks extension logs](https://docs.databricks.com/dev-tools/vscode-ext.html#send-usage-logs-to-databricks)

**Additional context**

ilia-db:
Almost all of the magics we handle are actual cell magics, so it makes sense to only process them if they are at the start of the cell. But we also handle `$run` in the same way, which looks incorrect

mroy-seedbox:
Isn't that how `%run` works?

From the [docs](https://docs.databricks.com/en/notebooks/notebook-workflows.html#use-run-to-import-a-notebook):
> %run must be in a cell by itself, because it runs the entire notebook inline.

ilia-db:
Good point. In `ipython` it's possible to have multiple %run commands, but I forgot about databricks differences. In the extension we run things with the local ipython, but it makes more sense to be in line with databricks logic, otherwise it won't be possible to run the same file remotely. 

Going to close the issue and the PR, thanks a lot for the comment

PR #1430, opened 2024-10-30
Handle %run magic for each line

ilia-db:
## Changes
Currently we treat all magic commands as cell magics:
- We only detect magics at the beginning of the cell
- You can't have multiple commands in the same cell

It makes sense for everything except %run.

With the new changes it's now possible to have %run (or #MAGIC %run) command in the middle of the cell, or even multiple %run commands in the same cell.

https://github.com/databricks/databricks-vscode/issues/1424

## Tests
Manually

### DIFF
--- a/packages/databricks-vscode/resources/python/00-databricks-init.py
+++ b/packages/databricks-vscode/resources/python/00-databricks-init.py
@@ -166,12 +166,10 @@ def is_databricks_notebook(py_file: str):
         with open(py_file, "r") as f:
             return "Databricks notebook source" in f.readline()
 
-def strip_hash_magic(lines: List[str]) -> List[str]:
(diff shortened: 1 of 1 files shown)


mroy-seedbox:
Did something change in Databricks?

Because [the documentation](https://docs.databricks.com/en/notebooks/notebook-workflows.html#use-run-to-import-a-notebook) says this:
> %run must be in a cell by itself, because it runs the entire notebook inline.

AFAIK, it has never been possible to have a `%run` command in the middle of the cell, or multiple `%run` commands in the same cell.

ilia-db:
Context on why it's closed is in this issue: https://github.com/databricks/databricks-vscode/issues/1424

PR #1436, opened 2024-11-01
Add validate runs for pipelines

ilia-db:
## Changes
Validate action is available as a separate button on the pipeline item in the resource explorer:

Screenshot 2024-11-01 at 10 02 27

Haven't found a super suitable icon for this action (in the workspace we don't use any icons for the validate action). Currently using the "beaker" icon that's generally used in tests, and reached out to the designers for help.

## Tests
Yes

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/deploy_and_run_pipeline.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/deploy_and_run_pipeline.e2e.ts
@@ -86,7 +86,8 @@ describe("Deploy and run pipeline", async function () {
             resourceExplorerView,
             "Pipelines",
             pipelineName,
-            "Completed",
+            // TODO: the account we are using to run tests doesn't have permissions to create clusters for the pipelines
+            "Failed",
             // Long timeout, as the pipeline will be waiting for its cluster to start
             15 * 60 * 1000
         );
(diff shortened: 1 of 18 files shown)


packages/databricks-vscode/src/test/e2e/deploy_and_run_pipeline.e2e.ts
```diff
@@ -0,0 +1,95 @@
...
+        await deployAndRunButton.elem.click();
+
+        await waitForDeployment();
+
+        await waitForRunStatus(
+            resourceExplorerView,
+            "Pipelines",
+            pipelineName,
+            // TODO: the account we are using to run tests doesn't have permissions to create clusters for the pipelines
+            "Failed",
```
ilia-db, line 90: Investigating this right now, but I think it's ok to merge the tests as is for now

packages/databricks-vscode/src/test/e2e/utils/commonUtils.ts
```diff
@@ -200,21 +200,27 @@ export async function waitForInput() {
 }
 
 export async function waitForLogin(profileName: string) {
+    console.log("Waiting for login");
```
fjakobs, line 203: do you want to keep the `console.log` here?

fjakobs (approved):

packages/databricks-vscode/src/test/e2e/utils/commonUtils.ts
ilia-db: It's e2e test utility - the more logs there, the easier it is to debug them

PR #1474, opened 2024-11-28
Improve DLT UX

ilia-db:
## Changes
- Load events with `retry` instead of `setTimeout`
- Better handle cancellation
- Show spinning icon for more DLT states
- Merge Event Log and Run Status tree sections
- Remove run "cause" item, since the same thing is visible in one of the events

Screenshot 2024-11-28 at 16 24 04


fjakobs (approved):
Looking good

PR #1479, opened 2024-12-05
Add `Datasets` section with dataset schemas to the pipelines explorer

ilia-db:
## Changes
Initially Datasets aren't loaded, but you can expand them to trigger loading
Screenshot 2024-12-05 at 10 14 41

If we can't find dataset definitions in the latests runs we show "please run or validate" item:
Screenshot 2024-12-05 at 10 14 58

Screenshot 2024-12-05 at 10 15 14

Running or validating the pipeline will also show 'loading' state for the datasets section:

Screenshot 2024-12-05 at 10 28 09

## Tests
Unit and e2e tests


fjakobs (approved):
This is really nice!

PR #1519, opened 2025-01-13
Bundle selection ux

ilia-db:
## Changes

When you land on a workspace without dabs config in the root, you can now select a sub-folder as the active bundle folder (without changing the workspace root, and without reloading the whole IDE).

Screenshot 2025-01-16 at 10 40 16

The above is for the case when we can't find any sub-projects ourselves. Clicking on the top button opens an OS folder selection dialog.

When we can detect sub projects, we show the "select sub-folder" button on top:

Screenshot 2025-01-16 at 10 39 44

When you select the sub-folder and initialise the extension, we now show additional "Local Folder" UI at the top of the configuration:

Screenshot 2025-01-14 at 10 11 09

Screenshot 2025-01-14 at 10 17 39

The UI for selecting sub-folders:

Screenshot 2025-01-14 at 10 11 52

"Select another folder" opens a OS-level selection dialog. 

If you select a folder that's part of the workspace, we don't reload the IDE and just point our extension to it. When the selected folder is outside of the workspace, we reload the IDE with the folder being the new workspace root.

After you select some sub-folder it's saved to workspace-scoped vscode storage, so the next time this workspace is opened our extension already knows what sub-folder to use.

After the extension is initialised, you can now easily change the active bundle folder if you have multiple bundles in different sub-folders.

## Tests

Manually and with a few new unit tests, e2e tests will be in the follow up PR

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -204,6 +204,13 @@
                 "category": "Databricks"
             },
             {
+                "command": "databricks.bundle.selectActiveProjectFolder",
+                "icon": "$(gear)",
+                "title": "Select Active Databricks Project Folder",
+                "enablement": "databricks.context.activated && databricks.context.bundle.deploymentState == idle",
+                "category": "Databricks"
+            },
+            {
                 "command": "databricks.bundle.refreshRemoteState",
                 "icon": "$(refresh)",
                 "title": "Refresh remote state",
@@ -444,8 +451,13 @@
             },
             {
                 "view": "configurationView",
-                "contents": "There are multiple Databricks projects in the folder:\n[Open existing Databricks Project](command:databricks.bundle.openSubProject)",
+                "contents": "There are multiple Databricks projects in the folder:\n[Open existing Databricks Project](command:databricks.bundle.selectActiveProjectFolder)",
                 "when": "workspaceFolderCount > 0 && databricks.context.initialized && databricks.context.subProjectsAvailable"
+            },
+            {
+                "view": "configurationView",
+                "contents": "[Select a folder with a Databricks Project](command:databricks.bundle.selectActiveProjectFolder)",
+                "when": "workspaceFolderCount > 0 && databricks.context.initialized && !databricks.context.isBundleProject && !databricks.context.subProjectsAvailable"
             },
             {
                 "view": "configurationView",
@@ -558,6 +570,16 @@
                     "command": "databricks.utils.openExternal",
                     "when": "viewItem =~ /^databricks.*\\.(has-url).*$/ && databricks.context.bundle.deploymentState == idle",
                     "group": "navigation_2@0"
+                },
+                {
+                    "command": "databricks.bundle.selectActiveProjectFolder",
+                    "when": "viewItem =~ /^databricks.configuration.activeProjectFolder/ && databricks.context.bundle.deploymentState == idle",
+                    "group": "navigation_2@0"
+                },
+                {
+                    "command": "databricks.bundle.selectActiveProjectFolder",
+                    "when": "viewItem =~ /^databricks.configuration.activeProjectFolder/ && databricks.context.bundle.deploymentState == idle",
+                    "group": "inline@2"
                 },
                 {
                     "command": "databricks.connection.attachCluster",
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -451,22 +451,22 @@
             },
             {
                 "view": "configurationView",
-                "contents": "Detected multiple Databricks projects in the sub-folders of the workspace:\n[Select a sub-folder with a Databricks Project](command:databricks.bundle.selectActiveProjectFolder)",
+                "contents": "Detected multiple Databricks projects in the sub-folders of the workspace:\n[Select a sub-folder with a Databricks project](command:databricks.bundle.selectActiveProjectFolder)",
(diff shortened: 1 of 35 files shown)


packages/databricks-vscode/package.json
```diff
@@ -444,17 +451,22 @@
...
             },
             {
                 "view": "configurationView",
-                "contents": "[Create a new Databricks Project](command:databricks.bundle.initNewProject)",
+                "contents": "[Select a sub-folder with a Databricks project](command:databricks.bundle.selectActiveProjectFolder)",
+                "when": "workspaceFolderCount > 0 && databricks.context.initialized && !databricks.context.isBundleProject && !databricks.context.subProjectsAvailable"
+            },
+            {
+                "view": "configurationView",
+                "contents": "[Create a new Databricks project](command:databricks.bundle.initNewProject)",
```
juliacrawf-db, line 469: The UI text here needs to be shuffled and edited a bit so the button text can be much shorter. (I can pose options offline.) Also, we should be consistent in casing when we refer to a Databricks project (lower case).

packages/databricks-vscode/resources/python/dbconnect-bootstrap.py
```diff
@@ -3,23 +3,19 @@
...
-    if not os.path.exists(os.path.dirname(env_file_path)):
-        return
-    
-    with open(env_file_path, "r") as f:
-        for line in f.readlines():
-            key, value = line.strip().split("=", 1)
-            os.environ[key] = value
-    return
-
+def load_env_from_leaf(path: str) -> bool:
```
ilia-db, line 6: This part is the same logic we do in the notebooks init script. It's required here now since the cwd is the root of the workspace, not the root of a bundle folder (which can be nested now)

fjakobs (approved):

packages/databricks-vscode/package.json
```diff
@@ -203,6 +203,13 @@
                 "enablement": "databricks.context.activated",
                 "category": "Databricks"
             },
+            {
+                "command": "databricks.bundle.selectActiveProjectFolder",
+                "icon": "$(gear)",
+                "title": "Select Databricks Project Folder",
```
juliacrawf-db, line 209:
```suggestion
                "title": "Select a Databricks project folder",
```
I think we want to stick to sentence casing for directives.

packages/databricks-vscode/package.json
```diff
@@ -444,17 +451,22 @@
             },
             {
                 "view": "configurationView",
-                "contents": "There are multiple Databricks projects in the folder:\n[Open existing Databricks Project](command:databricks.bundle.openSubProject)",
+                "contents": "Detected multiple Databricks projects in the VSCode workspace:\n[Select a project](command:databricks.bundle.selectActiveProjectFolder)",
```
juliacrawf-db, line 454:
```suggestion
                "contents": "Multiple Databricks projects were detected in the workspace:\n[Choose a project](command:databricks.bundle.selectActiveProjectFolder)",
```
It's a fuzzy (and honestly not super important) line, but it seems like this is a choice, not just a selection.

And poking around, I'm not sure VSCode is approved product terminology, so instead of expanding it, I'd avoid it by removing it.

packages/databricks-vscode/package.json
```diff
@@ -444,17 +451,22 @@
...
                 "when": "workspaceFolderCount > 0 && databricks.context.initialized && databricks.context.subProjectsAvailable"
             },
             {
                 "view": "configurationView",
-                "contents": "Migrate current folder to a Databricks Project: \n[Migrate current folder to a Databricks Project](command:databricks.bundle.startManualMigration)",
+                "contents": "No Databricks project configuration detected in the root of the VSCode workspace:\n[Create configuration](command:databricks.bundle.startManualMigration)",
```
juliacrawf-db, line 459:
```suggestion
                "contents": "Databricks project configuration was not detected in the root of the workspace:\n[Create configuration](command:databricks.bundle.startManualMigration)",
```

packages/databricks-vscode/package.json
```diff
@@ -444,17 +451,22 @@
...
                 "when": "workspaceFolderCount > 0 && databricks.context.initialized && databricks.context.pendingManualMigration"
             },
             {
                 "view": "configurationView",
-                "contents": "[Create a new Databricks Project](command:databricks.bundle.initNewProject)",
+                "contents": "No Databricks projects detected in the VSCode workspace:\n[Select a project manually](command:databricks.bundle.selectActiveProjectFolder)",
```
juliacrawf-db, line 464:
```suggestion
                "contents": "A Databricks project was not detected in the workspace:\n[Select a project](command:databricks.bundle.selectActiveProjectFolder)",
```
Maybe this is "Choose" instead too?

packages/databricks-vscode/package.json
```diff
@@ -444,17 +451,22 @@
...
+                "when": "workspaceFolderCount > 0 && databricks.context.initialized && !databricks.context.isBundleProject && !databricks.context.subProjectsAvailable"
+            },
+            {
+                "view": "configurationView",
+                "contents": "Chose a new parent folder and create a Databricks project based on a [template](https://docs.databricks.com/en/dev-tools/bundles/templates.html#databricks-asset-bundle-project-templates):\n[Create a new project](command:databricks.bundle.initNewProject)",
```
juliacrawf-db, line 469:
```suggestion
                "contents": "Select a folder where you want to create a new Databricks project:\n[Create a new project](command:databricks.bundle.initNewProject)",
```
Typo fix, and I'm not sure the template link is necessary? If so make it "...project using a [template]..."

PR #1525, opened 2025-01-17
Show all dabs resources in the explorer panel

ilia-db:
## Changes

"Unknown" resources are shown with the "created/deleted" icons and external links (if they have been deployed already).

Screenshot 2025-01-17 at 11 51 54
Screenshot 2025-01-17 at 11 45 11

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -1125,7 +1125,7 @@
         "useYarn": false
     },
     "cli": {
-        "version": "0.236.0"
+        "version": "0.238.0"
     },
     "scripts": {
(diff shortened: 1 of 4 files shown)


fjakobs (approved):
thanks for the change.

Not a big fan of the generic icons but couldn't any better ones myself.

fjakobs:
Maybe you can spend a few more minutes to create specific resource types for experiment, dashboards, and models so we have better icons.

ilia-db:
> Maybe you can spend a few more minutes to create specific resource types for experiment, dashboards, and models so we have better icons.

Doing that in a follow up

PR #1539, opened 2025-01-31
Add serverless support to the extension

ilia-db:
## Changes
- Cluster quick-pick now has a new "Serverless" option
- Connection manager now has a new `serverless` mode
- EnvironmentDependenciesVerifier doesn't force a cluster if serverless mode is enabled
   - But if now requires a higher Databricks Connect version
   - EnvironmentDependenciesInstaller is also updated to install a more appropriate dbconnect version
- We now inject DATABRICKS_SERVERLESS_COMPUTE_ID env var when serverless is enabled
- Run file as Workflow command automatically
   - The run logic was moved from Cluster to WorkflowRun
   - Updated workflow output panel UI to show serverless when it's enabled
- Upload and run file command is disabled in the serverless mode
- Fix for the python wrapper to avoid exceptions when .venv folder is removed at runtime

Follow ups:
- Automatically enable serverless mode if the selected .databrickscfg profile has `serverless_compute_id = auto`
- Figure out if we can check if a workspace has serverless enabled before offering to use it
- Integ tests in a serverless-enabled environment (and we should finally cover dbconnect logic)

## Tests

Manual, a few new unit tests, integ tests will be in a follow-up

### DIFF
--- a/packages/databricks-vscode/src/language/EnvironmentDependenciesInstaller.ts
+++ b/packages/databricks-vscode/src/language/EnvironmentDependenciesInstaller.ts
@@ -2,7 +2,8 @@ import {commands, EventEmitter, OutputChannel, window} from "vscode";
 
 import {Disposable} from "vscode";
 import {MsPythonExtensionWrapper} from "./MsPythonExtensionWrapper";
-import {DATABRICKS_CONNECT_VERSION} from "../utils/constants";
+import {ConnectionManager} from "../configuration/ConnectionManager";
+import {DATABRICKS_CONNECT_VERSION as DATABRICKS_CONNECT_MINIMAL_VERSION} from "../utils/constants";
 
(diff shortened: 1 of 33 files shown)


packages/databricks-vscode/src/language/EnvironmentDependenciesInstaller.ts
```diff
@@ -53,10 +59,25 @@ export class EnvironmentDependenciesInstaller implements Disposable {
         this.onDidInstallAttemptEmitter.fire();
     }
 
+    async getSuggestedVersion() {
+        if (this.connectionManager.serverless) {
```
fjakobs, line 63: at some point we need to take the client version into consideration but this is good for now.

fjakobs (approved):
Lokking good

PR #1550, opened 2025-02-10
Automatically enable serverless when serverless_compute_id=auto

ilia-db:
## Changes

Automatically enable serverless when serverless_compute_id=auto

## Tests

Manually, e2e tests in the followup


packages/databricks-vscode/src/configuration/DatabricksWorkspace.ts
```diff
@@ -90,27 +106,36 @@ export class DatabricksWorkspace {
...
         }
+        try {
+            const serverlessEnablement = (await client.apiClient.request(
+                `/api/2.0/settings-api/workspace/${id}/serverless_job_nb`,
+                "GET"
+            )) as ServerlessEnablementResponse;
+            const enableServerless =
+                serverlessEnablement?.setting?.value?.preview_enablement_val
+                    ?.enabled === true;
+            state = {...state, enableServerless};
```
ilia-db, line 134: cc @denik - the place were we check if the workspace has serverless support or not

packages/databricks-vscode/src/configuration/DatabricksWorkspace.ts
```diff
@@ -90,27 +106,36 @@ export class DatabricksWorkspace {
...
+        } catch (e) {
+            ctx?.logger?.error("Can't detect serverless support", e);
```
denik, line 136: Does it fallback to serverless or non-serverless?

ilia-db: We simply don't offer users to select `serverless` mode in the cluster dropdown UI. So this happens if the serverless is not supported, or if the request fails.

fjakobs (approved):

PR #1571, opened 2025-02-20
Add devcontainer

### LINKED ISSUE #1572: Add devcontainer
I already create the pr, but decided to create an issue for further discussions anyway.
I'd love to see a devcontainer to simplify the setup of the dev env in a standardized way

twsl:
## Changes

In order to supply a standardized development environment for further development, I decided to add a devcontainer.
This also updates yarn to the latest stable version, as the mentioned yarn version was released almost exactly 3 years ago.

Fix #1572

## Tests

Tested manually


twsl:
I would also exclude the .yarn folder from being committed to git, but I am unsure, if there is a legit reason to do so

ilia-db:
Hi, thanks for that! But can you provide a bit more rationale of why this is useful in our repo? E.g is this only to improve the development of the extension itself? (also, can we not update yarn version in this same PR, I'd rather do it separately)

twsl:
> Hi, thanks for that! But can you provide a bit more rationale of why this is useful in our repo? E.g is this only to improve the development of the extension itself? (also, can we not update yarn version in this same PR, I'd rather do it separately)

yes, this only supports and improves the development of the extension itself by providing every developer who wants to contribute with the same docker-based, reproduce-able environment without the hassle of following any required setup instructions like an using and old yarn version.
I'll revert the yarn changes and can provide a separate pr for that after this one is merged

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,51 @@
...
+                },
+                "[javascript]": {
+                    "editor.defaultFormatter": "esbenp.prettier-vscode"
+                }
+            },
+            "extensions": [
+                "dbaeumer.vscode-eslint",
+                "connor4312.esbuild-problem-matchers",
+                "ms-vscode.extension-test-runner",
+                "dbaeumer.vscode-eslint",
```
ilia-db, line 42: Duplicate extension ID

.yarnrc.yml
```diff
@@ -1,12 +1,12 @@
 nodeLinker: node-modules
 
 plugins:
-  - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
-    spec: "@yarnpkg/plugin-workspace-tools"
-  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
-    spec: "@yarnpkg/plugin-interactive-tools"
-  - path: .yarn/plugins/@yarnpkg/plugin-version.cjs
-    spec: "@yarnpkg/plugin-version"
+    - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
```
ilia-db, line 4: Let's revert changes here too? Also in the `package.json` and `yarn.lock` files

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,51 @@
...
+                "amodio.tsl-problem-matcher",
+                "esbenp.prettier-vscode",
+                "Tobermory.es6-string-html"
+            ]
+        }
+    },
+    // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+    "remoteUser": "root"
```
ilia-db, line 50: Is "root" actually required?

ilia-db (commented):
Sorry for slow review, left a few of comments.

I'm still figuring out the contribution process - Databricks license is not a standard open source one, and we require all contributors to sign a Contributor License Agreement. I'll let you know how to do that soon.

.yarnrc.yml
twsl: The file is formatted according to prettier, so I'd argue, that it should stay this way.
The changes in package.json and yarn.lock are produced by the defined yarn version and the specified `databricks-sdk` version, so I'd also argue, that these are correct

twsl: If you want to remove the hash from the `packageManager` field,please provide, how you would like to install `yarn`, as this came from `corepack`

ilia-db: We have the setup guide in the readme here: https://github.com/databricks/databricks-vscode?tab=readme-ov-file#getting-started

Installing the yarn (and sdk) that way doesn't result in different sha, I'd prefer to keep it that way, at least in the scope of this PR. 

Sorry for being nit picky, I'm just not comfortable with accepting changes that modify dependency versions (and their hashes) without obvious (to me) reasons.

twsl: no worries, I'll fix this and setup a different pr for a possible switch to corepack and yarn v4

twsl: The referenced instructions define v2 while the `package.json` defined v3.2.1. which one is correct? I assume the later one?

twsl:
CLA has been signed, now I just need to know how to install which version of yarn

.yarnrc.yml
ilia-db: Reading the corepack docs, why do we need to run `corepack use`? It will [indeed modify package.json, based on their docs](https://github.com/nodejs/corepack?tab=readme-ov-file#corepack-use-nameversion). 

Should `corepack enable && yarn install` be enough?

yarn.lock
```diff
@@ -309,7 +309,7 @@ __metadata:
     ini: ^4.1.2
     reflect-metadata: ^0.2.1
     semver: ^7.6.0
-  checksum: be59ab37615aa63780fdb841cdd6199f930710d409e29233abc2c66d8e5a7ac9928497bdf2f38075970e89e6ade0d9ae37ea39efcf721bd85f0c484c535f46d0
+  checksum: fc3186dc4dda2a06e01c8e7e20b5f76ec6d1b4379b65deb80cacfd506427f38841af890858bd3335c3ab76f8a7255c4f4ce295f6470d7d0905f236b211fa617b
```
ilia-db, line 312: Why are these checksums different though? The `databricks-sdk.tgz` hasn't changed after all

twsl: Fixed it. The sdk changed throughout this pr and at some point i ran the `install:sdk` command.

twsl:
anything else blocking this, @ilia-db ?

ilia-db (approved):

ilia-db:
That was very long review process, sorry about it. Merged the PR

PR #1574, opened 2025-02-25
Hide auto-generated cell in ipynb workflows

ilia-db:
## Changes

Using the same metadata that web UI sets when you click on "hide code" option for a cell.

It looks like this, and the cell can be expanded by clicking on the eye button (after which you can't hide it anymore)
Screenshot 2025-02-25 at 15 31 09

Dbnb notebooks will still have the generated cells visible.

## Tests

Manually + existing integ tests


fjakobs (approved):
Nice, this is cleaner than I expected it to be

PR #1579, opened 2025-02-26
Add Documentation view

ilia-db:
## Changes

The new view is positioned below the Configuration view. It contains a Quick Start Guide and links to various documentation pages.

By default it's visible and expanded, but since it's a separate "view", users can close it, move it to a different place, or hide it completely.

Screenshot 2025-02-27 at 10 03 26

And that how it looks with experimental views enabled:

Screenshot 2025-02-27 at 10 02 40

## Tests

Manually


packages/databricks-vscode/package.json
```diff
@@ -419,6 +419,10 @@
                     "id": "configurationView",
                     "name": "Configuration"
                 },
+                {
+                    "id": "databricksDocsView",
+                    "name": "Documentation"
+                },
```
juliacrawf-db, line 425: I would not put this here, as it disconnects functionality sections. It should be the last section. (And on the topic of reorganizing, maybe the subsequent sections after Configuration should flow in the same order as items in the Configuration section? So Clusters would be next, then bundle sections?)

juliacrawf-db:
I think this is neat, but is this a common/familiar way to present docs for an extension?

packages/databricks-vscode/package.json
ilia-db: Good point, I'll change that

ilia-db:
> I think this is neat, but is this a common/familiar way to present docs for an extension?

I haven't seen other extensions who do that, but it looks and behaves pretty naturally

PR #1584, opened 2025-02-27
Properly terminate debug session if the workflow's webview panel is closed

ilia-db:
## Changes

Before we were getting uncaught exceptions if users close webview panel, and the debug session was stuck forever (unless you click on the stop button manually)

## Tests

Manually


fjakobs (commented):
I don't see where this would terminate the debug session. All I see is additional logging.

PR #1591, opened 2025-03-04
Rename workflows to jobs in the bundle explorer UI

ilia-db:
## Changes

This change renames workflows to jobs in the bundle explorer UI to align with the terminology used in the Databricks UI.

Note that we still have "run file as workflow" command and `databricks-workflow` launch config - decided to leave it as is to not break existing users' workflows (and it's still a correct term in that context)

## Tests

Manually and existing integ tests


fjakobs (approved):
lgtm

PR #1593, opened 2025-03-05
Simplify profile selection UI

ilia-db:
## Changes

And recommend OAuth profiles for new users (who don't have pre-existing profiles defined).

There are three main improvements:
- Remove confusing `(user to machine)` suffix for OAuth profile type
- Add `(recommended)` suffix to OAuth profile type (but only if there are no pre-existing profiles)
- Remove separators with headings (make them headless). We already give the context in the item descriptions, additional headings only create clutter (and can even be confused, since they show up for the first item in the group, instead of on top of the group)

New UI when there are no pre-existing profiles:

Screenshot 2025-03-05 at 13 45 02

New UI when there are pre-existing profiles (we don't show `recommended` for OAuth then):

Screenshot 2025-03-05 at 13 44 03

Old UI:

Screenshot 2025-03-05 at 13 48 20

Screenshot 2025-03-05 at 13 47 47

### DIFF
--- a/packages/databricks-vscode/src/configuration/LoginWizard.ts
+++ b/packages/databricks-vscode/src/configuration/LoginWizard.ts
@@ -134,8 +134,6 @@ export class LoginWizard {
     }
 
     private async getProfileQuickPickItems() {
-        const items: Array<AuthTypeQuickPickItem> = [];
-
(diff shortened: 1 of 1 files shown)


fjakobs (approved):
I like the tweaks

PR #1595, opened 2025-03-06
Avoid python extension interference in the bundle init terminal

ilia-db:
## Changes

Fixing bugs with `setTimeout`, yeehaw!

See the comment in the code.

How the bug looks like:

Screenshot 2025-03-06 at 10 20 02

After the fix:
Screenshot 2025-03-06 at 10 20 10

The bug only happens when python extension has specific A/B test enabled, not for everyone. 

## Tests

We have existing e2e tests for the wizard


fjakobs (approved):
Ugly but seems neccessry

PR #1608, opened 2025-03-18
Show error notifications after failed environment setup

### LINKED ISSUE #1601: Notify about Python Environment setup results
Right now there are two places you can trigger them: in the left panel, or with the "Databricks Connect disabled" button in the status line. By clicking on the latter it might be unclear why the setup doesn't work, as we don't communicate anywhere aside from the left panel "Python Environment" section (which is not well discoverable). 

We should show a success/failure notifications with a clear description after each time you try to setup Python environment / Databricks Connect.

ilia-db:
## Changes

Also change the setup logic slightly: we were re-checking the steps before taking an action, now we are doing it after the action. See comments in the code.

Fixes #1601

## Tests

Existing tests and manually


fjakobs (approved):
For these kind of changes adding a small video to the PR description would be helpful.

fjakobs (approved):

PR #1612, opened 2025-03-19
Improve handling of the %run magic for local runs

### LINKED ISSUE #1611: [BUG] %run commands fail to run when run when using VsCode extension.
**Describe the bug**
Run commands fail to run when run when using VsCode extension.

- Removing the `%` from `%run` and any cell title seems to work.
- Removing only `# MAGIC` prefix makes the cell run correctly
- Removing the `%` results in the cell running successfully in the python interactive window but then trying to call a function that lives in the notebook fails.

example cell:

`# MAGIC %run "../../my_notebook"`

Running this results in error:

```
with databricks_notebook_exec_env('notebook_file_path', '"../../my_notebook"') as file:
                                                                                                                                       ^
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
```

**To Reproduce**

1. create a cell which uses the %run command to run another notebook such as:
```
# DBTITLE 1,run_notebook
# MAGIC %run "../../my_notebook"
```
2. See error

**Screenshots**

ilia-db:
## Changes
- Strip quotes around filenames
- Use raw strings for the project root and filename (avoid unicode issues for paths with `\u` sequences - e.g. `\User\***` on Windows)

Fixes #1611

## Tests

Existing integ tests

### DIFF
--- a/packages/databricks-vscode/resources/python/00-databricks-init.py
+++ b/packages/databricks-vscode/resources/python/00-databricks-init.py
@@ -316,7 +316,8 @@ def register_magics(cfg: LocalDatabricksNotebookConfig):
                 if len(rest) == 0:
                     return lines
                 
-                filename = rest[0]
+                # Strip whitespace or possible quotes around the filename
+                filename = rest[0].strip('\'" ')
 
                 for suffix in ["", ".py", ".ipynb", ".ipy"]:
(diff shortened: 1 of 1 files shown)


fjakobs (approved):
I think we should start adding unit tests for the jupyter init code

PR #1614, opened 2025-03-20
Always show warning before destroying a bundle

ilia-db:
## Changes

WIP: need to update integ tests


pietern (approved):
LGTM, approving to unblock.

PR #1620, opened 2025-03-24
Fix ProgressHandler conflicts with datetime imports

### LINKED ISSUE #1618: [BUG] Cannot run any code or cell after importing * from datetime - Databricks Connect 15.4.7
**Describe the bug**
After upgrading cluster runtime to 15.4 LTS and then upgrading the Databricks Connect package to 15.4.7 using the VSCode extension, cells can no longer be run at all it seems. Every cell gives the error:

`AttributeError: 'ProgressHandler' object has no attribute 'p'`

Update:
The error message given is at the end of the stack trace.

The issue seems to occur when  you import datetime using `from datetime import *`. If this is removed then the notebook works!

**To Reproduce**

1. Use Databricks Copnnect 15.4.7
2. Run any cell that attempts to run on the cluster after running `from datetime import *`
```
from datetime import *
df = spark.read.table("samples.nyctaxi.trips")
df.show(5)
```

Image

The spark table read fails after importing * from datetime
Image

Image

ilia-db:
## Changes

Fixes #1618

## Tests

Manually

### DIFF
--- a/packages/databricks-vscode/resources/python/00-databricks-init.py
+++ b/packages/databricks-vscode/resources/python/00-databricks-init.py
@@ -4,7 +4,8 @@ import json
 from typing import Any, Union, List
 import os
 import sys
-import time
+# Avoid conflicts with possible "datetime" imports (in the user code)
+import time as _time
 import shlex
 import warnings
 import tempfile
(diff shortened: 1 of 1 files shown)


ilia-db:
> do we have similar problems with other imports?

Technically all our top level imports can be overriden by the user code, I'll follow up on this PR and move the imports inside the functions, or use `as` with a more unique name on the top level.

PR #1648, opened 2025-04-08
Add waiters for get methods as a extension

parthban-db:
## Changes

Waiters have been removed from the JS SDK as they are not LRO. This PR adds removed waiters for get methods as an extension of JS SDK.
## Tests

Existing CI.


packages/databricks-vscode/src/test/e2e/resources/dummy-test/package-lock.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "dummy-test",
```
ilia-db, line 2: Let's remove this file too

Also, what command creates them for you? Maybe it makes sense to have them, but I'm not sure

packages/databricks-vscode/src/bundle/run/JobRunStatus.ts
```diff
@@ -47,7 +48,9 @@ export class JobRunStatus extends BundleRunStatus {
         try {
             this.runState = "running";
             await (
-                await client.jobs.getRun({run_id: parseInt(this.runId)})
+                await WorkflowRun.getRun(client.apiClient, {
+                    run_id: parseInt(this.runId),
+                })
```
ilia-db, line 53: It seems like `WorkflowRun` already has all the logic to make it work. Can we do something like this:

```javascript
const run = await WorkflowRun.fromId(client.apiClient, this.runId)
await run.wait((state, run) => {
    this.data = run // Different type, so you'll need to change its usages (in the JobRunStatusTreeNode at least)
    ...
})
```

The existing `wait` doesn't have `timeout` option, but given how big of a timeout we use here, I don't think it's required.

packages/databricks-vscode/src/test/e2e/resources/dummy-test/package-lock.json
parthban-db: Not able to reproduce it.

packages/databricks-vscode/src/sdk-extensions/WorkflowRun.ts
```diff
@@ -116,6 +116,10 @@ export class WorkflowRun {
         return this.details.run_page_url || "";
     }
 
+    get run(): jobs.Run {
```
ilia-db, line 119: Let's call it `details`? `run.run` is hard to read. We can just make `details` public, but getter is also fine

Issue #1325, opened 2024-08-22
[BUG] stdout logs not appearing during databricks launch runs


rwpurvis:
**Describe the bug**

When running tests via the vscode debug functionality [per the tutorial](https://docs.databricks.com/en/dev-tools/vscode-ext/dev-tasks/pytest.html), the logs do not appear in the debug console. I am able to see the logs on the Driver logs page in the UI.

**To Reproduce**

1. Add a pytest runner script and `launch.json` configuration
2. Run `Unit Tests (on Databricks)`
3. Observe output with missing logs in Debug Console

Example output (anonymized):
```
8/22/2024, 3:24:37 PM - Creating execution context on cluster <cluster ID> ...
8/22/2024, 3:24:42 PM - Synchronizing code to <directory> ...
8/22/2024, 3:24:42 PM - Running <directory>\pytest_databricks.py ...


8/22/2024, 3:26:56 PM - Done (took 138949ms)
```

Previously, pytest logs appeared here.

**Screenshots**

See above output

**System information:**

1. Paste the output ot the `Help: About` command (CMD-Shift-P).
```
Version: 1.85.2 (user setup)
Commit: 8b3775030ed1a69b13e4f4c628c612102e30a681
Date: 2024-01-18T06:40:10.514Z
Electron: 25.9.7
ElectronBuildId: 26354273
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Windows_NT x64 10.0.19045
```
5. Databricks Extension Version
Tried with v2.3.2 and v1.3.1

**Databricks Extension Logs**
Please attach the [databricks extension logs](https://docs.databricks.com/dev-tools/vscode-ext.html#send-usage-logs-to-databricks)

**Additional context**

rwpurvis:
Some more info here...I mark all pyspark tests with `slow`, so that non-pyspark tests can be run locally via `pytest -m "not slow"`. When I select these tests via the `launch.json`, the stdout logs appear. Here's the `launch.json`:
```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "databricks",
            "request": "launch",
            "name": "Unit Tests (on Databricks)",
            "program": "${workspaceFolder}/<sub_directory>/pytest_databricks.py",
[...]

PR #1814, opened 2025-12-03
feat(runtime): Implement streaming execution output (fixes #1813)

### LINKED ISSUE #1813: [BUG] Real-time output missing when using “Upload and Run File” from VS Code
**Describe the bug**
I am running Python files on a Databricks cluster using the VS Code Databricks extension, specifically the “Upload and Run File” command.
I cannot get real-time output in the Debug Console. I have checked the official docs:
https://learn.microsoft.com/en-us/azure/databricks/dev-tools/vscode-ext/tutorial
[https://github.com/databricks/databricks-vscode/blob/release-v2.10.3/packages/databricks-vscode/DATA...](https://github.com/databricks/databricks-vscode/blob/release-v2.10.3/packages/databricks-vscode/DATABRICKS.quickstart.md#running-pxspark-code)
but these do not address the issue.

**To Reproduce**
The behavior I see is:
While the script is running, no output is shown in the Debug Console.
When the script finishes, all output appears at once.
If the script fails with an exception, only the error is shown and none of the printed output appears.
This makes it very difficult to test and debug. If this is expected behavior, I would like to know what the recommended or best-practice workflow is for running and testing a standalone Python file on a Databricks cluster with live output.
Below is a minimal example based on Databricks’ own sample script. I added a loop with prints and sleeps to demonstrate the missing streaming output:

```
from pyspark.sql import SparkSession
from pyspark.sql.types import *

spark = SparkSession.builder.getOrCreate()

schema = StructType([
    StructField('CustomerID', IntegerType(), False),
    StructField('FirstName', StringType(), False),
    StructField('LastName', StringType(), False)
])

data = [
    [1000, 'Matthijs', 'Oosterhout-Buntjes'],
    [1001, 'Joost', 'van Brunswijk'],
    [1002, 'Stan', 'Bokenkamp']
]

customers = spark.createDataFrame(data, schema)
customers.show()

import time
for i in range(100):
    print(f'{i}-Output')
    time.sleep(1)

time.sleep(10)
raise Exception("Demo failure")
```

tinodj:
Closes #1813.

## Description

This pull request addresses the lack of real-time feedback during command execution in the `DatabricksRuntime`.

### Problem

Previously, the output of executed commands was buffered and only displayed after the entire process had completed. 
For long-running tasks, this left the user without any visibility into the command's progress, making it difficult t
determine if the process was still running or had stalled.

### Solution

This PR refactors the execution logic to process output as a stream. This is achieved by:

1.  **Introducing an `onStatusUpdate` callback:** This function is now passed to the `executionContext.execute` 
method.
2.  **Handling Streaming Data:** The callback processes output in chunks as they are generated by the remote process
3.  **Tracking Output Position:** A new `outputPosition` variable tracks the portion of the stream that has already 
been sent to the UI, ensuring that only new output is displayed.

This change provides users with immediate, real-time feedback from their running commands, significantly improving t
user experience.

### Testing Notes

This feature has not yet been covered by automated tests. Manual verification is required.

**To Test:**
1.  Run a long-running command (e.g., a Python script with multiple `print` statements separated by `time.sleep` 
calls).
2.  Confirm that the output appears in the VS Code output panel in real-time as it is generated, rather than all at 
once at the end.


rugpanov:
Hi @tinodj — first, a genuine apology for the long silence here. This PR (and #1813) sat since **early December** without a proper review; the team was stretched thin and PRs didn't get the attention they deserved. That's on us. We now have more bandwidth to work through the backlog and invest in improvements, and I really appreciate your patience — and the fact that you filed a clear issue *and* took a run at fixing it.

On the change itself: the intent is spot-on, but I don't think this approach can deliver streaming output, and I want to explain why so we can steer it toward something that does help.

**The limitation is in the execution API, not the client.** This run path goes through the legacy `1.2/commands` execution API via `ExecutionContext.execute` → `Command`. The `onStatusUpdate` callback you're wiring up *does* fire on every status poll (see `Command.response()` in `src/sdk-extensions/Command.ts`), so that part works — but the callback only ever receives `result.results` populated once the command reaches a **terminal** state (`Finished`/`Error`/`Cancelled`). While `status === "Running"`, `results` is empty, so `onStatusUpdate` has nothing to emit and the output still arrives all at once at the end.

You can see this in the SDK model too: the only output-shaping fields on `Results` are `pos` and `truncated`, which handle **large** results being paginated/size-capped — there's no field carrying incremental stdout produced *over time*. So no client-side callback wiring can surface partial output the API doesn't return. (Real streaming needs API/server-side support, which is a bigger effort on our side.)

A couple of smaller notes:
- The status poll uses exponential backoff capped at ~10s, so even if partial results were available, the cadence wouldn't be truly "real-time."
- `(result: any)` / `(result.results as any).data` drop the typing the surrounding code relies on.

**Good news — there's a way to get live output today:** if you run the file via the **Databricks Connect** run/debug mode instead of "Upload and Run File", execution happens locally through Spark Connect and streams stdout to the integrated terminal in real time. That's the path we'd recommend for interactive, long-running scripts right now, and it's the one we're investing in making frictionless.

**Where I think we can still get value on this path** — #1813 actually describes two problems, and the second one *is* fixable client-side:

1. *Output doesn't appear live* — blocked on API-level streaming support for the Upload-and-Run path; can't be solved from the extension today (use the Databricks Connect mode above in the meantime).
2. *"If the script fails with an exception, none of the printed output appears"* — this is a separate, real bug. On failure the API returns only the traceback and the earlier `print` output is dropped; we can capture stdout in `resources/python/bootstrap.py` and surface it alongside the error.

Also worth noting: this is a duplicate of the older #1325, which reports the same thing — we'll consolidate those.

PR #1852, opened 2026-03-12
Add DABs LSP language client for bundle files

shreyas-goenka:
## Summary

Adds a `BundleLSPClient` that spawns `databricks experimental bundle-lsp` via stdio, providing autocomplete, go-to-definition, hover, document links, and diagnostics for bundle YAML files. Restarts the LSP when the bundle target changes.

Gated behind experimental flag: users must opt in via `databricks.experiments.optInto` with `"features.bundleLsp"`.

**Error handling:** Custom error handler suppresses crash notification popups when the CLI doesn't support the command. No restart storms — immediately gives up on server failure. `revealOutputChannelOn` set to `Never`.

Depends on: databricks/cli#4714

## Test plan
- [x] Extension compiles
- [x] Manual testing with Extension Development Host
- [x] Tested with real deployed bundle (3 targets)

This pull request was AI-assisted by Isaac.

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -1201,6 +1201,7 @@
         "minimatch": "^10.0.1",
         "shell-quote": "^1.8.1",
         "triple-beam": "^1.4.1",
+        "vscode-languageclient": "^9.0.1",
         "winston": "^3.11.0",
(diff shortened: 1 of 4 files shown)


rugpanov:
Close as outdated. Feel free to reopen if it is still relevant.

PR #1871, opened 2026-03-31
Unity catalog explorer

misha-db:
## Changes

# Unity Catalog Features

## Tree View

- Hierarchical browser: Catalogs → Schemas → Tables / Volumes / Functions → Columns
- Filterable by typing (`filterOnType`)
- Only visible when authenticated and extension is active

## Node Types

- **Catalog** — name, tooltip with comment
- **Schema** — name, tooltip with comment
- **Table** — type badge (TABLE / VIEW / MATERIALIZED_VIEW), format (delta, parquet, …), tooltip with owner, timestamps, comment; expandable to columns
- **Volume** — type badge (MANAGED / EXTERNAL), tooltip with owner, comment
- **Function** — full name in tooltip
- **Column** — name, data type, nullable indicator (key icon for non-nullable), comment

## Context Menu Actions

- **Copy** — copies full qualified name (e.g. `catalog.schema.table`) for all nodes; copies just column name for columns
- **Copy storage location** — copies storage path for tables/volumes that have one
- **Copy view SQL** — copies view definition SQL for views and materialized views
- **Open in Databricks** — opens the item in the Databricks Explore UI in a browser (not available for columns)
- **Refresh** — re-fetches a specific node and its children (not available for columns)

## Toolbar

- **Refresh** button — refreshes the entire Unity Catalog tree

## Error Handling

- Error nodes displayed inline when a catalog/schema/table listing fails

## Tests

`UnityCatalogTreeDataProvider.test.ts`


packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.ts
```diff
@@ -0,0 +1,292 @@
...
+                        schema_name: schemaName,
+                    })
+                ),
+                drainAsyncIterable(
+                    client.functions.list({
+                        catalog_name: catalogName,
+                        schema_name: schemaName,
+                    })
+                ),
+            ]);
```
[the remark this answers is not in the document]
misha-db: Added

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.ts
```diff
@@ -0,0 +1,386 @@
...
+    private catalogsCache: UnityCatalogTreeNode[] | undefined = undefined;
+
+    constructor(
+        private readonly connectionManager: ConnectionManager,
+        private readonly stateStorage: StateStorage,
+        private readonly extensionPath: string = ""
+    ) {
+        this.disposables.push(
+            this.connectionManager.onDidChangeState(() => {
+                this.catalogsCache = undefined;
```
anton-107, line 50: do we need to clean the `childrenCache` here as well?

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.ts
```diff
@@ -0,0 +1,386 @@
...
+    private async getChildrenCached(
+        key: string,
+        loader: () => Promise<UnityCatalogTreeNode[]>
+    ): Promise<UnityCatalogTreeNode[]> {
+        const cached = this.childrenCache.get(key);
+        if (cached) {
+            return cached;
+        }
+        const result = await loader();
+        this.childrenCache.set(key, result);
```
anton-107: A transient API blip turns this into a sticky error state that survives until the user manually refreshes that exact node. Detect error/empty results and skip caching, or attach a TTL here.

packages/databricks-vscode/src/ui/unity-catalog/loaders.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+}
+
+export async function loadSchemaChildren(
+    client: Client,
+    catalogName: string,
+    schemaName: string
+): Promise<UnityCatalogTreeNode[]> {
+    const [tablesResult, volumesResult, functionsResult, modelsResult] =
+        await Promise.allSettled([
```
anton-107, line 97: For workspaces with thousands of tables in a schema this is a heavy initial load, and loadSchemaChildren (loaders.ts:107) fires four such drains in parallel before the user sees anything. Consider adding a max-items cap with a "load more" sentinel

packages/databricks-vscode/src/ui/unity-catalog/registerDetailPanel.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    UnityCatalogDetailPanel.setNavigationHandler(showDetail);
+
+    return [
+        telemetry.registerCommand(
+            "databricks.unityCatalog.showDetail",
+            showDetail
+        ),
+        treeView.onDidChangeSelection(async (event) => {
```
anton-107, line 75: Performance concern for big catalogs:

Selection auto-opens the detail panel on every keyboard arrow — `onDidChangeSelection` fires on every focus change, not just clicks. Arrow-key browsing through a large tree will repeatedly create + post to the webview, and on schema/table selection will kick off loadNodeEnrichments (4 parallel API calls each). Consider gating this on an explicit user action or introduce a debounce.

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.ts
```diff
@@ -0,0 +1,386 @@
...
+        const favorites = this.getFavorites();
+        const nodeAsStored = node as StoredFavoriteNode;
+        if (
+            favorites.some((f) => favoriteKey(f) === favoriteKey(nodeAsStored))
+        ) {
+            return;
+        }
+
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const stored: StoredFavoriteNode = {...node} as StoredFavoriteNode;
```
anton-107: This will carry stale metadata as it deep-copies the whole node and stores it in workspace state. 

Owner/comment/format/storageLocation will silently age until the table is re-fetched; renamed or deleted favorites will display indefinitely. Store only `{kind, fullName, version?}` and rehydrate from a fresh fetch when the Favorites root is expanded.

packages/databricks-vscode/src/ui/unity-catalog/detailLoader.ts
```diff
@@ -0,0 +1,437 @@
...
+        name: string;
+        typeName?: string;
+        typeText?: string;
+        comment?: string;
+        nullable?: boolean;
+        position?: number;
+    }>;
+}
+
+async function loadChildrenForNode(
```
anton-107, line 58: Does this re-implement `loaders.ts:loadSchemaChildren` for the no-cache path? Consider deduplicating; otherwise the two implementations will drift

anton-107 (changes requested):

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.ts
misha-db: Yes, added

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.ts
misha-db: Done, skipping caching results with error nodes

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.ts
misha-db: I haven't found API for loading UC resources in bulk by fullnames, meaning if we rehydrate from a fresh fetch it will be N calls (where N is the number of pinned/favorite nodes). this might not be an issue, assuming most users might have 2-5 favorite resources

packages/databricks-vscode/src/ui/unity-catalog/detailLoader.ts
misha-db: cleaned up detailLoader.ts, re-uses loaders, extracting mapping code into mappers.ts

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.ts
```diff
@@ -0,0 +1,459 @@
...
+        const staleKeys = new Set(
+            refs.filter((_, i) => results[i] === null).map(favoriteKey)
+        );
+        if (staleKeys.size > 0) {
+            const cleaned = refs.filter((r) => !staleKeys.has(favoriteKey(r)));
+            await this.stateStorage.set(
+                "databricks.unityCatalog.favorites",
+                cleaned
+            );
+        }
```
anton-107: New issue introduced by the favorites-ref refactor: this prunes any ref whose `loadFavoriteNode` returned `null` and **persists** the removal. But `loadFavoriteNode` wraps everything in `catch { return null }`, so a transient network/5xx/auth error is indistinguishable from a genuinely deleted object — meaning a single API blip while expanding Favorites will permanently delete the user's favorites from workspace state.

This is the same failure mode as the error-caching comment from the last round (transient blip → wrong durable state), but worse here since it's destructive and persisted rather than just a stale cache entry.

Suggestion: only prune on a confirmed "not found" (404). Have `loadFavoriteNode` return a discriminated result, e.g. `{status: "ok", node} | {status: "gone"} | {status: "error"}`, and remove the ref only on `"gone"`. On `"error"`, keep the ref and surface an error/placeholder node instead of silently dropping it.

misha-db: Fixed

anton-107 (approved):

PR #1874, opened 2026-04-13
Workspace Explorer

misha-db:
## Changes

- Enable Workspace Explorer by default (without opting into)
- Add commands for:
  - Open file/folder in browser 
  - Copy path
  - Delete file/folder
  - Upload file
  - downloading file (Open in vscode)

## Tests

Manually


packages/databricks-vscode/src/workspace-fs/WorkspaceFsFileSystemProvider.ts
```diff
@@ -0,0 +1,145 @@
...
+            throw FileSystemError.NoPermissions(uri);
+        }
+
+        // If the file doesn't exist and create is not requested, reject.
+        const existing = await WorkspaceFsEntity.fromPath(client, uri.path);
+        if (!existing && !options.create) {
+            throw FileSystemError.FileNotFound(uri);
+        }
+
+        await parentEntity.createFile(posix.basename(uri.path), content, true);
```
anton-107: should we use `options.overwrite` here instead of hard-coding `true`?

packages/databricks-vscode/src/workspace-fs/WorkspaceFsFileSystemProvider.ts
```diff
@@ -0,0 +1,145 @@
...
+        if (!entity) {
+            throw FileSystemError.FileNotFound(uri);
+        }
+        const isDir =
+            entity.type === "DIRECTORY" || entity.type === "REPO";
+        return {
+            type: isDir ? FileType.Directory : FileType.File,
+            ctime: entity.details.created_at ?? 0,
+            mtime: entity.details.modified_at ?? 0,
+            size: 0,
```
anton-107: is it possible to use an actual file size here as reported by the workspace?

packages/databricks-vscode/src/workspace-fs/WorkspaceFsFileSystemProvider.ts
```diff
@@ -0,0 +1,145 @@
...
+        const entity = await WorkspaceFsEntity.fromPath(client, uri.path);
+        if (!entity) {
+            throw FileSystemError.FileNotFound(uri);
+        }
+        const isDir =
+            entity.type === "DIRECTORY" || entity.type === "REPO";
+        return {
+            type: isDir ? FileType.Directory : FileType.File,
+            ctime: entity.details.created_at ?? 0,
+            mtime: entity.details.modified_at ?? 0,
```
anton-107, line 47: how likely is it that we would default here? i don't think there is a good default that is not breaking the editor heuristics here, but you could consider using `Date.now()` instead of `0`

packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
```diff
@@ -122,6 +128,122 @@ export class WorkspaceFsCommands implements Disposable {
...
+        if (!picked || picked.length === 0) {
+            return;
+        }
+
+        const srcUri = picked[0];
+        const fileName = srcUri.path.split("/").pop() ?? "file";
+        const contentBytes = await workspace.fs.readFile(srcUri);
+
+        try {
+            await root.createFile(fileName, contentBytes, true);
```
anton-107, line 221: shall we prompt for overwrite instead of defaulting to overwriting?  If a user uploads a file into a folder that already has a file with the same name, the remote file is destroyed without warning.

packages/databricks-vscode/src/sdk-extensions/wsfs/WorkspaceFsFile.ts
```diff
@@ -6,7 +6,15 @@ export class WorkspaceFsFile extends WorkspaceFsEntity {
...
 
     override async generateUrl(host: URL): Promise<string> {
-        return `${host.host}#folder/${(await this.parent)?.id ?? ""}`;
+        return `${host.origin}/editor/files/${this.details.object_id}`;
+    }
+
+    async readContent(): Promise<Uint8Array> {
+        const result = await this._workspaceFsService.export({
+            path: this.path,
+            format: "AUTO",
```
anton-107, line 15: i am curious how it works with notebooks files. i believe this would return a source code file without an extension (.ipynb) - could you please double check?

packages/databricks-vscode/src/workspace-fs/WorkspaceFsDataProvider.ts
```diff
@@ -81,6 +94,12 @@ export class WorkspaceFsDataProvider
                         "notebook",
                         new ThemeColor("charts.orange")
                     ),
+                    contextValue: "wsfs.notebook",
+                    command: {
+                        command: "databricks.wsfs.openInBrowser",
```
anton-107, line 103: is it expected that we are opening notebooks in a web browser instead of in the code editor in vscode?

anton-107 (changes requested):

packages/databricks-vscode/src/workspace-fs/WorkspaceFsFileSystemProvider.ts
misha-db: Date.now() would return a different value on every stat call for directories. vscode uses mtime to decide whether to re-read a file. a constantly changing mtime could trigger spurious refresh loops or dirty-state confusion.

packages/databricks-vscode/src/workspace-fs/WorkspaceFsFileSystemProvider.ts
misha-db: Done

packages/databricks-vscode/src/workspace-fs/WorkspaceFsFileSystemProvider.ts
misha-db: Good catch! done

packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
misha-db: Added prompting modal asking for overwrite permission if file with same name already exists

packages/databricks-vscode/src/workspace-fs/WorkspaceFsDataProvider.ts
misha-db: it looks .ipynb are stored in Databricks own format (run/edited on Databricks platform) so when clicking notebook node we deliberately open it in the browser. regular files are opened directly in editor

packages/databricks-vscode/src/sdk-extensions/wsfs/WorkspaceFsFile.ts
misha-db: Yes, notebooks are returned without extension. we open notebooks in browser

misha-db:
@anton-107 @ilia-db 

We register custom file system provider with scheme "wsfs" in extension.
vscode requires it to be unique or activation will fail. it happens so that "Databricks Power tools" is also registering file system provider with same scheme. so depending the order of extension installations one of it will fail. 

Do you remember if it's arbitrary choice or is intentional?

anton-107 (approved):

PR #1877, opened 2026-04-16
 fix: Preserve profile name in Databricks CLI auth provider

misha-db:
## Changes

When the databricks-cli auth type is used with a named profile (e.g. from ~/.databrickscfg), the profile name was not being threaded through to DatabricksCliAuthProvider. As a result `auth login` always used --host, causing issues with new databricks cli when same host exists under multiple profiles.

AuthProvider.ts
- Added optional profile parameter to DatabricksCliAuthProvider constructor
- Threads json.profile / config.profile through both deserialization paths (fromJSON, fromConfig)
- toEnv() now sets DATABRICKS_CONFIG_PROFILE when a profile is present
- toJSON() / toConfig() now persist the profile name so round-trips are lossless

DatabricksCliCheck.ts
- Passes profile in the workspace client config object
- Adds a temporary process.env mutation workaround for the JS SDK not yet supporting profile passthrough to the databricks-cli auth type (marked with a
TODO to remove once the SDK adds support)
- login() now calls auth login --profile <name> when a profile is set, falling back to --host otherwise

Notes:
The process.env["DATABRICKS_CONFIG_PROFILE"] mutation in DatabricksCliCheck is a workaround and should be revisited once the upstream JS SDK exposes a proper API for this.

## Tests

Manually

### DIFF
--- a/packages/databricks-vscode/src/configuration/auth/DatabricksCliCheck.ts
+++ b/packages/databricks-vscode/src/configuration/auth/DatabricksCliCheck.ts
@@ -92,15 +92,6 @@ export class DatabricksCliCheck implements Disposable {
             }
         );
 
-        // TODO: Temporary workaround until the JS SDK supports passing a profile
-        // directly to the databricks-cli auth type. Once supported, remove process.env mutation
-        const profile = this.authProvider.profile;
-        if (profile) {
-            process.env["DATABRICKS_CONFIG_PROFILE"] = profile;
-        } else {
-            delete process.env["DATABRICKS_CONFIG_PROFILE"];
-        }
-
         try {
             await workspaceClient.currentUser.me(
                 new Context({cancellationToken})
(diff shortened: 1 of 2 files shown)


packages/databricks-vscode/src/configuration/auth/DatabricksCliCheck.ts
```diff
@@ -84,13 +84,23 @@ export class DatabricksCliCheck implements Disposable {
...
             },
             {
                 product: "databricks-vscode",
                 productVersion: extensionVersion,
             }
         );
 
+        // TODO: Temporary workaround until the JS SDK supports passing a profile
+        // directly to the databricks-cli auth type. Once supported, remove process.env mutation
+        const profile = this.authProvider.profile;
```
ilia-db: Is this something that's only necessary for the following `currentUser.me` call? Shall we clean it up after it in the `finally` clause?

misha-db: There are other places calling `currentUser.me` after `tryLogin`, doing cleanup directly after breaks them. 
like here: https://github.com/databricks/databricks-vscode/blob/main/packages/databricks-vscode/src/configuration/DatabricksWorkspace.ts#L93

ilia-db (approved):

packages/databricks-vscode/src/configuration/auth/DatabricksCliCheck.ts
misha-db: Moved setting DATABRICKS_CONFIG_PROFILE env var higher to AuthProvider to cover cases switching auth providers

ilia-db (approved):

PR #1879, opened 2026-04-17
Install extension deps from vendored VSIX when available

ilia-db:
## Summary
- When `EXTENSION_VSIX_DIR` env var is set, resolve extension dependencies from local VSIX files instead of VS Code marketplace
- Falls back to marketplace IDs when unset (local dev unchanged)
- Matches platform-specific VSIX (`-linux-x64.vsix`) then falls back to universal (`-universal.vsix`)

Companion PR: https://github.com/databricks-eng/eng-dev-ecosystem/pull/1246

## Test plan
- [ ] Local `yarn run test:integ:extension` still works without `EXTENSION_VSIX_DIR`
- [ ] CI integration tests pass with vendored VSIX files

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -210,7 +210,6 @@
                 "icon": "$(gear)",
                 "title": "Select a Declarative Automation Bundle target",
                 "enablement": "databricks.context.activated && !databricks.context.remoteMode",
-
                 "category": "Databricks"
             },
             {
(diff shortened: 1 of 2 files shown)


misha-db (approved):

anton-107 (approved):
Code review looks good. Clean, well-scoped change.

- Local dev path unchanged (no `EXTENSION_VSIX_DIR` → identical behavior)
- Platform-specific → universal → marketplace fallback is sensible
- Prefix matching works correctly for the current `extensionDependencies` list
- `readdirSync` throwing on missing dir is correct (fail-fast on CI misconfiguration)
- Good logging on both vendored hits and marketplace fallbacks

Minor nit (non-blocking): `readdirSync(vsixDir)` and `targetPlatform` are loop-invariant and could be hoisted above the `for` loop, but irrelevant for test setup code.

LGTM ✅

PR #1889, opened 2026-04-28
Dev QOL improvements

janniklasrose:
## Changes

- gitignore the root databricks.yml file
- Add comment regarding triple slash in `rootConfigSchemaUri`


rugpanov:
Thanks @janniklasrose, and sorry we're only getting to this now! The schema rename landed separately via #1885 (`a91942c`), so that part is now identical to `main` and is causing the conflict here.

The other two changes are still useful — the `.gitignore` `/databricks.yml` entry and the template doc-link fix (the old `#databricks-asset-bundle-project-templates` anchor no longer exists on the docs page). Could you rebase onto latest `main`? The schema hunks should drop out cleanly, leaving just those. Happy to review once it's green. 🙏

rugpanov (approved):

PR #1893, opened 2026-05-04
Fix new profile sign in using already existing host under different profile

misha-db:
## Changes

Fix new profile sign in using already existing host under different profile

## Tests

Manually

### DIFF
--- a/packages/databricks-vscode/src/configuration/LoginWizard.ts
+++ b/packages/databricks-vscode/src/configuration/LoginWizard.ts
@@ -311,7 +311,8 @@ export class LoginWizard {
                 authProvider = new DatabricksCliAuthProvider(
                     this.state.host!,
                     this.cliWrapper.cliPath,
-                    this.cliWrapper
+                    this.cliWrapper,
+                    profileName
                 );
                 break;
 


anton-107 (approved):
Approve. Minimal, targeted fix — passing the new profile name through to `DatabricksCliAuthProvider` ensures `tryLogin` queries the not-yet-existing profile (correctly fails) and `login` runs `databricks auth login --profile <newName>`, preventing implicit credential reuse from a pre-existing profile sharing the same host.

A couple of non-blocking observations for follow-up:

- In `DatabricksCliCheck.login` (`auth/DatabricksCliCheck.ts:108-115`), when `profile` is set we pass only `--profile`, not `--host`. For a brand-new profile not yet in `.databrickscfg`, it would be safer to pass both `--profile` and `--host` so the CLI doesn't have to infer/prompt for the host. Manual testing covers it, but worth confirming.
- `saveNewProfile` appends a profile section to `.databrickscfg` after the CLI's `auth login` already wrote one — pre-existing behavior, but more visible now that both writes use the same profile name. Possible duplicate `[profileName]` block.
- An automated test covering "new profile name + host already used by another profile" would be cheap insurance against regressions of exactly this bug class.

PR #1894, opened 2026-05-07
 Include profiles with account_id in listProfiles results

misha-db:
## Changes

Previously, `listProfiles` filtered out any profile with `account_id` set, using it as a signal to distinguish account-level profiles from  workspace-level ones. This assumption broke when the Databricks CLI began populating `account_id` on workspace profiles as well  causing valid workspace profiles to be silently dropped from the list.

Removing the filter restores correct behaviour: all profiles are returned
regardless of whether `account_id` is present.

## Tests

Unit test: CliWrapper.test.ts

### DIFF
--- a/packages/databricks-vscode/src/cli/CliWrapper.test.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.test.ts
@@ -169,6 +169,34 @@ nothing = true
         });
     });
 
+    it("should include profiles with account id", async () => {
+        const logFilePath = getTempLogFilePath();
+        const cli = createCliWrapper(logFilePath);
+
+        await withFile(async ({path}) => {
+            writeFileSync(
+                path,
+                `[regular-profile]
+host = https://cloud.databricks.com/
+token = dapitest1234
+
+[profile-with-accunt-id]
+host = https://accounts.cloud.databricks.com/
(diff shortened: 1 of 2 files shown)


anton-107:
**Could you link the SDK / CLI change that motivates this?** My understanding is that the upstream change started populating `account_id` on workspace profiles too, which would invalidate the original filter. If that's the case, please reference the relevant SDK/CLI PR or release in the description so reviewers (and future readers via `git blame`) can see why the filter is no longer a reliable workspace/account discriminator.

**Consider keeping the filter and moving it, rather than removing it.** `CliWrapper.listProfiles` is the lowest layer and every caller now receives account profiles. If the goal is "the login wizard should show account-tagged workspace profiles," the wrapper should still reflect raw CLI output, and the workspace-vs-account policy should live at the call site (e.g. inside `LoginWizard.listProfiles`), filtering on a more reliable signal than `account_id` — e.g. host shape (`accounts.*`) or the SDK's own classification. That keeps the wrapper neutral and prevents account profiles from leaking into any future caller that expects workspace-only profiles.

**Nit:** typo `profile-with-accunt-id` in the new test (used twice).

anton-107 (approved):

PR #1905, opened 2026-06-12
Fix Python environment issues in the Databricks Connect run/debug flow

anton-107:
## Changes

Quick wins from a live CUJ run (2026-06-12, fresh macOS profile, default-python bundle project, serverless): every "run/debug with Databricks Connect" capability works, but users get lost in the Python environment setup in between. This PR fixes the three failure modes that don't depend on any larger architectural decision. A follow-up PR adds opt-in managed environment provisioning (see design note below).

### 1. Run/debug interpreter split-brain (P0)

"Run with Databricks Connect" launches `dbconnect-bootstrap.py` with `MsPythonExtensionWrapper.getPythonExecutable()` (the active environment for the project), but "Debug with Databricks Connect" started a `debugpy` session **without a `python` attribute**, so debugpy fell back to the MS Python extension's folder-level interpreter selection. With a system Python 3.9 selected there, debug crashed with a raw `ModuleNotFoundError: No module named 'databricks'` traceback while run worked fine.

- `RunCommands.debugFileUsingDbconnect` now resolves the same executable as the run path and pins it via the debug configuration's `python` attribute.
- `checkDbconnectEnabled` (shared run+debug preflight) now force-re-verifies the environment when the selected interpreter changed since the cached check, so a mismatch surfaces the normal actionable setup flow instead of a bootstrap traceback.

### 2. Pip-less venvs broke the extension itself (P0)

The MS Python extension's env-creation flow can produce a venv without pip. Our own introspection (`python -m pip list --format json`) then failed with "No module named pip", taking the whole "Python Environment" checklist down with it. Pip commands now seed pip via `python -m ensurepip --upgrade` and retry once when they fail with "No module named pip" (native pip path only; uv-managed environments are not affected). Pip command failures now also include the stderr tail instead of just "Command exited with code 1".

### 3. "Python 3.12 or greater" is the wrong requirement (P1)

Databricks Connect needs the local Python *minor* version to match the remote one (UDF/pickle compatibility). Python 3.13 satisfied the old "3.12 or greater" check for serverless and then broke at runtime.

- The DBR→Python mapping now lives in one table (`computeTargetSpec.ts`: 13/14→3.10, 15→3.11, 16+→3.12), shared by the check and the messaging.
- Serverless now requires an exact minor match, and the message explains why ("the local minor version must match the Python version of the serverless environment").
- Cluster mismatches keep warning-level reporting (non-UDF workloads still work), with corrected copy.
- The "Select Python Environment" quick pick and the "Create new environment" handoff to the MS Python extension now surface the version requirement, since the generic creation flow knows nothing about it (its "Recommended" Quick Create happily offers Python 3.9).

### Current flow map (who resolves the interpreter where)

| Path | Interpreter resolution (before this PR) |
|---|---|
| Verification checklist (`EnvironmentDependenciesVerifier`) | MS Python API `getActiveEnvironmentPath(activeProjectUri)`, resolved |
| Run with dbconnect (`runFileUsingDbconnect`) | same as verification (`getPythonExecutable()`), sent to a terminal |
| Debug with dbconnect (`debugFileUsingDbconnect`) | **debugpy's own fallback** = MS Python selected interpreter for the workspace folder |
| Env creation ("Create new environment") | delegated to `python.createEnvironment`, unaware of our version constraint |
| Dependency install | `python -m pip` (or `uv pip` if `uv.lock` present) in the active env |

This PR makes the first three rows resolve identically and verifies freshness preflight.

### Compute → Python/dbconnect version matrix

| Compute | Python (local) | databricks-connect |
|---|---|---|
| Serverless (`serverlessDbconnectVersion` = 17.3) | 3.12 — exact match required | `17.3.*` |
| Serverless 15.4 | 3.11 — exact | `15.4.*` |
| Cluster DBR 13/14 | 3.10 (warning on mismatch) | `<major>.<minor>.*` |
| Cluster DBR 15 | 3.11 (warning) | `<major>.<minor>.*` |
| Cluster DBR 16+ | 3.12 (warning) | `<major>.<minor>.*` |

### Follow-up: managed environment provisioning (separate PR, experimental)

The remaining CUJ failures (env creation handoff, opaque pip failures behind corp proxies, interpreter discovery) are addressed by an opt-in managed flow (`databricks.experiments.optInto: ["python.managedEnvironment"]`):

- **Engine: uv**, acquired PATH-first (respects corp-managed installs and their proxy/mirror config), with download-on-demand into `globalStorageUri` (pinned version, SHA256-verified, mirroring the `fetch-databricks-cli.sh` pattern) and graceful fallback to the current manual flow. Bundling uv in the vsix was considered and deferred: it adds ~15–20MB per platform and doesn't rescue offline networks (interpreter and package downloads would fail anyway).
- **Flow** (single cancellable progress notification): resolve target versions from the table above → `uv python install <X.Y>` if needed → `uv venv .venv --python <X.Y> --seed` (seeded pip keeps the existing introspection working) → `uv pip install databricks-connect==<matched> nbformat` + project deps (`requirements*.txt` / `pyproject.toml`) → set the MS Python active environment so run, debug, terminal and status bar all agree.
- **Respecting user environments**: a satisfied env is never touched; venvs we created are tagged with a marker file and can be repaired/recreated silently; foreign `.venv`s are only modified after an explicit prompt (Repair / Recreate / Set up manually).
- **Failure UX**: failures are classified (network blocked / no matching interpreter / disk / cancelled) into actionable messages with Retry, honoring `UV_INDEX_URL`, `PIP_INDEX_URL`, `pip.conf` index-url, and `UV_PYTHON_INSTALL_MIRROR`; telemetry per funnel step so drop-off becomes measurable.
- **Risks**: `uv python install` downloads from GitHub (blocked corps need the mirror env var — called out in the error message); MS Python extension can lag on env discovery (mitigated by `refreshEnvironments()` before selection); the flag default-off contains KTLO risk.

Related prior art: #1886 (thanks @twsl) extends the same verifier to support global interpreters — different scope, no conflict in intent; happy to rebase whichever lands second.

## Tests

- New unit tests: `computeTargetSpec.test.ts` (full version matrix), `EnvironmentDependenciesVerifier.test.ts` (serverless exact-match incl. 3.13 rejection, cluster warning/rejection, copy), `MsPythonExtensionWrapper.test.ts` (ensurepip seeding semantics), `RunCommands.test.ts` (debug config pinning, stale-interpreter re-verification).
- `yarn test:lint` and `yarn test:unit` pass (189 passing).
- Manual: macOS, default-python bundle, serverless — with system Python 3.9 selected, debug now fails preflight with the actionable setup flow instead of a bootstrap traceback; after switching to `.venv`, run and debug both use `.venv`.

### DIFF
--- a/packages/databricks-vscode/src/language/EnvironmentCommands.ts
+++ b/packages/databricks-vscode/src/language/EnvironmentCommands.ts
@@ -1,5 +1,8 @@
 import {window, commands, QuickPickItem, ProgressLocation} from "vscode";
-import {FeatureManager} from "../feature-manager/FeatureManager";
+import {
+    FeatureManager,
+    FeatureStepState,
+} from "../feature-manager/FeatureManager";
 import {MsPythonExtensionWrapper} from "./MsPythonExtensionWrapper";
 import {Cluster} from "../sdk-extensions";
 import {EnvironmentDependenciesInstaller} from "./EnvironmentDependenciesInstaller";
@@ -77,14 +80,31 @@ export class EnvironmentCommands {
     async selectPythonInterpreter() {
         const environments =
             await this.pythonExtension.getAvailableEnvironments();
+        const state = await this.featureManager.isEnabled(
+            "environment.dependencies"
+        );
+        const pythonStep = state.steps.get("checkPythonEnvironment");
+        const requirement = !pythonStep?.available ? pythonStep : undefined;
         if (environments.length > 0) {
-            await this.showEnvironmentsQuickPick(environments);
+            await this.showEnvironmentsQuickPick(environments, requirement);
         } else {
-            await this.pythonExtension.createPythonEnvironment();
+            await this.createPythonEnvironment(requirement);
         }
     }
 
-    async showEnvironmentsQuickPick(environments: Environment[]) {
+    private async createPythonEnvironment(requirement?: FeatureStepState) {
+        if (requirement?.message) {
+            // The environment creation flow of the MS Python extension knows
+            // nothing about our version requirements, so we surface them here.
+            window.showInformationMessage(requirement.message);
+        }
+        await this.pythonExtension.createPythonEnvironment();
+    }
+
+    async showEnvironmentsQuickPick(
+        environments: Environment[],
+        requirement?: FeatureStepState
+    ) {
         const envPicks: (QuickPickItem & {path?: string})[] = environments.map(
             (env) => ({
                 label: environmentName(env),
@@ -101,11 +121,14 @@ export class EnvironmentCommands {
         ];
         const selectedPick = await window.showQuickPick(
             envPicks.concat(staticPicks),
-            {title: "Select Python Environment"}
+            {
+                title: requirement?.title ?? "Select Python Environment",
+                placeHolder: requirement?.message,
+            }
         );
         if (selectedPick) {
             if (selectedPick.label === createNewLabel) {
-                await this.pythonExtension.createPythonEnvironment();
+                await this.createPythonEnvironment(requirement);
             } else if (selectedPick.label === usePythonExtensionLabel) {
                 await this.pythonExtension.selectPythonInterpreter();
             } else if (selectedPick.path) {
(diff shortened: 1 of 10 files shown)


anton-107:
Follow-up commit d3daba4 addresses self-review findings: environments with an unresolvable Python version are accepted again (historic behavior), the stale-interpreter re-check is skipped while disconnected (it blocks on an established connection), ensurepip failures rethrow the original error, the setup command now returns whether the environment is ready (so run/debug can proceed right after a successful setup), and the interpreter-picker requirement hint is fetched with a timeout. Unit suite: 191 passing.

This comment was written by Isaac.

PR #1903, opened 2026-06-04
Pass DATABRICKS_CLI_PATH to subprocess environment in DatabricksCliAuthProvider

DM87:
## Summary

Fixes #1896

Bundle operations (deploy, validate) fail with `databricks CLI not found` since v2.10.7 because the spawned CLI subprocess cannot locate itself for token refresh.

## Root cause

The CLI bump from v0.286.0 to v0.297.2 brought Go SDK v0.126.0, which changed the `databricks-cli` auth strategy to invoke the CLI binary (via `cli_token_source.go` → `findDatabricksCli()`) instead of refreshing tokens in-process.

`DatabricksCliAuthProvider.toEnv()` sets `DATABRICKS_AUTH_TYPE=databricks-cli` but never sets `DATABRICKS_CLI_PATH`. The spawned CLI sees an empty `DatabricksCliPath`, falls back to searching `PATH`, and fails because the extension's `bin/` directory isn't on the system PATH.

The in-process SDK config (`_getSdkConfig()`) already passes `databricksCliPath: this.cliPath` correctly — the subprocess environment just needs the same treatment.

## Changes

- Add `DATABRICKS_CLI_PATH: this.cliPath` to `toEnv()` in `DatabricksCliAuthProvider`

## Test plan

- [ ] Untested — submitting based on code analysis only
- [ ] Existing unit tests pass
- [ ] Bundle deploy no longer errors with "databricks CLI not found"

PR #1908, opened 2026-06-12
Pass DATABRICKS_CLI_PATH and resolve the platform-specific CLI binary

rugpanov:
## Why
- On Windows, `databricks bundle` deploys fail with `databricks CLI not found` because the SDK / Terraform provider can't locate the bundled CLI when running in a subprocess.
- The path that was forwarded was extensionless (`bin/databricks`), but the bundled Windows binary is `databricks.exe`. The Go SDK / Terraform provider do a literal file lookup and do **not** auto-append `.exe` the way Windows' `CreateProcess` does — so the lookup fails.
- The legacy `.databricks/project.json` persisted the bundled CLI path (`databricksPath`), and `AuthProvider.fromJSON` preferred it over the freshly resolved path. That snapshot is version-pinned (e.g. `.../databricks.databricks-2.11.0/bin/...`) and, on Windows, was stored without `.exe`, so it overrode the corrected path during legacy→bundle migration.

This builds on #1903 (which adds the `DATABRICKS_CLI_PATH` env var) by also making the path it forwards correct on Windows and not letting a stale persisted path win.

## What
- Forward the resolved CLI path to subprocesses via the `DATABRICKS_CLI_PATH` env var so they don't fall back to a `PATH` search (`DatabricksCliAuthProvider.toEnv`).
- Make `CliWrapper.cliPath` return the platform-specific binary name: `databricks.exe` on `win32`, `databricks` elsewhere.
- In `AuthProvider.fromJSON`, always use the freshly resolved bundled CLI path and ignore the persisted `databricksPath` — it's an extension-managed, version-pinned value that is stale after any upgrade. (`fromSdkConfig` is intentionally left alone, since its `databricksCliPath` comes live from the user's `.databrickscfg`.)
- Added explanatory comments at each site.

## Verification
- Added unit tests for: the platform-specific binary name, `toEnv()` exposing `DATABRICKS_CLI_PATH`, and `fromJSON` ignoring a stale persisted `databricksPath`. Suite passes in the VS Code test host (14 relevant tests green).
- Built the `win32-x64` VSIX and confirmed the bundled `extension/out/extension.js` contains both the env var and the `.exe` path, with `bin/databricks.exe` packaged.

### DIFF
--- a/packages/databricks-vscode/src/cli/CliWrapper.test.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.test.ts
@@ -53,6 +53,33 @@ describe(__filename, function () {
         assert.ok(result.stdout.indexOf("databricks") > 0);
     });
 
+    it("should resolve the platform-specific CLI binary name", () => {
+        const cli = createCliWrapper();
+        const originalPlatform = process.platform;
+        const setPlatform = (platform: NodeJS.Platform) =>
+            Object.defineProperty(process, "platform", {value: platform});
+        try {
+            // On Windows the bundled binary is `databricks.exe`. The `.exe` is
+            // required because cliPath is forwarded to the SDK/Terraform via
+            // DATABRICKS_CLI_PATH, which does a literal (no auto-`.exe`) lookup.
+            setPlatform("win32");
+            assert.ok(
+                cli.cliPath.endsWith(path.join("bin", "databricks.exe")),
+                `expected win32 cliPath to end with bin/databricks.exe, got ${cli.cliPath}`
+            );
+
+            for (const platform of ["linux", "darwin"] as NodeJS.Platform[]) {
+                setPlatform(platform);
+                assert.ok(
+                    cli.cliPath.endsWith(path.join("bin", "databricks")),
+                    `expected ${platform} cliPath to end with bin/databricks, got ${cli.cliPath}`
+                );
+            }
+        } finally {
+            setPlatform(originalPlatform);
+        }
+    });
+
     let mocks: any[] = [];
     afterEach(() => {
         mocks.forEach((mock) => reset(mock));
--- a/packages/databricks-vscode/src/cli/CliWrapper.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.ts
@@ -298,7 +298,15 @@ export class CliWrapper {
     }
 
     get cliPath(): string {
-        return this.extensionContext.asAbsolutePath("./bin/databricks");
+        // The bundled binary is named `databricks.exe` on Windows. We must
+        // include the extension here: while spawning the CLI ourselves works
+        // without it (Windows' CreateProcess auto-appends `.exe`), this path is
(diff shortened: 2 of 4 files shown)


anton-107:
On the persisted `databricksPath` path:

`DatabricksCliAuthProvider.toJSON()` writes `databricksPath: this.cliPath` into `.databricks/project.json`, and `AuthProvider.fromJSON()` reads it back as `json.databricksPath ?? cli.cliPath`. So for a Windows user who connected *before* this fix, `project.json` already holds an extensionless `…/bin/databricks`, and after upgrading the `??` won't fall through to the new `.exe`-aware `cli.cliPath` — meaning `toEnv()` would forward the stale extensionless path and reproduce the exact `databricks CLI not found` this PR fixes.

Two things I could not confirm from the diff:

1. Is `project.json` rewritten with a fresh `cli.cliPath` on every connect/activation? The persisted value is also absolute and version-pinned (`databricks.databricks-<version>/bin/...`), which would be stale across *any* upgrade on *any* platform — so I suspect it does get refreshed, but I'd like to confirm.
2. If so, is there a window where `toEnv()` runs against a freshly-loaded provider *before* that rewrite happens? If yes, existing Windows users would still hit the failure on the first connect post-upgrade.

If the persisted path can win over the getter here, it might be worth ignoring/normalizing `databricksPath` for the bundled-CLI case so `CliWrapper.cliPath` stays the single source of truth. Could you confirm the rewrite timing?

rugpanov:
Good catch. Digging in, project.json turns out to be an outdated way we used to persist config - nothing writes it anymore. But for the small fraction of users who still have one lying around, it's read once during the legacy -> bundle migration (BundleProjectManager → ProjectConfigFile.load → fromJSON), and that's exactly where this can fail.

On your questions: nothing rewrites or revalidates the file, so a persisted value is taken as-is. It only falls through when the field is absent, it never checks whether the path still exists. And since the stored path is absolute and points into the versioned extension folder `(…databricks.databricks-2.11.0/bin/…)`, it's already pointing at the wrong install after an upgrade. This doesn't hit modern users, though: once a databricks.yml exists, auth comes from a ProfileAuthProvider built off the profile name, which resolves cli.cliPath fresh and never touches databricksPath. So it only affects the small set still on an un-migrated project.json, and the migration is the only window - which is why the fix has to live on the read side.

So I did what you suggested and made cli.cliPath the single source of truth: fromJSON now ignores databricksPath (commit 0534870), with a test that feeds a stale extensionless path plus a fresh .exe one and asserts the getter wins. I left fromSdkConfig alone since its databricksCliPath comes live from the user's .databrickscfg - intentional config rather than a stale snapshot - but happy to change that too if you'd prefer.

anton-107 (approved):

rugpanov:
Superseded by https://github.com/databricks/databricks-vscode/pull/1910, which runs from an upstream branch so CI (JFrog OIDC) can run — fork PRs don't get OIDC tokens/secrets, so the checks here couldn't execute. Same commit and contents.

PR #1912, opened 2026-06-16
Fix Databricks Connect run/debug interpreter split-brain

rugpanov:
## Why
- **"Debug with Databricks Connect" used the wrong interpreter.** It started a `debugpy` session **without a pinned interpreter**, so debugpy fell back to the Python extension's folder-level selection — which can differ from the environment used by "run" and by the verification checks. With e.g. a system Python 3.9 selected there, debug crashed with a raw `ModuleNotFoundError: No module named 'databricks'` while run worked fine.
- **Stale interpreter checks.** The cached environment state can refer to a previously selected interpreter, because the Python extension doesn't always notify us when the interpreter changes — so a mismatch surfaced as a bootstrap traceback instead of the actionable setup flow.

## What
- Pin debugpy to the verified interpreter via the debug configuration's `python` attribute (`DatabricksPythonDebugConfiguration.python`).
- Resolve the interpreter once in a shared run/debug preflight (`resolveDbconnectLaunch`), so run and debug provably use the **same** executable (removes the duplicated preamble).
- Re-verify the environment before launch when the selected interpreter changed since the cached check — only while `CONNECTED`, since a re-check blocks on the connection.
- After an in-flow environment setup, re-check and let the launch **proceed** instead of aborting and forcing the user to re-trigger.

## Verification
- New/updated unit tests in `RunCommands.test.ts`: debug-config pinning, stale-interpreter re-verification, no-reverify-while-disconnected, and proceed-after-setup. RunCommands suite passes (6 tests) in the VS Code test host.
- `eslint` and `prettier` clean on the changed files.

## Notes
This is the first of three slices carved out of #1905 (run/debug interpreter split-brain). The pip-less-venv and Python-version-requirement fixes follow in separate PRs.

### DIFF
--- a/packages/databricks-vscode/src/run/DatabricksDebugConfigurationProvider.ts
+++ b/packages/databricks-vscode/src/run/DatabricksDebugConfigurationProvider.ts
@@ -13,6 +13,8 @@ export interface DatabricksPythonDebugConfiguration extends DebugConfiguration {
     env?: Record<string, any>;
     isInternalDatabricksRun?: boolean;
     console?: "integratedTerminal" | "externalTerminal" | "internalConsole";
+    /** Path to the python interpreter debugpy should use */
+    python?: string;
 }
 
 function isTest(debugConfiguration: DebugConfiguration) {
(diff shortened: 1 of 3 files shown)


rugpanov:
(EMU) Integration tests run: https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/27627073249

rugpanov:
(EMU) Integration tests run: https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/27627073249

misha-db (approved):

PR #1918, opened 2026-06-19
[telemetry] Detect Python package manager(s) at project setup

rugpanov:
## Changes

Measurement-only telemetry to learn which Python package manager(s) our users' projects actually use (pip / conda / uv / poetry), so the VPEX setup-flow investment can be prioritized from first-party data instead of public-survey estimates. No setup behavior changes — this is detection only.

The work splits cleanly into three layers so each is independently testable and the dependency direction stays correct (high-level → low-level):

- **Pure classifier** (`packageManagerDetection.ts`): given a set of already-collected signals, reports every applicable manager, a best-guess primary (priority `uv > poetry > conda > pip`), the firing signals, `hasLockfile`, and interpreter source. Side-effect free and total.
- **Emit** (`telemetry/packageManagerExtensions.ts`): adds `recordPackageManagerDetection` to the existing `Telemetry` class via the same `declare module` pattern as `commandExtensions.ts`. Keeps disk/Python-extension dependencies out of the telemetry client.
- **Collection** (`PackageManagerTelemetry.ts`): a best-effort, non-blocking collector that reads disk and already-resolved interpreter metadata, runs the pure classifier, and calls the emit method. Deduplicated per session on `(trigger, projectRoot)`; any failure degrades to `unknown` and is swallowed so it never disrupts setup.

Emission is wired into three setup touchpoints: project-open environment check (`auto_open`), the set-up-environment command (`explicit_command`), and first Run/Debug with Databricks Connect (`run`/`debug`).

A new `Events.PYTHON_ENV_SETUP_DETECTED` event carries a typed, documented schema (reuses the existing telemetry transport; opt-out honored; categorical data only — no paths, package names, or cluster names). A handoff note for the analytics/dashboard owner is included at `src/telemetry/PACKAGE_MANAGER_DETECTION.md`.

**Detection correctness** (the parts most worth reviewing):
- `interpreterSource` is derived from the active interpreter alone, never from project files. A `uv.lock` project running a conda/venv/system interpreter reports that interpreter's real source, keeping the "uv project, interpreter not uv-managed yet" setup-flow gap visible. A genuinely uv-provisioned venv is identified by the `uv =` marker in `pyvenv.cfg`, not by `uv.lock`.
- conda is attributed only when the active interpreter resides under `CONDA_PREFIX` (path-boundary checked), not on the bare env var — which is session-global in the extension host (launching VS Code from an activated conda shell) and would otherwise over-count conda for uv/poetry/pip projects.
- `pyproject` `[tool.uv]`/`[tool.poetry]` detection uses a bounded table-header scan, not substring matching: ignores comments and in-value mentions, rejects prefix collisions (e.g. `tool.uvicorn`), and matches subtable and array-of-table headers (`[tool.uv.sources]`, `[[tool.poetry.source]]`).
- No external executable is run for telemetry: the uv-on-PATH probe was removed (it spawned a PATH-resolved `uv` for a weak, non-attributing signal). Detection reads only disk and already-resolved interpreter metadata.

**Scope / privacy:** measurement only — no changes to setup behavior (the VPEX flows are a separate effort). Only enum/categorical data and a closed set of signal identifiers are emitted; the existing telemetry opt-out (`telemetry.telemetryLevel`) is respected by the transport.

## Tests

- [x] `yarn run test:unit`: 202 passing, 0 failing — includes the pure classifier (each manager, interpreter sources, overlaps like uv+pip / conda+pip / poetry+uv, weak signals, none) and pure helpers (`pyprojectHasToolSection`, `pyvenvCfgMarksUv`, `interpreterUnderCondaPrefix`), covering the conda-prefix boundary and shell-global false-positive cases.
- [x] `yarn run build` (typecheck) passes.
- [x] `eslint` clean; `prettier` formatted.

Reviewer can validate with:

```bash
cd packages/databricks-vscode
yarn run build
yarn run test:unit
npx eslint src --ext ts && npx prettier . -c
```

### DIFF
--- a/packages/databricks-vscode/src/language/EnvironmentDependenciesVerifier.ts
+++ b/packages/databricks-vscode/src/language/EnvironmentDependenciesVerifier.ts
@@ -10,6 +10,7 @@ import {ResolvedEnvironment} from "./MsPythonExtensionApi";
 import {NamedLogger} from "@databricks/sdk-experimental/dist/logging";
 import {ConfigureAutocomplete} from "./ConfigureAutocomplete";
 import {workspaceConfigs} from "../vscode-objs/WorkspaceConfigs";
+import {PackageManagerTelemetry} from "./PackageManagerTelemetry";
 
 export class EnvironmentDependenciesVerifier extends MultiStepAccessVerifier {
     private readonly logger = NamedLogger.getOrCreate(Loggers.Extension);
@@ -18,7 +19,8 @@ export class EnvironmentDependenciesVerifier extends MultiStepAccessVerifier {
         private readonly connectionManager: ConnectionManager,
         private readonly pythonExtension: MsPythonExtensionWrapper,
         private readonly installer: EnvironmentDependenciesInstaller,
-        private readonly configureAutocomplete: ConfigureAutocomplete
+        private readonly configureAutocomplete: ConfigureAutocomplete,
+        private readonly packageManagerTelemetry: PackageManagerTelemetry
     ) {
         super([
             "checkCluster",
@@ -403,6 +405,9 @@ export class EnvironmentDependenciesVerifier extends MultiStepAccessVerifier {
     }
 
     override async check() {
+        // First environment check on project open: emit package-manager
+        // detection telemetry (deduplicated per session, never throws).
+        void this.packageManagerTelemetry.emitDetection("auto_open");
         await this.connectionManager.waitForConnect();
         await Promise.all([
             this.checkCluster(this.connectionManager.cluster),
--- a/packages/databricks-vscode/src/language/EnvironmentDependenciesVerifier.ts
+++ b/packages/databricks-vscode/src/language/EnvironmentDependenciesVerifier.ts
@@ -405,10 +405,11 @@ export class EnvironmentDependenciesVerifier extends MultiStepAccessVerifier {
     }
 
     override async check() {
-        // First environment check on project open: emit package-manager
-        // detection telemetry (deduplicated per session, never throws).
-        void this.packageManagerTelemetry.emitDetection("auto_open");
         await this.connectionManager.waitForConnect();
+        // Emit package-manager detection only once connected (waitForConnect
+        // resolves on CONNECTED), so unauthenticated sessions are not reported.
+        // Deduplicated per session; never throws.
+        void this.packageManagerTelemetry.emitDetection("auto_open");
         await Promise.all([
             this.checkCluster(this.connectionManager.cluster),
             this.checkWorkspaceHasUc(),
--- a/packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
+++ b/packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
@@ -54,6 +54,12 @@ export class PackageManagerTelemetry {
      */
     async emitDetection(trigger: SetupTrigger): Promise<void> {
         try {
+            // Bail before any disk access when telemetry is disabled, so an
+            // opted-out user gets zero telemetry-driven file reads (not just a
+            // dropped send).
+            if (!this.telemetry.isTelemetryEnabled) {
+                return;
+            }
             const projectRoot = this.getProjectRoot();
             if (projectRoot === undefined) {
                 return;
@@ -233,12 +239,18 @@ export class PackageManagerTelemetry {
         }
     }
 
-    /** True if any `requirements*.txt` file exists in the project root. */
+    /**
+     * True if any pip-style requirements file exists in the project root:
+     * `requirements.txt` or a separator-suffixed variant such as
+     * `requirements-dev.txt` / `requirements_test.txt` / `requirements.ci.txt`.
+     * Deliberately does not match `requirementsfoo.txt` (no separator), which
+     * isn't a conventional requirements file.
+     */
     private hasRequirementsTxt(projectRoot: string): boolean {
         try {
             return fs
                 .readdirSync(projectRoot)
-                .some((name) => /^requirements.*\.txt$/.test(name));
+                .some((name) => /^requirements([-_.].+)?\.txt$/.test(name));
         } catch (e) {
             this.logger.debug("Failed to list project root", e);
             return false;
--- a/packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
+++ b/packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
@@ -1,0 +1,251 @@
+import fs from "node:fs";
+import path from "node:path";
+import {NamedLogger} from "@databricks/sdk-experimental/dist/logging";
+import {Loggers} from "../logger";
+import {Telemetry} from "../telemetry";
+import {ComputeType, SetupTrigger} from "../telemetry/constants";
+import "../telemetry/packageManagerExtensions";
+import {MsPythonExtensionWrapper} from "./MsPythonExtensionWrapper";
+import {ResolvedEnvironment} from "./MsPythonExtensionApi";
+import {
+    detectPackageManagers,
+    interpreterUnderCondaPrefix,
+    InterpreterSource,
+    PackageManagerSignals,
+    pyprojectHasToolSection,
+    pyvenvCfgMarksUv,
+} from "./packageManagerDetection";
+
+export type {SetupTrigger};
+
+/**
+ * Collects package-manager signals at project-setup touchpoints and emits the
+ * {@link Events.PYTHON_ENV_SETUP_DETECTED} telemetry event.
+ *
+ * All probing is best-effort and non-blocking: any failure degrades to
+ * `unknown` and is swallowed, never thrown into the user's setup/run flow.
+ * Only categorical/enum data is emitted — no paths, package names, or other
+ * free-form content (see {@link detectPackageManagers}). Telemetry opt-out is
+ * honoured by the underlying {@link Telemetry} client.
+ */
+export class PackageManagerTelemetry {
+    private readonly logger = NamedLogger.getOrCreate(Loggers.Extension);
+
+    /**
+     * Triggers already emitted for the current `(project, trigger)` pair, to
+     * deduplicate within a session so one project open doesn't inflate counts.
+     */
+    private readonly emitted = new Set<string>();
+
+    constructor(
+        private readonly telemetry: Telemetry,
+        private readonly pythonExtension: MsPythonExtensionWrapper,
+        private readonly getProjectRoot: () => string | undefined,
+        private readonly getComputeType: () => ComputeType | "none"
+    ) {}
+
+    /**
+     * Detect the package manager(s) for the active project and emit telemetry.
+     * Deduplicated per `(project root, trigger)` within the session. Never
+     * throws.
+     */
+    async emitDetection(trigger: SetupTrigger): Promise<void> {
+        try {
+            const projectRoot = this.getProjectRoot();
+            if (projectRoot === undefined) {
+                return;
+            }
+            const dedupeKey = `${trigger}:${projectRoot}`;
+            if (this.emitted.has(dedupeKey)) {
+                return;
+            }
+            this.emitted.add(dedupeKey);
+
+            const env = await this.resolveEnvironment();
+            const signals = this.collectSignals(projectRoot, env);
+            const detection = detectPackageManagers(signals);
+
+            this.telemetry.recordPackageManagerDetection(detection, {
+                pythonVersion: this.getPythonMinorVersion(env),
+                targetCompute: this.getComputeType(),
+                trigger,
+            });
+        } catch (e) {
+            // Detection is measurement-only and must never disrupt setup.
+            this.logger.debug("Package manager detection failed", e);
+        }
+    }
+
+    private async resolveEnvironment(): Promise<
+        ResolvedEnvironment | undefined
+    > {
+        try {
+            return await this.pythonExtension.pythonEnvironment;
+        } catch (e) {
+            this.logger.debug("Failed to resolve python environment", e);
+            return undefined;
+        }
+    }
+
+    /** Detected interpreter minor version (e.g. "3.11"), if available. */
+    private getPythonMinorVersion(
+        env: ResolvedEnvironment | undefined
+    ): string | undefined {
+        const version = env?.version;
+        if (version?.major === undefined || version.minor === undefined) {
+            return undefined;
+        }
+        return `${version.major}.${version.minor}`;
+    }
+
+    /**
+     * Classify the *active interpreter's* provenance from the resolved
+     * environment alone. This is deliberately independent of project files: a
+     * project carrying `uv.lock` but running a conda/venv/system interpreter
+     * must report that interpreter's real source, so the setup-flow gap ("uv
+     * project, interpreter not uv-managed yet") stays visible. `uv.lock` is
+     * still captured as a strong *project* signal via `hasUvLock`.
+     */
+    private getInterpreterSource(
+        env: ResolvedEnvironment | undefined
+    ): InterpreterSource {
+        if (env?.environment === undefined) {
+            // No managed environment: a global/system interpreter.
+            return env ? "system" : "unknown";
+        }
+
+        const tools = env.tools ?? [];
+        if (env.environment.type === "Conda" || tools.includes("Conda")) {
+            return "conda";
+        }
+        if (
+            tools.includes("Venv") ||
+            tools.includes("VirtualEnv") ||
+            tools.includes("Poetry") ||
+            tools.includes("Pipenv") ||
+            env.environment.type === "VirtualEnvironment"
+        ) {
+            // The MS Python extension reports uv-created venvs as plain virtual
+            // environments. Distinguish a genuinely uv-provisioned interpreter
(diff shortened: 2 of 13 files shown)


rugpanov:
Integration tests: https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/27942293534

packages/databricks-vscode/src/telemetry/constants.ts
```diff
@@ -43,6 +44,17 @@ export type WorkflowTaskType = "python" | "notebook" | "unknown";
 export type LaunchType = "run" | "debug";
 export type ComputeType = "cluster" | "serverless";
 
+/** A Python package/environment manager detected for a project. */
+export type PackageManagerName = "uv" | "poetry" | "pip" | "conda";
+/** Best-guess primary manager, or "unknown" when no signal fires. */
+export type PrimaryManagerName = PackageManagerName | "unknown";
+/** How the active interpreter was provisioned. */
+export type InterpreterSource = "uv" | "conda" | "system" | "venv" | "unknown";
```
misha-db: It looks to be duplicated from packageManagerDetection.ts, can we re-use same types?

packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
```diff
@@ -0,0 +1,251 @@
...
+ * Collects package-manager signals at project-setup touchpoints and emits the
+ * {@link Events.PYTHON_ENV_SETUP_DETECTED} telemetry event.
+ *
+ * All probing is best-effort and non-blocking: any failure degrades to
+ * `unknown` and is swallowed, never thrown into the user's setup/run flow.
+ * Only categorical/enum data is emitted — no paths, package names, or other
+ * free-form content (see {@link detectPackageManagers}). Telemetry opt-out is
+ * honoured by the underlying {@link Telemetry} client.
+ */
+export class PackageManagerTelemetry {
```
misha-db, line 32: No tests for PackageManagerTelemetry?

packages/databricks-vscode/src/language/packageManagerDetection.ts
```diff
@@ -0,0 +1,287 @@
...
+    if (!sysPrefix || !condaPrefix) {
+        return false;
+    }
+    const stripTrailingSep = (p: string) => p.replace(/[\\/]+$/, "");
+    const prefix = stripTrailingSep(sysPrefix);
+    const base = stripTrailingSep(condaPrefix);
+    return (
+        prefix === base ||
+        prefix.startsWith(base + "/") ||
+        prefix.startsWith(base + "\\")
```
misha-db, line 338: This assumes filesystem is case sensitive, on Windows case `C:\Some` and `C:\some` returning false, but they denote same folder

packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
```diff
@@ -0,0 +1,251 @@
...
+            return false;
+        }
+    }
+
+    /** True if any `requirements*.txt` file exists in the project root. */
+    private hasRequirementsTxt(projectRoot: string): boolean {
+        try {
+            return fs
+                .readdirSync(projectRoot)
+                .some((name) => /^requirements.*\.txt$/.test(name));
```
misha-db: This regex matches "requirementswhatever.txt" as well, was it intentional?

misha-db (changes requested):

packages/databricks-vscode/src/telemetry/packageManagerExtensions.ts
```diff
@@ -0,0 +1,51 @@
...
+        ): void;
+    }
+}
+
+Telemetry.prototype.recordPackageManagerDetection = function (
+    detection: PackageManagerDetection,
+    context: PackageManagerDetectionContext
+) {
+    this.recordEvent(Events.PYTHON_ENV_SETUP_DETECTED, {
+        managersDetected: detection.managers as PackageManagerName[],
```
anton-107: **Unsafe casts hide schema divergence.** `detection.managers as PackageManagerName[]` (and `interpreterSource as InterpreterSource` on L46) bridge two parallel union definitions: `PackageManager`/`InterpreterSource`/`PrimaryManager` in `packageManagerDetection.ts` vs `PackageManagerName`/`InterpreterSource`/`PrimaryManagerName` in `constants.ts`. These `as` casts defeat the one check that matters here — if someone later adds a manager to one union but not the other, it compiles silently and ships a mismatched event schema. Suggest a single source of truth: import the detection types into `constants.ts` (or re-export from one place) and drop the casts so the compiler enforces alignment.

packages/databricks-vscode/src/language/EnvironmentDependenciesVerifier.ts
```diff
@@ -403,6 +405,9 @@ export class EnvironmentDependenciesVerifier extends MultiStepAccessVerifier {
     }
 
     override async check() {
+        // First environment check on project open: emit package-manager
+        // detection telemetry (deduplicated per session, never throws).
+        void this.packageManagerTelemetry.emitDetection("auto_open");
```
anton-107: **`targetCompute` is nondeterministic for `auto_open`.** This fires before `await this.connectionManager.waitForConnect()` on the next line. `emitDetection` awaits `resolveEnvironment()` and then reads `connectionManager.cluster`/`.serverless` via `getComputeType()`, so whether the connection has resolved by then is a race against the parallel `waitForConnect()`. Result: `targetCompute` for `auto_open` will usually be `none` but sometimes the real value — dirty data on the trigger that fires most. Either move the emit after `waitForConnect()` to make it deterministic, or accept+document that `auto_open` always reports `none` for compute.

packages/databricks-vscode/src/language/packageManagerDetection.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    /** One or more `requirements*.txt` files exist. */
+    hasRequirementsTxt?: boolean;
+    /** A `constraints.txt` file exists. */
+    hasConstraintsTxt?: boolean;
+    /**
+     * A `pyproject.toml` exists but declares neither `[tool.uv]` nor
+     * `[tool.poetry]` (i.e. a plain PEP 621 / pip-installable project).
+     */
+    hasPyprojectPipOnly?: boolean;
```
anton-107, line 95: **`hasPyprojectPipOnly` can misattribute uv projects as pip.** uv works with a bare `[project]` table and no `[tool.uv]` section, so a uv project without a committed `uv.lock` classifies as `pip`. In the common case (`uv.lock` present) uv still fires and wins primary, so impact is limited to lockfile-less uv projects — but since the whole point is sizing uv adoption accurately, worth a one-line caveat here (or in the handoff doc) so the analytics owner knows `pyproject.pipOnly` slightly over-counts pip / under-counts uv.

packages/databricks-vscode/src/language/packageManagerDetection.ts
```diff
@@ -0,0 +1,287 @@
...
+ * Whether an interpreter's `sysPrefix` lies inside a conda prefix -- i.e. the
+ * active interpreter is that conda environment, not merely a shell that has
+ * `CONDA_PREFIX` exported globally. Both inputs are expected to be absolute
+ * paths; comparison is done with a trailing-separator boundary so that
+ * `/x/envs/ab` is not treated as inside `/x/envs/a`.
+ *
+ * Pure over its inputs; returns false if either is missing. Accepts both `/`
+ * and `\\` separators so it is platform-agnostic and deterministic in tests.
+ */
+export function interpreterUnderCondaPrefix(
```
anton-107, line 321: **Case-sensitivity vs the "platform-agnostic" claim.** The boundary check handles `/` vs `\` separators but compares case-sensitively, so on Windows `C:\Conda\envs\ml` vs `c:\conda\...` would not match even though Windows paths are case-insensitive. Low impact (both values usually come from the same source), but the doc comment says "platform-agnostic" which overstates it. Either lowercase both sides when on Windows, or soften the comment.

packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
```diff
@@ -0,0 +1,251 @@
...
+        private readonly getProjectRoot: () => string | undefined,
+        private readonly getComputeType: () => ComputeType | "none"
+    ) {}
+
+    /**
+     * Detect the package manager(s) for the active project and emit telemetry.
+     * Deduplicated per `(project root, trigger)` within the session. Never
+     * throws.
+     */
+    async emitDetection(trigger: SetupTrigger): Promise<void> {
```
anton-107, line 56: **Suggestion: short-circuit when telemetry isn't at `all`.** The event payload is correctly suppressed by `@vscode/extension-telemetry` when the user's `telemetry.telemetryLevel` isn't `all` (regular events only send at `all`), so nothing leaks. But `recordEvent` only gates on `!this.reporter`, and `reporter` is constructed regardless of the user's setting — so for an opted-out user, `collectSignals` still runs `fs.readdirSync(projectRoot)` + `fs.readFileSync` on `pyproject.toml`/`pyvenv.cfg` on every `auto_open`/`run`/`setup`; only the *send* is dropped. That's wasted work, and the optics are poor for a privacy-conscious user auditing the extension (a telemetry code path reading project files despite telemetry being off). Recommend bailing out of `emitDetection` early when telemetry is disabled (e.g. check `vscode.env.isTelemetryEnabled` / the reporter level before collecting) so opted-out users get zero telemetry-driven disk access. Separately, the handoff doc says "no paths/names" but doesn't mention that every event — including this one — inherits the ambient `user.hashedUserName` / `user.host` / `workspaceId` envelope; worth stating so reviewers know this links toolchain choices to a stable identity.

anton-107 (commented):
A few follow-up comments from a closer pass (correctness/maintainability + one telemetry-gating suggestion). Nice structure overall — the pure/emit/collect split and the privacy posture of the payload are good. None of these are blockers; the casts (1) and the opt-out gating (5) are the two I'd most want addressed.

packages/databricks-vscode/src/telemetry/constants.ts
rugpanov: Done in ab23a28 — `constants.ts` now imports `PackageManager`/`PrimaryManager`/`InterpreterSource` (type-only) from the pure `packageManagerDetection` module as the single source of truth; the duplicate unions and the `as` casts are gone, so the compiler enforces schema/classifier alignment.

packages/databricks-vscode/src/telemetry/packageManagerExtensions.ts
rugpanov: Done in ab23a28 — unified the unions (single source of truth in `packageManagerDetection.ts`, imported type-only into `constants.ts`) and removed both `as` casts here, so a divergence now fails to compile.

packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
rugpanov: Added in ab23a28 — `PackageManagerTelemetry.test.ts` covering the emit payload, per-`(trigger, project)` dedupe, the disconnected guard (incl. not consuming the dedupe slot), the telemetry-disabled short-circuit, and the requirements-suffix cases.

packages/databricks-vscode/src/language/packageManagerDetection.ts
rugpanov: Fixed in ab23a28 — `interpreterUnderCondaPrefix` now folds case via a `caseInsensitive` param defaulting to `process.platform === "win32"`, so Windows paths differing only in case match. Added tests.

packages/databricks-vscode/src/language/packageManagerDetection.ts
rugpanov: Fixed in ab23a28 — case-insensitive comparison on Windows, and corrected the "platform-agnostic" comment to reflect it. Added tests for both case modes.

packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
rugpanov: Fixed in ab23a28 — tightened to `/^requirements([-_.].+)?\.txt$/`, so `requirements.txt` / `requirements-dev.txt` / `requirements_test.txt` match but `requirementsfoo.txt` does not. Covered by tests.

packages/databricks-vscode/src/language/EnvironmentDependenciesVerifier.ts
rugpanov: Fixed in 34693f5 — moved the emit to after `await waitForConnect()` (which resolves only on `CONNECTED`) and added an `isConnected()` guard. `targetCompute` is now deterministic for `auto_open`, and unauthenticated sessions are no longer reported at all.

packages/databricks-vscode/src/language/packageManagerDetection.ts
rugpanov: Documented in ab23a28 — caveat added on the `hasPyprojectPipOnly` field and a "Known measurement caveats" section in the handoff doc, noting pip is slightly over-counted / uv under-counted for lockfile-less uv projects.

packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
rugpanov: Done in ab23a28 — `emitDetection` now bails before any disk read when telemetry is disabled (new `Telemetry.isTelemetryEnabled`, true only at level `all`), so opted-out users get zero telemetry-driven file access. Also documented the ambient `user.hashedUserName`/`host`/`workspaceId` identity envelope in the handoff doc.

packages/databricks-vscode/src/language/packageManagerDetection.ts
rugpanov: Follow-up: in b05df5d this is now fixed in code, not just documented. `pyproject.toml` is attributed to pip only when it declares an actual packaging table (`[project]`/`[build-system]`) and no uv/poetry section — a file carrying only tool config like `[tool.ruff]` is no longer counted as pip. (Found independently by a Codex review pass too.) The lockfile-less-uv caveat in the handoff doc still stands.

rugpanov:
### Update: review feedback + Codex review addressed

Pushed three follow-up commits on top of the original (no force-push, so the inline threads above stay anchored):

- **`34693f5`** — Gate detection on workspace connection. Emit only while `CONNECTED` (so unauthenticated sessions / non-Databricks folders are never reported); `auto_open` moved after `waitForConnect()`, which also makes `targetCompute` deterministic.
- **`ab23a28`** — Address review feedback: single source of truth for the manager/interpreter unions (dropped the unsafe casts), `PackageManagerTelemetry` tests, Windows case-insensitive conda-path compare, tightened requirements regex, short-circuit before any disk read when telemetry is disabled, and doc caveats (`pyproject.pipOnly`, identity envelope).
- **`b05df5d`** — Address a Codex review pass: pip is attributed only when `pyproject.toml` has a real packaging table (`[project]`/`[build-system]`), Poetry-managed interpreters now attribute to `poetry` (new `poetry` interpreter source) instead of pip, and `pythonVersion` is omitted when unresolved (no `"undefined"` string).

I replied inline on each addressed thread with the specific commit. Schema note for the analytics owner: `interpreterSource` gained a `poetry` value and `signals` gained `interpreter.poetry` (both additive) — see `src/telemetry/PACKAGE_MANAGER_DETECTION.md`.

**Verification:** `yarn run build` clean, eslint 0 warnings, prettier clean, `yarn run test:unit` 246 passing / 0 failing, and `codex review --base main` reports no actionable findings.

anton-107 (approved):
Verified all 8 addressed review comments against the source (not just the reply text), and built + ran the unit suite locally.

**Comments — all correctly resolved:**
- Type duplication / unsafe casts (misha-db, anton-107): single source of truth now in `packageManagerDetection.ts`, imported type-only into `constants.ts`; both `as` casts gone, so schema/classifier divergence fails to compile.
- Case-sensitive path compare (misha-db, anton-107): `interpreterUnderCondaPrefix` folds case via `caseInsensitive` defaulting to `win32`; doc comment corrected.
- `requirements*.txt` over-broad regex (misha-db): tightened to `/^requirements([-_.].+)?\.txt$/`; `requirementsfoo.txt` excluded.
- `hasPyprojectPipOnly` misattributing uv (anton-107): went beyond the doc caveat — now gated on `pyprojectHasPackagingTable()` so a tool-only `pyproject.toml` isn't counted as pip.
- Telemetry-off short-circuit (anton-107): `emitDetection` bails before any disk read when telemetry isn't `all`.
- `auto_open` compute race (anton-107): emit moved after `await waitForConnect()` + `isConnected()` guard; `targetCompute` deterministic, unauthenticated sessions not reported.
- Collector tests (misha-db): `PackageManagerTelemetry.test.ts` added, covering emit payload, dedupe, disconnected-guard (incl. not consuming the dedupe slot), telemetry-disabled, and requirements-suffix cases.

**Local verification:** `yarn run build` passes; `yarn run test:unit` = 260 passing. The 6 failing tests are all in `CliWrapper.test.js` (missing bundled CLI binary in a clean checkout) — the PR touches no CLI files, so they're pre-existing/environmental, not regressions.

LGTM. Only non-blocking note: the lockfile-less-uv skew remains a documented known limitation by design.

misha-db (approved):

PR #1922, opened 2026-06-22
Add create file support to WSFS explorer

misha-db:
## Changes

Adds the ability to create new files directly in the Databricks Workspace Filesystem (WSFS) explorer, alongside the existing folder-creation and upload commands. Users can now create a file from the view toolbar or from a directory/repo context menu, name it via the existing wizard, and have it open automatically in the editor.

## Tests

Updated `WorkspaceFsCommands.test.ts`

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -197,6 +197,20 @@
                 "category": "Databricks"
             },
             {
+                "command": "databricks.wsfs.createNewFile",
+                "title": "Create File",
+                "icon": "$(new-file)",
+                "enablement": "databricks.context.activated && databricks.context.loggedIn && !databricks.context.remoteMode",
+                "category": "Databricks"
+            },
+            {
+                "command": "databricks.wsfs.createNewFile.toolbar",
+                "title": "Create File",
+                "icon": "$(new-file)",
+                "enablement": "databricks.context.activated && databricks.context.loggedIn && !databricks.context.remoteMode",
+                "category": "Databricks"
+            },
+            {
                 "command": "databricks.wsfs.openInBrowser",
                 "title": "Open in Browser",
                 "icon": "$(link-external)",
@@ -727,6 +741,11 @@
                     "group": "navigation@1"
                 },
                 {
+                    "command": "databricks.wsfs.createNewFile.toolbar",
+                    "when": "view == workspaceFsView",
+                    "group": "navigation@1"
+                },
+                {
                     "command": "databricks.unityCatalog.filter",
                     "when": "view == unityCatalogView",
                     "group": "navigation@2"
@@ -803,9 +822,14 @@
                     "group": "wsfs_mut@0"
                 },
                 {
-                    "command": "databricks.wsfs.uploadFile",
+                    "command": "databricks.wsfs.createNewFile",
                     "when": "view == workspaceFsView && (viewItem == wsfs.directory || viewItem == wsfs.repo)",
                     "group": "wsfs_mut@1"
+                },
+                {
+                    "command": "databricks.wsfs.uploadFile",
+                    "when": "view == workspaceFsView && (viewItem == wsfs.directory || viewItem == wsfs.repo)",
+                    "group": "wsfs_mut@2"
                 },
                 {
                     "command": "databricks.wsfs.downloadFile",
--- a/packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
+++ b/packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
@@ -210,13 +210,18 @@ export class WorkspaceFsCommands implements Disposable {
             return;
         }
 
+        const isIpynb = inputName.toLowerCase().endsWith(".ipynb");
+        const existingName = isIpynb
+            ? inputName.replace(/\.ipynb$/i, "")
+            : inputName;
+
         const existing = await WorkspaceFsEntity.fromPath(
             client,
-            `${root.path}/${inputName}`
+            `${root.path}/${existingName}`
         );
         if (existing) {
             const answer = await window.showWarningMessage(
-                `"${inputName}" already exists in the workspace. Overwrite it?`,
+                `"${existingName}" already exists in the workspace. Overwrite it?`,
                 {modal: true},
                 "Overwrite"
             );
@@ -225,12 +230,11 @@ export class WorkspaceFsCommands implements Disposable {
             }
         }
 
-        const content = inputName.toLowerCase().endsWith(".ipynb")
-            ? EMPTY_IPYNB_CONTENT
-            : "";
+        const content = isIpynb ? EMPTY_IPYNB_CONTENT : "";
 
+        let created: WorkspaceFsEntity | undefined;
         try {
-            await root.createFile(inputName, content, true);
+            created = await root.createFile(inputName, content, true);
         } catch (e: unknown) {
             const msg = e instanceof Error ? e.message : String(e);
             window.showErrorMessage(`Failed to create "${inputName}": ${msg}`);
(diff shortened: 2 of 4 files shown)


packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
```diff
@@ -141,6 +164,111 @@ export class WorkspaceFsCommands implements Disposable {
...
+        }
+
+        this.workspaceFsDataProvider.refresh();
+        const uri = Uri.from({
+            scheme: WorkspaceFsFileSystemProvider.scheme,
+            path: created?.path ?? `${root.path}/${inputName}`,
+        });
+        this.fsp.notifyCreated(uri);
+
+        if (isIpynb && created) {
```
anton-107: This open-mode decision keys off `isIpynb`, but `WorkspaceFsDir.createFile` strips `(py|ipynb|scala|r|sql)` on lookup — so a `.py`/`.sql`/`.scala`/`.r` file is also stored as a notebook (extension dropped), and `created.path` here is the stripped path. Those fall into the editor branch below and get opened as a text document at the stripped notebook path, inconsistent with `.ipynb`.

Suggest deciding from the actual entity instead of the extension — `WorkspaceFsUtils` is already imported and `isNotebook` exists:
```ts
if (created && WorkspaceFsUtils.isNotebook(created)) {
    try { await this.openInBrowser(created); }
    catch (e: unknown) { ctx?.logger?.error(`Can't open ${inputName} in browser after creation`, e); }
    return;
}
```
This removes the brittle string check entirely and makes `.py`-as-notebook behave like `.ipynb`.

packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
```diff
@@ -141,6 +164,111 @@ export class WorkspaceFsCommands implements Disposable {
...
+            "File Name",
+            root
+        );
+
+        if (inputName === undefined) {
+            return;
+        }
+
+        const isIpynb = inputName.toLowerCase().endsWith(".ipynb");
+        const existingName = isIpynb
```
anton-107, line 218: The existence check only strips `.ipynb`, but the SDK stores `(py|ipynb|scala|r|sql)` at the extension-stripped path. So creating `script.py` when a `script` notebook already exists won't detect the clash, won't prompt, and — because `createFile(..., true)` overwrites — silently destroys the existing notebook. Consider stripping the same extension set the SDK does (`/\.(py|ipynb|scala|r|sql)$/i`) for this lookup.

packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
```diff
@@ -141,6 +164,111 @@ export class WorkspaceFsCommands implements Disposable {
...
+            window.showErrorMessage("Please login first to create a file");
+            return;
+        }
+
+        const root = await this.getValidRoot(rootPath, ctx);
+        if (!root) {
+            return;
+        }
+
+        const inputName = await createDirWizard(
```
anton-107, line 207: Non-blocking: `createDirWizard` is folder-oriented when reused for filenames — it pre-fills the input `value`/`placeHolder` with the project folder basename (an odd default for a new file) and its validation returns `"Invalid name: Folders cannot contain '/'"`. Consider a file-appropriate default and neutral wording.

packages/databricks-vscode/package.json
```diff
@@ -726,6 +740,11 @@
                     "when": "view == workspaceFsView",
                     "group": "navigation@1"
                 },
+                {
+                    "command": "databricks.wsfs.createNewFile.toolbar",
```
anton-107, line 744: Nit (optional): this toolbar button shares `navigation@1` with `createFolder.toolbar` and `uploadFile.toolbar`, so ordering between the three is implicit. The context-menu entries use explicit `wsfs_mut@0/@1/@2`; consider the same explicit ordering here for predictability. Pre-existing pattern, so optional.

anton-107 (changes requested):
Solid addition — the command wiring, menu placement, and toolbar-vs-context-menu target resolution cleanly mirror the existing `createFolder`/`uploadFile` patterns, and the test coverage is thorough. One correctness/consistency issue should be addressed before merge, plus a couple of minor UX nits (left inline).

**Main issue — the `.ipynb`-only heuristic is too narrow.** `doCreateFile` decides everything from `inputName.endsWith(".ipynb")`, but the SDK imports with `format: "AUTO"` and `WorkspaceFsDir.createFile` strips a broader set on lookup: `(py|ipynb|scala|r|sql)`. That regex is the SDK telling us `.py`/`.scala`/`.r`/`.sql` are also stored as notebooks (extension dropped). Consequences for e.g. `script.py`:

1. It lands in the editor branch and opens `created.path` (the stripped notebook path) as a text document — inconsistent with `.ipynb`, which opens in the browser.
2. The existence pre-check looks at `${root}/script.py` while the notebook lands at `${root}/script`, so the clash is missed and — since `createFile(..., true)` passes `overwrite=true` — an existing notebook is silently overwritten with no prompt. (`doUploadFile` shares this gap, so it's pre-existing, but worth not propagating.)

Recommend deciding the open mode from the created entity via the already-imported `WorkspaceFsUtils.isNotebook(created)` instead of the extension string, and widening the existence-check stripping to the SDK's set. See inline comments.

**Heads-up on tests:** the test fake's `createFile` only strips `.ipynb`, so the suite is green but wouldn't catch the above. If you switch to `isNotebook(created)`, please add a test where the fake returns a `NOTEBOOK`-typed entity for a `.py` name.

packages/databricks-vscode/package.json
rugpanov, line 744: [regression] That toolbar button is available regardless of selected item. VS Code can pass the selected tree item to the toolbar command, and createFileFromToolbar currently uses activeElement?.path without checking type. So if a file/notebook is selected, it tries to create under that file path and getValidRoot rejects it.

rugpanov (changes requested):

packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
misha-db: using WorkspaceFsUtils.isNotebook(created) to detected if created file was notebook

packages/databricks-vscode/src/workspace-fs/WorkspaceFsCommands.ts
misha-db: When creating new file, content is empty string "" for .sql, .py, .scala, .r etc.. so it ends up with type FILE, not converted to notebook type automatically. so extension will not be stripped, and no silent overwrite will happen.

in case of ipynb file, valid empty notebook content (json) is provided, so it will be converted to notebook format, that's the reason why we check it explicitly

packages/databricks-vscode/package.json
misha-db: Yes, and user gets descriptive error message in case they select wrong parent to add file under. like: "Cannot create a directory as a child of a NOTEBOOK/FILE"

misha-db: Updated error message to make it more specific

rugpanov: Sounds good

rugpanov:
Looks good from my side, I let @anton-107 to make a pass before approving.

anton-107 (approved):
Follow-ups all addressed — approving.

- **Open-mode decision:** now branches on `WorkspaceFsUtils.isNotebook(created)` instead of the extension string, so the open path is driven by the actual created entity. My main concern is resolved.
- **Existence-check stripping:** confirmed correct for this flow — new files are created with empty content, so `.py`/`.sql`/`.scala`/`.r` import as plain `FILE`s with the extension preserved (only `.ipynb` converts to a notebook at the stripped path). The created-path and lookup keys stay consistent, so the silent-overwrite case can't occur here. The narrow `.ipynb` stripping is the right choice.
- **Toolbar regression (selecting a file/notebook):** `getValidRoot` rejects a non-directory target with a clear message, so it's a graceful error rather than a crash.

Remaining items are cosmetic (folder-oriented `createDirWizard` wording for filenames; implicit `navigation@1` toolbar ordering) and non-blocking.

rugpanov (approved):

PR #1925, opened 2026-06-23
Fix Volume "Open in Databricks" routing to volume URL

### LINKED ISSUE #1923: [BUG] Volume Open in Databricks Fails - Wrong URI
**Describe the bug**
With the Unity Catalog explorer built into the extension now, viewing a Volume offers the same Open in Databricks option as tables, but when you try to open the Volume it takes you to a "table" referenced by the name of your volume and fails.

**To Reproduce**

1. Go to 'Unity Catalog'
2. Click on a catalog and schema where the volume exists
3. Click on a volume
4. Click on 'Open in Databricks' button
5. See object that doesn't exist (unless you have a table by the same name; and that's possible?)

**Additional context**
Specifically, it's trying to route to a table like 

`https://<workspace>.azuredatabricks.net/explore/data/<catalog>/<schema>/<asset>`

When the volume object actually is

`https://<workspace>.azuredatabricks.net/explore/data/volumes/<catalog>/<schema>/<asset>`

misha-db:
## Changes

When opening a Volume from the Unity Catalog explorer, the "Open in Databricks" action routed to a table-style URL
(/explore/data/<catalog>/<schema>/<name>), which points at a non-existent table rather than the volume.

getNodeExploreUrl() special-cased models and functions by prefixing their path segment but had no case for volumes, so volumes fell through to the default (table) path. Add a "volume" case that prefixes the path with "volumes/", producing the correct
/explore/data/volumes/<catalog>/<schema>/<name> URL.

## Tests

UnityCatalogTreeDataProvider.test.ts

### DIFF
--- a/packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.test.ts
+++ b/packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.test.ts
@@ -352,6 +352,36 @@ describe(__filename, () => {
         assert.strictEqual(item.copyText, "cat");
     });
 
+    it("getTreeItem sets volume url with volumes path segment", async () => {
+        const stubManager = {
+            onDidChangeState: () => ({dispose() {}}),
+            databricksWorkspace: {
+                host: new URL("https://adb-123.azuredatabricks.net/"),
+            },
+        } as unknown as ConnectionManager;
+
+        const provider = new UnityCatalogTreeDataProvider(
+            stubManager,
+            stubStateStorage
+        );
+        disposables.push(provider);
+
+        const volume: UnityCatalogTreeNode = {
+            kind: "volume",
+            catalogName: "cat",
+            schemaName: "sch",
+            name: "ev",
+            fullName: "cat.sch.ev",
+        };
+        const item = provider.getTreeItem(volume) as UnityCatalogTreeItem;
+
+        assert(item.url, "url should be set");
+        assert(
+            item.url!.includes("explore/data/volumes/cat/sch/ev"),
+            `url should contain explore/data/volumes/cat/sch/ev, got: ${item.url}`
+        );
+    });
+
     it("getTreeItem omits url when no host", async () => {
         const provider = new UnityCatalogTreeDataProvider(
             instance(mockConnectionManager),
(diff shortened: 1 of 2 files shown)


packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.test.ts
```diff
@@ -352,6 +352,36 @@ describe(__filename, () => {
...
+            catalogName: "cat",
+            schemaName: "sch",
+            name: "ev",
+            fullName: "cat.sch.ev",
+        };
+        const item = provider.getTreeItem(volume) as UnityCatalogTreeItem;
+
+        assert(item.url, "url should be set");
+        assert(
+            item.url!.includes("explore/data/volumes/cat/sch/ev"),
```
anton-107, line 380: Nit (non-blocking): consider asserting the exact URL with `assert.strictEqual(item.url, "https://adb-123.azuredatabricks.net/explore/data/volumes/cat/sch/ev")` instead of `includes(...)`. A substring check won't catch accidental double-slashes or a wrong prefix/host being concatenated; an exact match locks the route down.

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.test.ts
```diff
@@ -352,6 +352,36 @@ describe(__filename, () => {
         assert.strictEqual(item.copyText, "cat");
     });
 
+    it("getTreeItem sets volume url with volumes path segment", async () => {
+        const stubManager = {
```
anton-107, line 356: Nit (non-blocking): this hand-rolls a `stubManager` with an inline host, while the sibling tests use the shared `mockConnectionManager`. The inline stub is justified here because this test needs a host present, but if the surrounding tests have a helper for a host-configured manager it'd be worth reusing for consistency.

anton-107 (approved):
Correct, minimal fix. `getNodeExploreUrl()` was missing a `volume` case, so volumes fell through to the bare table path; prefixing `volumes/` matches the canonical Databricks Explore route and is consistent with the existing `models/` and `functions/` cases. Volume nodes carry `fullName`, so the path conversion is safe, and the new test guards the regression.

Approving. Left two non-blocking nits on the test for you to consider.

PR #1960, opened 2026-07-03
Bump bcryptjs from 2.4.3 to 3.0.3

rugpanov:
## Why

Replaces Dependabot's #1937, which bumped `package.json` without regenerating the lockfile (Yarn Berry `YN0028` failure). bcryptjs is used at **runtime** in telemetry (`src/telemetry/index.ts`) to hash the username into `user.hashedUserName`, so the critical requirement is that v3 produces the **same hash** as v2 for a given `(input, salt)` — otherwise telemetry identity would silently change.

## What

- Bump `bcryptjs` → `^3.0.3` via `yarn up`; regenerated `yarn.lock`.
- Remove the now-redundant `@types/bcryptjs` devDependency — bcryptjs 3 ships its own type definitions (`umd/index.d.ts`), so the separate `@types` package is obsolete. Build type-checks cleanly on the bundled types.

## Verification

- **Hash stability confirmed**: `bcrypt.hash("testuser@example.com", "$2b$07$" + salt)` produces the **byte-identical** output under v2 and v3.
- **Telemetry unit test passes** — `src/telemetry/index.test.ts` asserts an exact hardcoded `user.hashedUserName` bcrypt hash; full unit suite green (**274 passing**).
- `require("bcryptjs")` resolves fine — v3 is ESM-only but ships a UMD entry, so the esbuild CJS bundle is unaffected.
- `yarn install --immutable`, `yarn run build`, and `eslint` all pass.

**Backward compatibility:** hash output is identical to v2, so existing `user.hashedUserName` telemetry values remain consistent. No API/state/config change.

Closes #1937.

This pull request and its description were written by Isaac.

PR #1966, opened 2026-07-06
Upgrade bcryptjs 2.4.3 → 3.0.2 and drop @types/bcryptjs

rugpanov:
## Why

Dependabot raised `@types/bcryptjs` to 3.0.0 (#1963), but that release is a **deprecated stub** — `bcryptjs` ships its own type definitions from v3, so the stub is only correct once the runtime package is on 3.x. Merging the stub while staying on `bcryptjs ^2.4.3` (which bundles no types) would strip types from `src/telemetry/index.ts`. This PR upgrades the runtime package and drops the separate `@types` entry, **superseding #1963**.

## What

- `dependencies`: `bcryptjs ^2.4.3` → `^3.0.2`
- `devDependencies`: remove `@types/bcryptjs` (bcryptjs 3 bundles `umd/index.d.ts`)
- Regenerated root `yarn.lock` accordingly

No source change: telemetry hashing already passes an explicit `$2b$07$` salt, so bcryptjs 3's *"generate 2b hashes by default"* change is a no-op here.

## Verification

- `yarn install --immutable` passes (no `YN0028`); bcryptjs resolves to 3.0.3, `@types/bcryptjs` no longer present.
- `tsc --build` passes — bcryptjs 3's bundled types resolve for `import bcrypt from "bcryptjs"` with `@types` removed.
- `esbuild --format=cjs --platform=node` bundles cleanly; bcrypt is present in `out/extension.js` (the ESM-default entry resolves via the UMD fallback, so the CJS bundle is unaffected).
- eslint clean.
- **Telemetry hash is byte-identical across the bump** for the same username + salt, so `user.hashedUserName` telemetry stays continuous — no analytics discontinuity for existing users.

## Backward compatibility

Telemetry hash output verified unchanged; dependency + lockfile only, no API or build-output change.

Closes #1963.

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -1550,7 +1550,7 @@
         "@vscode/webview-ui-toolkit": "^1.4.0",
         "add": "^2.0.6",
         "ansi-to-html": "^0.7.2",
-        "bcryptjs": "^2.4.3",
+        "bcryptjs": "^3.0.2",
         "highlight.js": "^11.10.0",
         "lodash": "^4.17.21",
         "markdown-it": "^14.2.0",
@@ -1563,7 +1563,6 @@
     "devDependencies": {
         "@istanbuljs/nyc-config-typescript": "^1.0.2",
         "@sinonjs/fake-timers": "^11.2.2",
-        "@types/bcryptjs": "^2.4.6",
         "@types/chai": "^4.3.11",
         "@types/eslint": "^8.44.9",
         "@types/fs-extra": "^11.0.4",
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests triggered for `b131101d` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/28778658591)

rugpanov:
Superseded by #1960, which already upgraded bcryptjs to 3.0.3 and dropped @types/bcryptjs on main. This PR only pins an older 3.0.2 and adds nothing over main. Closing.

PR #1967, opened 2026-07-06
Upgrade @vscode/extension-telemetry 0.9.9 → 1.5.2

rugpanov:
## Why

Dependabot #1965 bumped `@vscode/extension-telemetry` to 1.5.2, but the one-line `package.json` change **breaks the build**. The major release changed the module's export shape:

| | 0.9.9 | 1.5.2 |
|---|---|---|
| Export | `export default class TelemetryReporter` | `export class TelemetryReporter` (named, no default) |

The extension imports it as a **default import**, so under `esModuleInterop` + CommonJS, `new TelemetryReporter(...)` fails to compile:

```
error TS2351: This expression is not constructable.
  Type 'typeof import(".../telemetryReporter")' has no construct signatures.
```

That is why the integration tests were red on #1965 (3 of 4 jobs failed — the extension couldn't build). Dependabot also only touched `package.json`, leaving `yarn.lock` pinned to 0.9.1.

## What

- Bump `@vscode/extension-telemetry` to `^1.5.2` in `packages/databricks-vscode/package.json` and regenerate `yarn.lock` (pulls `applicationinsights` 3.4.x transitively).
- Switch the default import to a **named import** in `src/telemetry/index.ts` and `src/telemetry/index.test.ts`.

The only API surface the extension uses is backward compatible, so there is **no behavior change**:
- `sendTelemetryEvent(...)` — 1.5.2 only *adds* an optional 4th `tagOverrides` arg.
- `.telemetryLevel` — unchanged (`'all' | 'error' | 'crash' | 'off'`).
- Emitted event/data format — unchanged.
- Engine — 1.5.2 requires vscode `^1.75.0`; the extension is at `^1.86.0`.

## Verification

- `yarn run build` (`tsc --build`) succeeds; compiled output emits `new extension_telemetry_1.TelemetryReporter(...)`.
- `yarn run test:unit`: **268 passing, 0 failing**, including all 4 telemetry tests that use `mock(TelemetryReporter)`.
- `eslint` + `prettier` clean on the changed files.

Supersedes #1965 (which cannot merge without this code change).

---

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -1546,7 +1546,7 @@
         "@types/lodash": "^4.14.202",
         "@types/shell-quote": "^1.7.5",
         "@vscode/debugadapter": "^1.64.0",
-        "@vscode/extension-telemetry": "^0.9.1",
+        "@vscode/extension-telemetry": "^1.5.2",
         "@vscode/webview-ui-toolkit": "^1.4.0",
         "add": "^2.0.6",
         "ansi-to-html": "^0.7.2",
--- a/packages/databricks-vscode/src/telemetry/index.test.ts
+++ b/packages/databricks-vscode/src/telemetry/index.test.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import TelemetryReporter from "@vscode/extension-telemetry";
+import {TelemetryReporter} from "@vscode/extension-telemetry";
 import assert from "assert";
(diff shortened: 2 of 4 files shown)


rugpanov:
🤖 Integration tests ❌ **8 of 35 test jobs failed** for `218ee889` (27 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/28780342518)

misha-db (approved):

PR #1970, opened 2026-07-06
Upgrade nyc 15.1.0 → 18.0.0

rugpanov:
## Why

Dependabot #1962 bumped `nyc` (coverage tool) to 18.0.0 but could not land. Its `@dependabot recreate` regenerated only `package.json` and left `yarn.lock` pinned at `nyc@15.1.0`, so CI's `yarn install --immutable` fails with:

```
YN0028: The lockfile would have been modified by this install, which is explicitly forbidden.
```

Dependabot's Yarn 3 support can't regenerate this lockfile itself (this is also why the original #1962 branch carried a manual lockfile commit). This PR supersedes #1962 with a correctly regenerated lock off current `main`.

nyc 18's only breaking change is that transitive deps now require node `20 || >=22`. The extension requires `node >=22.0` and the coverage job (`test:cov` in `unit-tests.yml`) runs on **Node 22.x**, so this is satisfied.

## What

- Bump `nyc` `^15.1.0` → `^18.0.0`.
- Regenerate `yarn.lock` (transitive churn: babel 7.29, glob 13, rimraf 6, test-exclude 8, istanbul-lib-instrument 6).

Dev-only dependency — nyc is the coverage runner, not shipped in the extension bundle. The existing nyc config block (`@istanbuljs/nyc-config-typescript`, lcov+json reporters) is unchanged and fully compatible with v18. No runtime, extension-behavior, or output change for users.

## Verification

- `yarn install --immutable` passes (no `YN0028`, no `YN0060`) — the exact step that failed on the Dependabot recreate.
- `npx nyc --version` → `18.0.0`.
- `yarn run test:cov` (nyc + full unit suite): **268 passing, 0 failing**, and nyc emits valid coverage artifacts (`coverage/lcov.info`, `coverage-final.json`, `lcov-report/`, `.nyc_output/`) — confirms v18 instruments the TS source and reports correctly with the existing config.

Supersedes #1962 (which cannot merge — Dependabot leaves the lockfile out of sync).

---

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -1594,7 +1594,7 @@
         "glob": "^10.5.0",
         "mocha": "^11.7.6",
         "mock-require": "^3.0.3",
-        "nyc": "^15.1.0",
+        "nyc": "^18.0.0",
         "prettier": "^3.1.1",
         "quicktype-core": "^23.0.0",
         "tmp-promise": "^3.0.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -12,17 +12,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@ampproject/remapping@npm:^2.2.0":
-  version: 2.2.1
-  resolution: "@ampproject/remapping@npm:2.2.1"
-  dependencies:
-    "@jridgewell/gen-mapping": ^0.3.0
-    "@jridgewell/trace-mapping": ^0.3.9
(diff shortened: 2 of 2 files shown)


rugpanov:
🤖 Integration tests ❌ **7 of 35 test jobs failed** for `f19c190c` (28 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/28789409034)

misha-db (approved):

PR #1971, opened 2026-07-06
[DECO-27543] Fix flaky bundle_variables e2e test: write override file directly

rugpanov:
## Why

The `bundle_variables.e2e.ts` suite's `should override default variable` test is flaky on both the Linux and Windows CI shards, failing with:

```
Error in "Bundle Variables.should override default variable"
Error: An error occurred while copying
    at ChildProcess.<anonymous> (.../node_modules/tinyclip/dist/index.js:129:44)
```

### Root cause — clipboard under headless Xvfb

The test set the variable value through the editor:

```ts
await editor.clearText();
await editor.setText(JSON.stringify({varWithDefault: "new value"}, null, 4));
await editor.save();
```

`wdio-vscode-service`'s `TextEditor.setText()` round-trips the text through the **system clipboard**. CI runs the extension host under a headless `Xvfb :99` display, where the clipboard helper (`tinyclip`, which spawns a native clipboard binary) intermittently fails — producing the `An error occurred while copying` child-process error above. It is an environment/tooling flake, not a product bug.

## What

Write the override file (`vscode.bundlevars.json`) on disk directly with `fs.writeFile` instead of typing into the editor. The extension already watches this file and refreshes the Bundle Variables tree on change, so the assertion still exercises the same code path — just without the clipboard dependency.

This mirrors the existing rationale in `wsfs_explorer.e2e.ts`, which deliberately avoids "clipboard-based editor interaction in headless test environments" by writing through the filesystem.

- The `databricks.bundle.variable.openFile` command is still fired (it materializes the file and is worth exercising); the file's real path is read from the opened editor's document URI, then written with `fs.writeFile`.
- No product-code change; scope is a single test block.

## Verification

- `tsc --build` — exit 0 (clean).
- `eslint src/test/e2e/bundle_variables.e2e.ts` — exit 0.
- `prettier -c` — passes.

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
@@ -12,7 +12,6 @@ import {
     writeRootBundleConfig,
 } from "./utils/dabsFixtures.ts";
 import {BundleSchema} from "../../bundle/types.ts";
-import fs from "fs/promises";
 
 describe("Bundle Variables", async function () {
     let workbench: Workbench;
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests ❌ **7 of 35 test jobs failed** for `4a5b87b3` (28 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/28791423749)

rugpanov:
🤖 Integration tests ❌ **5 of 35 test jobs failed** for `ea9c6f4c` (30 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/28794006696)

misha-db (approved):

PR #1989, opened 2026-07-08
Start ssh tunnel

misha-db:
## Changes

Adds a new Start SSH Tunnel entry point in the Databricks configuration view that lets users open a remote development window over databricks ssh connect. Users can pick their compute — serverless (including serverless GPU accelerators), or a dedicated single-user cluster they own — and the CLI opens a remote IDE window connected to it.

Also renames user-facing "Cluster" terminology to "Compute" across the UI to match current Databricks conventions.

## Tests

CliWrapper.test.ts


rugpanov:
🤖 Integration tests ❌ **6 of 35 test jobs failed** for `98d98ed0` (29 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29494379471)

rugpanov:
🤖 Integration tests ❌ **8 of 35 test jobs failed** for `8cc27998` (27 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29905496622)

rugpanov:
🤖 Integration tests ❌ **7 of 35 test jobs failed** for `c281a7b3` (28 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30434668024)

rugpanov:
🤖 Integration tests ❌ **6 of 35 test jobs failed** for `fb1d74f7` (29 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30438981826)

rugpanov:
🤖 Integration tests ❌ **7 of 35 test jobs failed** for `872b8994` (28 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30453007857)

rugpanov:
🤖 Integration tests ❌ **7 of 35 test jobs failed** for `204db740` (28 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30458522578)

packages/databricks-vscode/src/ssh/SshCommands.ts
```diff
@@ -0,0 +1,386 @@
...
+        }
+        return {
+            authProvider,
+            userName,
+            clusterSource: new StandaloneClusterSource(workspaceClient),
+            ownsClusterSource: true,
+        };
+    }
+
+    private pickCompute(
```
rugpanov, line 247: **Testability: `SshCommands` isn't a thin command handler**

Per `CODE_CONVENTIONS.md` §2, an `XCommands` class should be a *thin command handler that delegates to a model/manager*, and §8 notes that Model/Manager/util logic is what stays unit-testable. Right now `SshCommands` fuses the decision logic and direct SDK I/O into the command class, so this file lands with no coverage:

- **Eligible-cluster filtering, item-building, preselect matching, and the loading/empty-state transitions** are welded inside the `pickCompute` `new Promise(...)` closure — pure logic that can't run without a VS Code host. Extract these into a small `ssh/` util (or an `SshModel`) over plain data and unit-test them, the way `getSshConnectCommand` is already tested in `CliWrapper.test.ts`.
- **`StandaloneClusterSource`** fetches clusters via `Cluster.list(workspaceClient.apiClient)` directly. §5 says reach the workspace client through the connection seam rather than constructing client access inside a feature, and this is effectively an `XLoader` (§2). Pulling it into its own unit makes it testable with a fake client — and it currently swallows errors as an unhandled rejection (see the separate comment), which a test would catch.

Keep `SshCommands` as the thin handler + `launchSshTunnel` as the terminal boundary; that UI/launch seam doesn't need mocked-`window` tests. `CliWrapper`'s arg-builders are consistent with the `XWrapper` role (§2) and fine as-is.

packages/databricks-vscode/src/ssh/SshCommands.ts
```diff
@@ -0,0 +1,386 @@
...
+                    resolve({
+                        type: "serverless",
+                        accelerator: selectedItem.accelerator,
+                    });
+                }
+            });
+
+            quickPick.onDidHide(() => {
+                disposables.forEach((d) => d.dispose());
+                // resolve(undefined);
```
rugpanov: **Blocker: compute picker hangs the command when dismissed**

`resolve(undefined)` is commented out in `onDidHide`:
```ts
quickPick.onDidHide(() => {
    disposables.forEach((d) => d.dispose());
    // resolve(undefined);
});
```
VS Code fires `onDidAccept` only on selection; Escape / click-away fires `onDidHide` only. So on the common "user dismissed" path, `pickCompute`'s promise never settles → `startTunnelCommand` stays suspended at `await this.pickCompute(...)`, its `try/finally` never runs, and on the standalone path the `StandaloneClusterSource` (EventEmitter + WorkspaceClient) leaks on every dismissal. The command's telemetry also never completes.

**Fix:** uncomment `resolve(undefined)`. `Promise` resolve is idempotent, so the accept-then-hide ordering is harmless; optionally dispose the hide listener inside `onDidAccept` to be explicit.

packages/databricks-vscode/src/ssh/SshCommands.ts
```diff
@@ -0,0 +1,386 @@
...
+
+    get roots(): Cluster[] | undefined {
+        return this._clusters;
+    }
+
+    refresh() {
+        void this.load();
+    }
+
+    private async load() {
```
rugpanov, line 92: **`StandaloneClusterSource.load()` has no error handling → unhandled rejection + silent failure**

`refresh()` does `void this.load()`, and `load()` iterates `Cluster.list(...)` with no try/catch. The connected path (`ClusterLoader._load`, which wraps its load in try/catch) is protected; this standalone path is not. On a list failure (network, permission, token expiry right after login) the rejection is unhandled, `onDidChange` never fires, and the picker spins until the 10s fallback then silently shows "No dedicated clusters found" — the user can't tell discovery actually failed.

**Fix:** wrap the loop in try/catch; on error log it (and optionally surface a message), set `this._clusters = []`, and still `fire()` `onDidChange` so the spinner stops and the state is honest.

packages/databricks-vscode/src/ssh/SshCommands.ts
```diff
@@ -0,0 +1,386 @@
...
+        const {args} = this.cli.getSshConnectCommand({compute});
+
+        const env: Record<string, string> = {
+            ...this.cli.getSshConnectEnvVars(authProvider),
+            // The remote window opens at the user's home folder. Forward the
+            // file the user is currently editing so the remote extension can
+            // auto-open it (see the remote-mode branch in extension.ts). The
+            // CLI has no folder/file flag, so we pass it out of band.
+            /* eslint-disable @typescript-eslint/naming-convention */
+            DATABRICKS_REMOTE_HOME_FOLDER: `/Users/${userName}`,
```
rugpanov: **`DATABRICKS_REMOTE_HOME_FOLDER` has no consumer, and the comment doesn't match the code**

Two issues:
- **No reader.** A repo-wide search finds nothing that reads `DATABRICKS_REMOTE_HOME_FOLDER`. The remote-mode branch in `extension.ts` keys off `DATABRICKS_REMOTE_ENV` / `DATABRICKS_VIRTUAL_ENV` and doesn't open a file, so this value is currently inert.
- **Comment vs. code.** The comment says it forwards "the file the user is currently editing so the remote extension can auto-open it," but the code sets the *home folder* (`/Users/${userName}`), not a file. `userName` is the user's email, yielding `/Users/user@example.com` — a workspace-style path that may not be valid on the remote filesystem.

**Fix:** confirm the exact env var + value `databricks ssh connect` (or the remote-side extension) actually reads. If the goal is to auto-open the current file, forward `window.activeTextEditor`'s path and fix the comment. If nothing consumes it yet, drop it until the consumer lands.

packages/databricks-vscode/src/ssh/SshCommands.ts
```diff
@@ -0,0 +1,386 @@
...
+                quickPick.busy = false;
+                quickPick.placeholder = undefined;
+            };
+
+            quickPick.items = buildStaticItems(0);
+
+            const refreshItems = () => {
+                // Only dedicated single-user clusters owned by the current user
+                // can be used for an SSH tunnel.
+                const clusters = (clusterSource.roots ?? []).filter((c) =>
```
rugpanov, line 300: **The SSH picker inherits the cluster explorer's filter, which can hide stoppable clusters**

The connected path reads `clusterSource.roots`, which for `ClusterModel` applies the explorer's active filter (`ClusterModel.ts` `applyFilter`: `ALL` / `ME` / `RUNNING`). The default is `ALL`, so this only bites when the user has switched the explorer to the "Running" filter — but then terminated single-user clusters disappear from the SSH picker even though this flow passes `--auto-start-cluster` specifically to start stopped ones.

**Fix:** give the SSH picker the unfiltered cluster set, independent of the explorer filter (e.g. read from the loader / an unfiltered source rather than `ClusterModel.roots`).

packages/databricks-vscode/src/ssh/SshCommands.ts
```diff
@@ -0,0 +1,386 @@
...
+            /* eslint-enable @typescript-eslint/naming-convention */
+        };
+
+        const terminal = window.createTerminal({
+            name: "Databricks SSH Tunnel",
+            isTransient: true,
+            env,
+            strictEnv: false,
+        });
+        this.disposables.push(terminal);
```
rugpanov: **Nit: terminals accumulate for the extension lifetime**

`this.disposables.push(terminal)` grows unbounded — each `startTunnel` adds a terminal that is only released on `SshCommands.dispose()` (extension shutdown), which would also tear down an active tunnel on deactivate/reload. Consider not retaining the transient terminal on the instance, or tracking/removing it per-invocation via `window.onDidCloseTerminal`.

packages/databricks-vscode/src/ssh/SshCommands.ts
```diff
@@ -0,0 +1,386 @@
...
+                        // entry in the configuration section; the per-state icon
+                        // is dropped in favour of showing the state as text.
+                        label: `$(server) ${c.name!} (${c.id})`,
+                        detail: `${formatClusterState(
+                            c.state
+                        )} | ${formatQuickPickClusterDetails(c)}`,
+                        cluster: c,
+                    }))
+                );
+                this.preselect(quickPick);
```
rugpanov: **Nit: `preselect` re-applies `activeItems` on every refresh**

`refreshItems` runs on every `clusterSource.onDidChange` and re-runs `preselect`, resetting `activeItems`. With `onlyShowAccessibleClusters` enabled, `ClusterLoader._load` fires `onDidChange` once per cluster permission check, so the highlight can jump back to the preselected entry while the user is navigating. Consider preselecting only on the first populated repaint, or skipping the re-apply once the user has moved the selection.

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -574,6 +575,53 @@ export class CliWrapper {
...
+        return removeUndefinedKeys({
+            ...EnvVarGenerators.getEnvVarsForCli(
+                this.extensionContext,
+                workspaceConfigs.databrickscfgLocation
+            ),
+            ...EnvVarGenerators.getProxyEnvVars(),
+            ...this.getLogginEnvVars(),
+            ...authProvider.toEnv(),
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            DATABRICKS_OUTPUT_FORMAT: "text",
```
rclarey, line 623: I'm curious, I thought `text` is the default output format already so why is this needed explicitly?

rugpanov (changes requested):

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -574,6 +575,53 @@ export class CliWrapper {
...
+     * IDE window. Serverless is the default when no cluster is given.
+     *
+     * The --ide flag matches the host editor so the CLI opens the right remote
+     * window: Cursor identifies itself via env.uriScheme === "cursor",
+     * everything else (VS Code, Insiders) uses vscode.
+     *
+     * Logging is configured out of band via the DATABRICKS_LOG_* env vars (see
+     * getSshConnectEnvVars), so we do not pass --log-* flags here.
+     */
+    getSshConnectCommand(opts: {
```
rclarey: This is a pure function that doesn't need any state from `CliWrapper`, so I don't think it needs to be a method here. It can just be a standalone function

rugpanov:
**Integration tests are broken by this PR — verified not flaky.** Same specs fail on **both Windows and Linux** across all 3 isolated runs of this PR, while `main` is 32/32 e2e green the same day.

**Root cause:** this PR renames the command title `"Configure cluster"` → `"Configure compute"` in `package.json`, but the e2e suite still invokes the old title:

```ts
// run_dbconnect.ucws.e2e.ts:646
await executeCommandWhenAvailable("Databricks: Configure cluster");
```

Log: `Failed to execute Databricks: Configure cluster: Error: Command not found`.

**Failing specs:** `run_dbconnect`, `sync`, `deploy_and_run_pipeline` (all Win+Linux), and `bundle_variables` → *wait for extension activation* (`Can't find view section "CONFIGURATION"`).

Fix the test string, sweep the e2e suite for other old "Cluster" labels, and check the sync/pipeline/activation failures — they're consistent across every run, so a re-run won't clear them.

packages/databricks-vscode/src/ssh/SshCommands.ts
```diff
@@ -0,0 +1,386 @@
...
+                authProvider: workspace.authProvider,
+                userName: workspace.userName,
+                clusterSource: this.clusterModel,
+                ownsClusterSource: false,
+            };
+        }
+
+        const authProvider = await LoginWizard.run(this.cli);
+        if (authProvider === undefined || !(await authProvider.check())) {
+            return undefined;
```
rclarey, line 229: Shouldn't we show some message to the user here?

packages/databricks-vscode/src/ui/configuration-view/ClusterComponent.ts
```diff
@@ -114,7 +114,7 @@ export class ClusterComponent extends BaseComponent {
             // We are logged in -> Select cluster prompt
             return [
                 {
-                    label: LabelUtils.highlightedLabel("Select a cluster"),
+                    label: LabelUtils.highlightedLabel("Select a compute"),
```
rclarey: I think just "Select compute" sounds better
```suggestion
                    label: LabelUtils.highlightedLabel("Select compute"),
```

packages/databricks-vscode/src/ui/configuration-view/ClusterComponent.ts
```diff
@@ -123,7 +123,7 @@ export class ClusterComponent extends BaseComponent {
                     ),
                     id: TREE_ICON_ID,
                     command: {
-                        title: "Select a cluster",
+                        title: "Select a compute",
```
rclarey:
```suggestion
                        title: "Select compute",
```

packages/databricks-vscode/DATABRICKS.quickstart.md
```diff
@@ -48,12 +48,12 @@ The Databricks extension for Visual Studio Code enables you to connect to your r
 
 If your folder has multiple [Declarative Automation Bundles](#dabs), you can select which one to use by clicking "Open Existing Databricks project" button and selecting the desired project.
 
-## <a id="select-cluster"></a>Select a cluster
+## <a id="select-cluster"></a>Select a compute
```
rclarey:
```suggestion
## <a id="select-cluster"></a>Select compute
```

packages/databricks-vscode/DATABRICKS.quickstart.md
```diff
@@ -48,12 +48,12 @@ The Databricks extension for Visual Studio Code enables you to connect to your r
...
 
-The extension uses an interactive cluster to run code. To select an interactive cluster:
+The extension uses an interactive compute to run code. To select an interactive compute:
```
rclarey:
```suggestion
The extension uses interactive compute to run code. To select interactive compute:
```

packages/databricks-vscode/DATABRICKS.quickstart.md
```diff
@@ -48,12 +48,12 @@ The Databricks extension for Visual Studio Code enables you to connect to your r
...
 
 1. Open the Databricks panel by clicking on the Databricks icon on the left
-2. Click on the "Select Cluster" button.
+2. Click on the "Select Compute" button.
     - If you wish to change the selected cluster, click on the "Configure Cluster" gear icon, next to the name of the selected cluster.
```
rclarey:
```suggestion
    - If you wish to change the selected compute, click on the "Configure compute" gear icon, next to the name of the selected compute.
```

packages/databricks-vscode/package.json
```diff
@@ -539,10 +546,15 @@
...
+                {
+                    "id": "sshTunnelView",
+                    "name": "SSH Tunnel",
+                    "when": "!databricks.context.remoteMode"
+                },
                 {
                     "id": "clusterView",
                     "when": "databricks.feature.views.cluster && !databricks.context.remoteMode",
-                    "name": "Clusters"
+                    "name": "Computes"
```
rclarey:
```suggestion
                    "name": "Compute"
```

packages/databricks-vscode/package.json
```diff
@@ -1439,7 +1455,7 @@
                     "databricks.clusters.onlyShowAccessibleClusters": {
                         "type": "boolean",
                         "default": false,
-                        "description": "Enable/disable filtering for only accessible clusters (clusters on which the current user can run code)"
+                        "description": "Enable/disable filtering for only accessible computes (computes on which the current user can run code)"
```
rclarey:
```suggestion
                        "description": "Enable/disable filtering for only accessible compute (compute on which the current user can run code)"
```

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: This is refactoring better to be done in separate PR (as it'll take substantial changes). rules where introduced after PR

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: Wasn't able to reproduce picker hanging

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: Removed DATABRICKS_REMOTE_HOME_FOLDER (as it's not handled on remote IDE yet)

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: Done, wrapped in try catch

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: Done

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: resolving undefined onDidHide, changed order of disposing

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: Done, not including into disposables

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: Done, Doing preselection on first populated repaint

rugpanov:
🤖 Integration tests ❌ **7 of 35 test jobs failed** for `3e3a58bd` (28 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30558971510)

packages/databricks-vscode/src/cli/CliWrapper.ts
misha-db: It overrides the json default that [getEnvVarsForCli](https://github.com/databricks/databricks-vscode/blob/main/packages/databricks-vscode/src/utils/envVarGenerators.ts#L163) sets. ssh connect runs interactively in a terminal, so its output should be human readable text rather than JSON.

packages/databricks-vscode/src/cli/CliWrapper.ts
misha-db: Done. extracted as pure function

packages/databricks-vscode/src/ssh/SshCommands.ts
misha-db: No, as all check() implementations ( _check() method) are already displaying error to the user before returning false.
 
and in case of undefined it means user has cancelled/closed the view/flow

misha-db:
@rugpanov e2e tests are passing: https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30613304285

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `7c5bcb3d` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30616228546)

rugpanov (approved):
Approval is contingent on a committed follow-up to address my earlier unresolved comment on SshCommands.ts (#discussion_r3682676177). The error-swallowing sub-point there is already fixed; the core testability ask is not, and I don't want it to get lost — so let's land this and track the extraction as a fast-follow.

PR #2000, opened 2026-07-10
Upgrade ESLint 8.57 → 10.6 and migrate to flat config

rugpanov:
## Why

Dependabot opened #1985 and #1935 to bump `eslint` `8.57.1 → 10.6.0`, but a **bare version bump cannot work**: ESLint 9+ removed support for the legacy `.eslintrc` format entirely and requires the [flat config](https://eslint.org/docs/latest/use/configure/migration-guide) (`eslint.config.js/mjs`). Staying on 8.x is increasingly untenable — `.eslintrc` is end-of-life upstream.

The linter toolchain is **ready**: `@typescript-eslint` stable (8.62.x) already declares `eslint ^10` in its peer range, so no prerelease dependency is needed.

This is a **faithful migration**: it reproduces the exact enforcement of the old `.eslintrc` (same rules, same overrides, same TS-only scope) with **zero new findings**, so it's a pure infrastructure bump rather than a code-change PR. **Supersedes #1985 and #1935**, which can be closed.

## What

- Bump `eslint` 8.57 → 10.6, `eslint-config-prettier` 9.1 → 10.1, `eslint-plugin-no-only-tests` 3.1 → 3.4 across all three `package.json` files; add `@eslint/js` (flat-config recommended preset).
- Replace the two `.eslintrc.json` files with flat config:
  - **Root `eslint.config.mjs`**: shared base, ported 1:1 from the old root config (parser, recommended presets, all rule tweaks, both overrides, ignores).
  - **`packages/databricks-vscode/eslint.config.mjs`**: extends the base and adds the custom `local-rules/mutex-synchronised-decorator` rule.
  - **`packages/databricks-vscode-types/eslint.config.mjs`**: re-exports the base (previously inherited the root `.eslintrc` via the cascade).
- Drop the removed `--ext ts` flag from the four lint scripts; TS-only scope is now expressed by ignoring non-TS files in the flat config.
- **Faithfulness shims** (documented inline): `no-undef` off for TS (tsc handles it; matches typescript-eslint), `reportUnusedDisableDirectives` off (v9+ default would newly warn), and the two rules new to `eslint:recommended` in v10 (`no-useless-assignment`, `preserve-caught-error`) disabled. Renamed the retired `no-empty-interface` override to `no-empty-object-type` (typescript-eslint v8).

## Verification

- Established the **ESLint 8 baseline** first: `test:lint` exit 0 on both packages.
- After migration: `test:lint` **exit 0 on both packages** (eslint + prettier), matching the baseline — **zero new findings**.
- **Proved rules actively fire** (not silently disabled) via probe files: `local-rules/mutex-synchronised-decorator`, `no-only-tests`, and typescript-eslint rules all reported as expected.
- `tsc --build --force` **exit 0** on both packages.
- `eslint` resolves to **v10.6.0**; flat-config helpers resolve from stable `@typescript-eslint` sub-packages already in the tree (**no alpha deps added**).

## Backward compatibility

Dev-tooling only — no shipped extension code changes. The `.eslintrc → eslint.config.mjs` cutover lands **atomically** (both configs + all 4 script edits + deps in one commit) so there's no broken intermediate lint state for contributors.

⚠️ **Note (out of scope):** ESLint 10 requires `node ^20.19 || ^22.13 || >=24`; root `engines.node` is `">=22.0"`. CI runs node `22.x` (latest) so it's satisfied. Tightening `engines.node` is left out to keep this PR focused on the lint bump.

---


rugpanov:
🤖 Integration tests ❌ **3 of 35 test jobs failed** for `b2005b76` (32 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29082740290)

misha-db (approved):

rugpanov:
🤖 Integration tests ❌ **4 of 35 test jobs failed** for `b517282d` (31 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29085429897)

PR #2006, opened 2026-07-10
Harden run_dbconnect e2e test against Windows notebook flakiness

rugpanov:
## Why

The `run_dbconnect.ucws` Windows e2e shard intermittently fails with a raw `ENOENT` at `checkOutputFile` when a notebook cell's output JSON does not appear within the wait window. Investigation of a recent failed integration run showed:

- The old helper let the `ENOENT` abort the `waitUntil`, so the failure surfaced as an opaque IO stack instead of the intended readable timeout message — and gave **no signal** about *why* the output never landed (kernel bind? DBConnect session? cell error?).
- The 120s budget is too tight for a **cold serverless DBConnect session + kernel bind** on the slower Windows runner (the sibling install step already documents needing 180s on Windows).
- The `Python Environments... -> .venv` kernel quick-pick is a known-flaky UI interaction.

The failure is pre-existing flakiness, not tied to any product change.

## What

All changes are confined to `packages/databricks-vscode/src/test/e2e/run_dbconnect.ucws.e2e.ts`:

- **`checkOutputFile`**: a missing file is now a normal poll-miss (keep polling) instead of an `ENOENT` throw, so a genuine timeout surfaces the readable `timeoutMsg`; other IO errors still propagate. Captures the last content seen and accepts a configurable `timeout`.
- **`dumpNotebookDiagnostics`** (new, failure-only): logs the last file content, the `nested/` directory listing, and the VS Code Output panel text — where kernel/DBConnect/cell-execution errors surface.
- **Cold-start budget**: the first `checkOutputFile` in each notebook test gets 180s; warm subsequent checks keep the 120s default.
- **Kernel selection retry**: the two-step `.venv` kernel quick-pick is wrapped in a bounded `waitUntil` retry.

Test-only change: no user-facing surface, config, or persisted state.

## Verification

- `tsc --noEmit` clean (workspace-local TypeScript 6.0.3).
- ESLint 10.6 (flat config) passes; Prettier passes.
- Full runtime verification requires the Windows serverless CI shard against the Databricks test workspace and cannot be run locally; real proof is a CI re-run of the `run_dbconnect` shard.

This pull request and its description were written by Isaac.

PR #2013, opened 2026-07-14
test(e2e): capture notebook cell errors in run_dbconnect diagnostics

rugpanov:
## Summary

Makes the `run_dbconnect.ucws` e2e test capture **why** a notebook/magic cell fails, so the current flaky failures can be root-caused from CI logs instead of only from the `.webm` video.

## Background

The `ensureVenvHasKernelDeps` fix (#2006) works — the kernel now starts and the notebook's first cell runs. But two failures remain (seen in run 29318458802):
- **notebook test:** `Notebook: Run All` completes cell 1, but the 2nd cell (`%run "./hello.py"`) never executes (video shows cell `[ ]`, status bar "Cell 1 of 2").
- **magic test:** the `%sql` → `_sqldf` cell doesn't emit its output — fails on **both** Linux and Windows.

We couldn't see the actual cell error because `dumpNotebookDiagnostics` crashed enumerating Output channels:
```
could not list Output channels: Can't call $$ on element with selector "select[title="Tasks"]"
=== Output channels available === []
```
That `select[title="Tasks"]` locator (from wdio-vscode-service) is stale on the pinned VS Code `^1.86.0`, so the kernel's traceback never reached CI logs.

## What this does (diagnostics only — no product/assertion changes)

- **`dumpActiveNotebookCells()`** — reads the active notebook's cells, execution summaries (`executionOrder`/`success`), and output items via the VS Code API through `browser.executeWorkbench`. No DOM selector, so a cell error/traceback or a never-run cell is captured directly.
- **`dumpVscodeLogFiles()`** — dumps VS Code's on-disk logs (extension host + Output-channel logs) resolved from `vscode.env.logUri` — another selector-independent source for the kernel/DBConnect error.
- Both run ahead of the existing Output-channel scrape, which is kept as best-effort (its failure was already caught; now clearly documented as a stale locator).
- Runs **only on the failure path**, so the happy path is unaffected.

## Verification

- `tsc --noEmit -p src/test/e2e/tsconfig.json`: 0 errors.
- eslint + prettier on the file: clean.
- The e2e itself runs only in the databricks-eng CI; the purpose of this change is to make the next run capture the cell error so the underlying `%run`/`%sql` execution failure can be diagnosed and fixed as a follow-up.

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/run_dbconnect.ucws.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/run_dbconnect.ucws.e2e.ts
@@ -97,6 +97,95 @@ async function listFilesRecursive(dir: string, base = dir): Promise<string[]> {
     return out;
 }
 
+// Dumps the active notebook's cell outputs straight from the VS Code API. This
+// is the ground truth for "did the cell error out?": a cell that raised shows
+// up here as an `error` output item with the name/message/stack, and a cell
+// that never ran has an empty `executionOrder`. We read it via
+// `executeWorkbench` (runs inside the extension host with the full `vscode`
+// API) so it does not depend on any DOM selector — unlike the Output-channel
+// scrape below, whose `select[title="Tasks"]` locator is stale on current
+// VS Code and silently returns nothing.
+async function dumpActiveNotebookCells() {
+    try {
+        const cells = await browser.executeWorkbench((vscode) => {
+            const editor = (vscode.window as any).activeNotebookEditor;
+            const doc = editor?.notebook;
+            if (!doc) {
+                return "<no active notebook editor>";
+            }
+            return doc
+                .getCells()
+                .map((cell: any) => {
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests ❌ **5 of 35 test jobs failed** for `26e7b058` (30 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29332204304)

misha-db (approved):

rugpanov:
🤖 Integration tests triggered for `5b80ab31` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29343001620)

rugpanov:
🤖 Integration tests ❌ **6 of 35 test jobs failed** for `28c6ffc8` (29 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29344928263)

PR #2016, opened 2026-07-16
Enable Unity Catalog (and Docs) in Databricks Remote SSH mode

misha-db:
## Changes

When the extension runs inside a Databricks Remote SSH session ("remote mode"), it previously short-circuited activation and surfaced no data views. This PR surfaces the Unity Catalog browser (and the Documentation view) in remote mode, connecting Unity Catalog directly from the ambient environment credentials — no bundle/config project (host + target) required.

## Tests

ConnectionManager.test.ts


rugpanov:
🤖 Integration tests ❌ **5 of 35 test jobs failed** for `5f72aa0e` (30 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29494371075)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `4ac9f0fe` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30434657225)

packages/databricks-vscode/package.json
```diff
@@ -564,12 +564,12 @@
                 {
                     "id": "unityCatalogView",
                     "name": "Unity Catalog",
-                    "when": "databricks.context.activated && databricks.context.loggedIn"
+                    "when": "databricks.context.activated && (databricks.context.loggedIn || databricks.context.remoteMode)"
```
rugpanov, line 579: **Must-address: the remote reconnect/refresh button is disabled in exactly the failure state it exists to recover from.**

The new `reconnect` wiring exists so the refresh command re-establishes a failed connection in remote mode (`extension.ts` `connectRemote`). But the `databricks.unityCatalog.refresh` command keeps `"enablement": "databricks.context.activated && databricks.context.loggedIn"` (package.json:259). In remote mode `loggedIn` is set `true` only on `updateState("CONNECTED")`; while `CONNECTING` and after a failed connect (`DISCONNECTED`) it's `false`. So when the initial `connectRemote()` fails and the tree shows the `Not connected: <error>` node, the toolbar refresh button is greyed out. I confirmed there is no other trigger — no `commandPalette` override and no `viewsWelcome` entry for `unityCatalogView` — so the user's only recovery is a full window reload.

Suggest relaxing the command `enablement` at package.json:259 to mirror this view's `when`, e.g. `databricks.context.activated && (databricks.context.loggedIn || databricks.context.remoteMode)`.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -261,6 +278,67 @@ export class ConnectionManager implements Disposable {
...
+            // already hold it would deadlock. We also deliberately skip the
+            // sync/cluster/config-project machinery they run, since remote mode
+            // only needs a workspace client for the Unity Catalog view.
+            this._connectionError = undefined;
+            this.updateState("CONNECTING");
+            try {
+                // The authProvider is only injected by tests; in production it
+                // is resolved from the SDK's default credential chain.
+                if (authProvider === undefined) {
+                    const config = new Config({});
```
rugpanov: **Must-address (please confirm intent): `new Config({})` resolves the full SDK credential chain, not only the injected environment.**

`ensureResolved()` runs the default loaders `[EnvironmentLoader, KnownConfigLoader]`. `EnvironmentLoader` wins when `DATABRICKS_HOST`/`DATABRICKS_TOKEN` are set, but if they're absent `KnownConfigLoader` silently falls back to the `DEFAULT` profile in `~/.databrickscfg`. So the `config.token !== undefined` check below doesn't prove the token came from the remote-injected env — a stray local `DEFAULT` PAT profile would satisfy it and connect to a possibly-wrong workspace instead of failing fast, which contradicts the method's doc-comment ("the remote environment always provides a token, so we fail fast if one isn't present").

If only the injected env should be honored, read `DATABRICKS_HOST`/`DATABRICKS_TOKEN` directly (or pass explicit values into `Config`) rather than relying on default resolution.

rugpanov (commented):
Review via Codex + Claude review agent + a manual pass. Core mechanism (env-based PAT connect, inlined mutex logic to avoid the non-reentrant `loginLogoutMutex` deadlock, status-node rendering, and the `registerUnityCatalog`/`registerDocsView` refactor) is correct and well-tested — `this`-binding in the refactored refresh command is preserved, there's no double registration (remote branch returns early), and no init-render race (the provider subscribes to `onDidChangeState`). `when`-clause changes are additive/backward-compatible.

Every comment below was verified against the code/SDK before posting; one candidate finding (that the `connectRemote` `.catch` might log the PAT) was checked and dropped — the SDK only logs request headers to its own internal context logger, and the thrown `ApiError` carries just a message, so there's no leak.

**CI:** unit tests + packaging pass. The two red checks are not PR-caused: *Trigger Tests* failed with "exceeded max execution time while awaiting a runner for 24h" (infra), and *Integration Tests* failed only on the Windows `run_dbconnect.*.e2e.ts` shard (databricks-connect/serverless-setup — the known flake bucket; this PR touches none of those files). Worth a re-run.

packages/databricks-vscode/src/extension.ts
```diff
@@ -271,6 +449,66 @@ export async function activate(
...
+        const remoteConnectionManager = new ConnectionManager(
+            cli,
+            remoteConfigModel,
+            workspaceFolderManager,
+            customWhenContext,
+            telemetry
+        );
+        context.subscriptions.push(
+            remoteBundleFileWatcher,
+            remoteConfigModel,
```
rugpanov, line 514: [nit] The four models passed to `remoteConfigModel` (`BundleValidateModel`, `OverrideableConfigModel`, `BundlePreValidateModel`, `BundleRemoteStateModel`) are constructed above but not pushed to `context.subscriptions` — only the watcher, `remoteConfigModel`, and `remoteConnectionManager` are. `ConfigModel.dispose()` disposes only its own listeners, not its child models (the normal flow pushes all four individually). Impact is bounded (one activation, no CLI work since target stays undefined), but it's an inconsistency. Better still: since `connectFromEnvironment()` never reads the `ConfigModel` (per the comment above), consider not building the bundle stack here at all and just satisfying the constructor.

packages/databricks-vscode/src/configuration/ConnectionManager.ts
```diff
@@ -261,6 +278,67 @@ export class ConnectionManager implements Disposable {
...
+    async connectFromEnvironment(authProvider?: AuthProvider): Promise<void> {
+        await this.loginLogoutMutex.synchronise(async () => {
+            // We intentionally inline the connect/disconnect steps here rather
+            // than delegating to _connect()/disconnect(): both of those acquire
+            // loginLogoutMutex, which is non-reentrant, so calling them while we
+            // already hold it would deadlock. We also deliberately skip the
+            // sync/cluster/config-project machinery they run, since remote mode
+            // only needs a workspace client for the Unity Catalog view.
+            this._connectionError = undefined;
+            this.updateState("CONNECTING");
```
rugpanov, line 348: [nit] `updateState("CONNECTING")` here doesn't clear `_workspaceClient`/`_databricksWorkspace` first (only the `catch` clears them). On a refresh-triggered reconnect, a concurrent `getChildren()` — which checks only `workspaceClient` — could briefly use the old client while re-authenticating. Low impact in remote mode since credentials don't change, but clearing both before entering `CONNECTING` (or having the provider treat non-`CONNECTED` as unavailable) would make it airtight.

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.test.ts
```diff
@@ -175,15 +175,53 @@ describe(__filename, () => {
         disposables.forEach((d) => d.dispose());
     });
 
-    it("returns undefined when not connected", async () => {
+    it("shows a status node at the root when not connected", async () => {
```
rugpanov, line 178: [nit] Nice coverage of the `DISCONNECTED` empty/error branches. The `state === "CONNECTING"` branch of `getDisconnectedRootChildren` (the `Connecting…` node) is still untested — worth a case stubbing `state` = `"CONNECTING"` and asserting a single `empty` node with message `Connecting…`.

packages/databricks-vscode/src/configuration/ConnectionManager.test.ts
```diff
@@ -1,24 +1,110 @@
...
+            new URL("https://test.databricks.com")
+        );
     });
 
     afterEach(() => {
         disposables.forEach((d) => d.dispose());
+        reset(mockConfigModel);
+    });
+
+    it("connectFromEnvironment connects using the injected auth provider", async () => {
```
rugpanov, line 67: [nit] All three tests inject an `AuthProvider`, so the production credential-resolution path (`new Config({})`, missing host/token, PAT-only enforcement, `normalizeHost`) is never exercised — which is where the credential-fallback concern above lives. Consider a test around the no-authProvider branch. An activation-level test covering "initial connect fails → refresh reconnects" would also lock down the recovery flow.

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `36e88d90`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30616218647)

packages/databricks-vscode/package.json
misha-db: Done

packages/databricks-vscode/src/configuration/ConnectionManager.ts
misha-db: Done

packages/databricks-vscode/src/extension.ts
misha-db: Done

packages/databricks-vscode/src/configuration/ConnectionManager.ts
misha-db: Done

packages/databricks-vscode/src/ui/unity-catalog/UnityCatalogTreeDataProvider.test.ts
misha-db: Done

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `04216ebd`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30667409731)

rugpanov (approved):

PR #2018, opened 2026-07-16
fix(bundle): match modern CLI run URL in JobRunStatus.parseId

anton-107:
## Summary
Fixes a product bug where the Bundle Resource Explorer shows a permanent, **false** "Timeout while fetching run status" for a job that actually **succeeded**.

The databricks CLI changed the run URL it prints from the legacy `#job/<id>/run/<id>` fragment to the modern `/jobs/<id>/runs/<id>?o=..` path. `JobRunStatus.parseId` only matched the singular `/run/` form, so with a current CLI the extension never captures the run id — `startPolling()` never runs, and the 60s no-run-id timer flips the tree item to a timeout state even though the run completed fine.

## Fix
- Match both `/run/` and `/runs/` with `/\/runs?\/(\d+)/`.
- Use `\d+` instead of `\d*` so a stdout chunk that splits right after `/run(s)/` can't yield an empty capture (`parseInt("") === NaN`).
- Add `JobRunStatus.test.ts` pinning the CLI-output contract for both the modern and legacy URL forms (there was no unit coverage on this parser, which is how the regression shipped).

## Evidence
Found via the vscode nightly e2e run ([https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29475140909](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29475140909)). The CLI `--debug` log shows `POST /api/2.2/jobs/run-now` succeeding and the run reaching `SUCCESS`/`TERMINATED`, with the CLI printing `Run available at https://.../jobs/<id>/runs/<id>?o=...`, while the extension tree item stayed at "Timeout while fetching run status" for the full window. Verified in node that the new regex captures the id from the modern URL, the `Run available at` form, and the legacy fragment, and does not mis-capture the job id from `/jobs/<id>`.

## Impact
Affects every user whose bundled CLI prints the modern run URL — candidate for a point release.

### DIFF
--- a/packages/databricks-vscode/src/bundle/run/JobRunStatus.test.ts
+++ b/packages/databricks-vscode/src/bundle/run/JobRunStatus.test.ts
@@ -1,0 +1,100 @@
+import {expect} from "chai";
+import {mock, instance, when} from "ts-mockito";
+import {install, Clock} from "@sinonjs/fake-timers";
+import {JobRunStatus} from "./JobRunStatus";
+import {AuthProvider} from "../../configuration/auth/AuthProvider";
+
+// These tests pin the CLI-output contract that JobRunStatus.parseId depends on.
+// The run id is scraped from the run URL the `databricks bundle run` CLI prints
+// to the terminal. When the CLI changed the URL format from the legacy
+// "#job/<id>/run/<id>" fragment to the modern "/jobs/<id>/runs/<id>?o=.." path
+// (databricks/cli), the old singular-"/run/" regex silently stopped matching and
+// the extension showed a permanent false "Timeout while fetching run status" for
+// jobs that actually succeeded. These tests guard against that class of drift.
+describe("JobRunStatus.parseId", () => {
+    let clock: Clock;
+    let authProvider: AuthProvider;
+
+    beforeEach(() => {
+        // The BundleRunStatus base class arms a 60s "no run id -> timeout" timer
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests triggered for `361d8f30` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29494422973)

anton-107:
### ✅ Validated in CI

The nightly PR pipeline ([`vscode-isolated-pr` run 29494422973](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29494422973)) confirms the fix against the real CLI:

- `deploy_and_run_job.e2e.ts` — **PASSED** on the Linux shard. The tree item reached `Run status: Success`, whereas before this change it hung on the false `Timeout while fetching run status` for the full window.
- Unit tests (incl. the new `JobRunStatus.test.ts`) pass.

The two remaining Linux failures on this run are **pre-existing and unrelated** to this PR:
- `run_dbconnect.ucws` — the notebook "Run All" flakiness (tracked separately in the run_dbconnect de-flake PR).
- `deploy_and_run_pipeline` — a genuine backend **pipeline update failure** (`Update … is FAILED`); `PipelineRunStatus.parseId` polled correctly, so this is not a parsing issue and is out of scope here.

rugpanov (approved):

PR #2025, opened 2026-07-17
Add AI tools integration

rclarey:
## Changes

Integrate the `databricks aitools` family of commands into the extension. 

- on install (or first launch of the extension with this feature) it prompts to install AI tools
- add `AI tools` row to the configuration pane

## Tests

Added tests pass


rugpanov:
🤖 Integration tests ❌ **2 of 35 test jobs failed** for `4b8a5593` (33 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29570566009)

rugpanov:
🤖 Integration tests ❌ **2 of 35 test jobs failed** for `fb06b04b` (33 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29731848252)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `cda54dbe`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30550061358)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `2f661600`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30810957412)

packages/databricks-vscode/src/extension.ts
```diff
@@ -344,15 +416,11 @@ export async function activate(
         // We show a welcome view when there's no workspace folders, prompting users
         // to either open a new folder or to initialize a new databricks project.
         // In both cases we expect the workspace to be reloaded and the extension will
-        // be activated again.
+        // be activated again. The AI tools setup affordance is added as a
+        // viewsWelcome entry (see package.json) since it works without a folder.
```
misha-db: stale comment? no viewsWelcome in package.json

packages/databricks-vscode/src/extension.ts
```diff
@@ -313,6 +323,67 @@ export async function activate(
...
+        telemetry.registerCommand(
+            "databricks.aitools.installAgent",
+            aiToolsCommands.installAgentCommand(),
+            aiToolsCommands
+        )
+    );
+    // Detect install state on activation and, if installed, auto-apply any
+    // available update; otherwise prompt the user (once) to install the tools.
+    // Non-blocking so it doesn't delay activation.
+    aiToolsCommands.initializeCommand()();
```
misha-db: Can initializeCommand() throw? (try/catch)

misha-db: Also it's running in remote mode, even though commands are gated to not run

packages/databricks-vscode/src/extension.ts
```diff
@@ -300,6 +305,11 @@ export async function activate(
             "databricks.bundle.showLogs",
             () => loggerManager.showOutputChannel("Databricks Bundle Logs"),
             loggerManager
+        ),
+        telemetry.registerCommand(
+            "databricks.logs.show",
```
misha-db: looks like duplicate of existing "databricks.internal.showOutput" command

misha-db: and has no enablement (will appear in remote mode)

packages/databricks-vscode/src/aitools/AiToolsManager.ts
```diff
@@ -0,0 +1,556 @@
...
+                cursorPlugin = true;
+                void this.addCursorPlugin(source);
+            }
+            cliAgents = agents.filter((a) => a !== CURSOR_AGENT_ID);
+            // The user picked *only* the Cursor plugin: there are no skills to
+            // install via the CLI. Record the install (the plugin) and bail out
+            // before the CLI call — passing an empty `--agents` list would make
+            // the CLI act on every detected agent, which is not what was chosen.
+            if (agents.length > 0 && cliAgents.length === 0) {
+                this.telemetry.recordEvent(Events.AITOOLS_INSTALL, {
```
misha-db, line 452: Here AITOOLS_INSTALL is reported as success (Cursor specific case), but where do we install it? 
addCursorPlugin() call before seems to only open a modal

packages/databricks-vscode/src/cli/CliWrapper.test.ts
```diff
@@ -349,6 +387,29 @@ token = dapitest5678
...
+    it("completes a stdin-reading process when closeStdin is set", async () => {
+        const {stdout} = await cancellableExecFile("cat", [], {}, undefined, {
+            closeStdin: true,
+        });
+        assert.strictEqual(stdout, "");
+    });
+
+    it("hangs on a stdin-reading process without closeStdin", async () => {
+        const raced = await Promise.race([
+            cancellableExecFile("cat", []).then(() => "completed"),
```
misha-db: minor: "cat" process seems to stay alive/leaking

packages/databricks-vscode/src/vscode-objs/CustomWhenContext.ts
```diff
@@ -116,4 +116,28 @@ export class CustomWhenContext {
             value
         );
     }
+
+    setIsCursor(value: boolean) {
+        commands.executeCommand(
+            "setContext",
+            "databricks.context.isCursor",
```
misha-db: No function reads "databricks.context.isCursor" key. is it needed?

misha-db: and no "when" clause is checking it in package.json

packages/databricks-vscode/src/aitools/AiToolsManager.ts
```diff
@@ -0,0 +1,556 @@
...
+    }));
+}
+
+/**
+ * Owns all non-UI logic for the Databricks AI tools feature: detecting whether
+ * tools are installed (and where), running install/update via the CLI, checking
+ * for available updates, and caching the resolved install location.
+ */
+export class AiToolsManager implements Disposable {
+    private disposables: Disposable[] = [];
```
misha-db, line 97: nit: nothing is pushed in disposables

packages/databricks-vscode/src/ui/configuration-view/AiToolsComponent.ts
```diff
@@ -0,0 +1,258 @@
...
+                    description: installLocation,
+                    collapsibleState: TreeItemCollapsibleState.None,
+                },
+                version !== undefined && {
+                    label: "Version",
+                    id: getTreeIconId("version"),
+                    description: version,
+                    collapsibleState: TreeItemCollapsibleState.None,
+                },
+                agents !== undefined && {
```
misha-db: agents !== undefined is always true as it defaults to [] in model

packages/databricks-vscode/src/ui/configuration-view/AiToolsComponent.ts
```diff
@@ -0,0 +1,258 @@
...
+            {
+                label: "AI tools",
+                id: TREE_ICON_ID,
+                description: description ?? "",
+                tooltip: `AI tools installed (${installLocation})`,
+                contextValue: getContextValue(state),
+                iconPath: icon,
+                collapsibleState: TreeItemCollapsibleState.Collapsed,
+            },
+        ];
```
misha-db, line 104: In case installLocation !== undefined (maybe cached) but detectError is true. it will show collapsed node and when user clicks it will expand to empty node (as getChildren returns [] in case of detectError)

is this intentional?

misha-db (changes requested):

packages/databricks-vscode/src/aitools/AiToolsManager.ts
rclarey: Installing the cursor plugin via the modal opened by `addCursorPlugin` is not observable by this extension. In this case (`agents.length > 0 && cliAgents.length === 0`) only cursor is selected so we don't need to run the CLI, therefore we pass `success: true` since there is no other success signal we can use

packages/databricks-vscode/src/aitools/AiToolsManager.ts
rclarey: yeah I know, I was just following the boilerplate. I can remove it 👍

packages/databricks-vscode/src/ui/configuration-view/AiToolsComponent.ts
rclarey: Yeah probably it's better to just show the error state in all cases and ignore any cached state

packages/databricks-vscode/src/ui/configuration-view/AiToolsComponent.ts
rclarey: good catch. This should be `agents.length > 0`

packages/databricks-vscode/src/vscode-objs/CustomWhenContext.ts
rclarey: This must have been left behind at some point. I'll remove 👍

packages/databricks-vscode/src/extension.ts
rclarey: I'll remove this then 👍

packages/databricks-vscode/src/extension.ts
rclarey: Yeah it seems it can if `window.withProgress` throws, but it probably shouldn't. I'll update this to show an error toast instead 👍

packages/databricks-vscode/src/extension.ts
rclarey: Yes, whoops! Will remove

packages/databricks-vscode/package.json
```diff
@@ -37,6 +37,7 @@
         "url": "https://github.com/databricks/databricks-vscode.git"
     },
     "activationEvents": [
+        "onStartupFinished",
```
misha-db: Why do we need this? it makes extension activate for all cases, even when user opens some unrelated (to databricks) project. We should try to avoid it as it makes VSCode UX slow

misha-db: Also it will force activating all dependent extensions (python, yaml, debugger, etc..)

rclarey: It's needed to prompt to install the AI tools when the IDE launches

packages/databricks-vscode/src/aitools/AiToolsManager.ts
misha-db: If it's not observable by extension I would suggest not to report as success, maybe another status "possible success" or similar, as it will inflate success number were we aren't sure

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `f219908b` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30988022383)

packages/databricks-vscode/src/aitools/AiToolsManager.ts
rclarey: It's still a "success" in that the user completed the multi-step install flow.

I can update this to be more explicit, but I worry that with a `"possible-success"` status we might naively look at `total - success` and imply a bigger amount of error results than there actually are. We'll have to be careful in how we track this

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -136,6 +136,70 @@ export function getSshConnectCommand(opts: {compute: SshConnectCompute}): {
...
+     * only received skills the UI annotates the row as "skills only".
+     */
+    delivery?: "skills" | "plugin";
+}
+
+/** A single agent entry from `databricks aitools list --output json`. */
+export interface AiToolsAgent {
+    name: string;
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    display_name: string;
```
rugpanov, line 190: **Blocker: blocked on CLI v1.11.0.**

`AiToolsAgent` reads `display_name`, `detected`, and `installed[].delivery`. Those fields were added by [cli#6020](https://github.com/databricks/cli/pull/6020) (`7d8db8e4`), which is **22 commits ahead of `v1.10.0`** — so v1.10.0 won't work either. In both v1.9.0 (currently pinned) and v1.10.0, `agentEntry` is only `{name, managed, installed}` and `agents` is `omitempty`, so it's usually absent entirely.

On a shipped CLI this fails end-to-end, not just in CI: `resolveInstalled()` throws on `undefined.map` → row stuck at "Update check failed"; `listAgents()` swallows it → `[]` → the agent picker is silently skipped → `--agents` is omitted → the CLI installs for every detected agent.

Needs v1.11.0 released **and** `package.json` → `cli.version` bumped (this PR still inherits `1.9.0` from main).

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -500,6 +578,133 @@ export class CliWrapper {
...
+        cwd: string,
+        cancellationToken?: CancellationToken,
+        agents?: string[],
+        @context ctx?: Context
+    ): Promise<void> {
+        const args = ["aitools", "install", "--scope", scope];
+        // When a specific set of agents is chosen, restrict the install to them
+        // (`--agents a,b`). With no selection the CLI acts on every detected
+        // agent, which is the desired default.
+        if (agents && agents.length > 0) {
```
rugpanov: **Blocker: an explicitly empty selection installs for everything.**

This only passes `--agents` when the list is non-empty, so "user deselected everything" and "we have no agent info" both fall through to the CLI default of **installing for every detected agent** — the opposite of an empty selection.

`pickAgents` also returns `[]` on a list failure (`AiToolsCommands.ts:241`), so a transient CLI error silently becomes a broad install that touches agents the user never chose.

Suggest making the two cases distinct: `undefined` = "unknown, use the CLI default", `[]` = "user chose nothing → cancel", and have `install()` bail early on `[]`.

packages/databricks-vscode/src/aitools/AiToolsManager.test.ts
```diff
@@ -0,0 +1,980 @@
...
+    global?: StateFileLoader;
+}
+
+function listResult(
+    skills: Array<{
+        name: string;
+        latest_version: string;
+        installed: Record<string, string>;
+    }>,
+    agents: AiToolsAgent[] = []
```
rugpanov, line 58: **Medium: the `agents` contract has no test that would fail if the CLI omitted it** — which is why this reached CI.

1. `AiToolsListResult.agents` is declared **required**, but it comes from an unvalidated `JSON.parse(res.stdout) as AiToolsListResult` (`CliWrapper.ts:705`), so the compiler trusts a shape the CLI doesn't guarantee. Suggest `agents?: AiToolsAgent[]` plus `result.agents ?? []` at `AiToolsManager.ts:332` and `:361`.
2. This helper defaults `agents` to `[]`, so all 37 manager tests inject the field. Worth adding cases for `agents` absent and for the legacy `{name, managed, installed}` shape.
3. The one test that touches real CLI output (`CliWrapper.test.ts:67`) asserts on `agents` unconditionally, so it fails opaquely on any older CLI — "expected true, got false" rather than "bundled CLI lacks agents". Either gate it on the pinned version or assert the pinned contract explicitly.

Item 1 is worth doing regardless of the version bump: users can point the extension at an older CLI via `DATABRICKS_CLI_PATH`, so tolerating the legacy shape prevents a hard break there too.

packages/databricks-vscode/src/aitools/AiToolsManager.ts
```diff
@@ -0,0 +1,555 @@
...
+        }
+    }
+
+    /**
+     * Check whether an update is available by comparing each installed skill's
+     * version against its latest version (via `aitools list --output json`).
+     * `aitools update --check` only prints text, so `list` is the reliable
+     * source of truth.
+     */
+    async resolveInstalled(): Promise<void> {
```
rugpanov, line 374: **Medium: no serialization between operations** (UI state, self-healing).

`install` / `update` / `uninstall` / `resolveInstalled` have no serialization, so they can interleave. `checkForUpdatesCommand()` (`AiToolsCommands.ts:289-292`) calls `resolveInstalled()` directly while `update()` is still blocked on its CLI subprocess (`:542`), and both write the model unguarded (`:355`, `:359-363`, `:552`).

Concretely: click Update, then Check for Updates while it runs → the second call's `updateStatus: "checking"` overwrites the first's resolved result, so the row shows "Checking" right after the user watched the update finish.

State does converge (every path reconciles via `resolveInstalled()` in a `finally`), so this is wrong-transient-UI rather than corruption — hence medium, not blocking.

The repo already has the idiom for this — `BundleRemoteStateModel.ts:53,68`:
```ts
protected mutex = new Mutex();

@Mutex.synchronise("mutex")
async update(token?: CancellationToken): Promise<void> { ... }
```
Applying that to the four methods serializes them in ~2 lines.

rugpanov (commented):
Review notes on the AI tools integration. Two blockers (the CLI version dependency and the empty-agent-selection fallback) plus two medium items, left as inline comments.

On sequencing: the `--agents` fix is independent of the CLI release and worth landing now. It also shares a root cause with the typing item — both stem from `[]` being overloaded to mean two different things — so those two go well together.

rugpanov (changes requested):

packages/databricks-vscode/src/cli/CliWrapper.ts
rclarey: This will be resolved after this PR merges. I'll skip the problematic test until the CLI bump

packages/databricks-vscode/src/aitools/AiToolsManager.test.ts
rclarey: Not actionable. We will only release this with a CLI which outputs `agents`, so this issue is not real

rugpanov:
🤖 Integration tests triggered for `9367e012` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31010172052)

misha-db (approved):

rugpanov:
🤖 Integration tests ❌ **3 of 35 test jobs failed** for `4f719bec` (32 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31011506415)

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -500,6 +578,138 @@ export class CliWrapper {
...
+            await execFile(
+                this.cliPath,
+                args,
+                {cwd, env: this.aitoolsEnv()},
+                cancellationToken,
+                {closeStdin: true}
+            );
+        } catch (e: any) {
+            ctx?.logger?.error("Failed to install Databricks AI tools", e);
+            throw new ProcessError(e.message, e.code ?? null);
```
rugpanov, line 628: **Cancelling an AI tools operation is reported to the user as a failure.** (Non-blocking, but please fix asap — see below on the telemetry.)

The progress notifications are `cancellable: true` (`AiToolsCommands.ts:65`) and cancellation aborts the child process via `AbortController` (`:73`). Node rejects an aborted `execFile` with an `AbortError` (`code: "ABORT_ERR"`) — confirmed locally:

```
name=AbortError code=ABORT_ERR killed=undefined signal=undefined msg="The operation was aborted"
```

This `catch` treats any rejection as a CLI failure, so cancelling produces a **"Failed to install Databricks AI tools." error toast** plus `result: "error"` telemetry (`AiToolsManager.ts:474`). Same pattern at `:653` (update) and `:678` (uninstall).

Most visible on the activation-time auto-update, which applies silently but *is* cancellable — a user who cancels an update they never requested gets an error toast. The toast is cosmetic, but the telemetry isn't: cancellations will inflate the error rate and that can't be cleaned up after the fact, so I'd like this in before the release that turns the feature on for users.

Suggest distinguishing cancellation before wrapping: rethrow a dedicated cancellation error (or check `cancellationToken?.isCancellationRequested`, the pattern used in ~10 places in this repo, e.g. `sdk-extensions/Cluster.ts:226`), then have `withProgress` swallow it and the manager record it as cancelled rather than `error`.

packages/databricks-vscode/src/aitools/AiToolsManager.ts
```diff
@@ -0,0 +1,594 @@
...
+        // `activeProjectUri` throws when no folder is active; treat that as
+        // "no project folder" rather than propagating.
+        try {
+            return this.workspaceFolderManager.activeProjectUri !== undefined;
+        } catch {
+            return false;
+        }
+    }
+
+    private get projectRoot(): string {
```
rugpanov, line 203: **No re-detect when the active project changes.** (Non-blocking — fine as a follow-up.)

`AiToolsManager` never subscribes to `workspaceFolderManager.onDidChangeActiveProjectFolder`, but `projectRoot` (and therefore `cwdForScope("project")`) reads `activeProjectUri` live on every call. So after the active project changes, the model's `installLocation` / `version` / `agents` still describe the *previous* folder while the cwd now points at the new one.

The trigger isn't casual folder switching — `setActiveProjectFolder` has only two callers: the explicit "Choose a project" flow (`databricks.bundle.selectActiveProjectFolder` → `activeBundleUtils.ts:79`) and the fallback when the active folder is removed (`WorkspaceFolderManager.ts:59`). So the repro is: a workspace with multiple Databricks projects where A has project-scope AI tools and B doesn't; use "Choose a project" to switch to B. The AI tools row still shows installed (project) with A's version, and Update/Uninstall run against B.

Every other collaborator on this event re-syncs on it — `BundleProjectManager.ts:58`, `BundleWatcher.ts:43`, `BundleFileSet.ts:104`, `ConnectionManager.ts:243`, `WorkspaceFolderComponent.ts:12`. Suggest the same here: subscribe in the constructor and call `detectInstall()` on change (it already reconciles the model and the when-context).

No test covers a project switch, so the current suite wouldn't catch this.

rugpanov (approved):
LGTM — approving. The mutex serialization and the empty-agent-selection guard both look good now, and the test coverage is genuinely thorough (the concurrency test at `AiToolsManager.test.ts:755` would actually fail if serialization broke).

Two things I'd like fixed as soon as possible, neither blocking this merge:

1. **Cancelling an operation is reported as a failure** — cosmetically a wrong toast, but it also records `result: "error"` telemetry, so every cancellation inflates the AI-tools error rate. Worth landing before the release that enables this for users, since that data can't be cleaned up retroactively.
2. **No re-detect when the active project changes** — narrow repro (multi-root, multiple Databricks projects, explicit "Choose a project"), and recoverable via the reload affordance, but the model and the CLI cwd disagree until something re-detects.

Details inline.

rugpanov:
🤖 Integration tests triggered for `23e5813c` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31088524153)

rugpanov:
🤖 Integration tests ❌ **8 of 35 test jobs failed** for `2fe766d6` (27 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31088997259)

PR #2033, opened 2026-07-22
test(e2e): harden bundle_variables override against watcher refresh race

rugpanov:
## Why
`bundle_variables :: should override default variable` flakes: the test writes `{varWithDefault:"new value"}` to `vscode.bundlevars.json` via `workspace.fs.writeFile`, then asserts the Bundle Variables tree shows `"new value"` — but within the **10 s** window it still read the `"dev"` default, because the extension's FileSystemWatcher hadn't refreshed the tree yet. (Observed on the Linux shard; mechanism is OS-independent timing.)

## What
- Widen the assertion `waitUntil` **10 s → 30 s**.
- On each still-stale poll, re-issue the **idempotent** `workspace.fs.writeFile` to re-trigger the watcher (content is deterministic, so rewriting is safe).
- The `assertVariableValue` ground-truth assertion is unchanged.

## Verification
`tsc --noEmit` exit 0; eslint + prettier clean. e2e can't run locally — the `bundle_variables` shard (incl. Linux) is exercised across ≥4 isolated CI runs.

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
@@ -199,14 +199,33 @@ describe("Bundle Variables", async function () {
 
         await browser.waitUntil(
             async () => {
-                await assertVariableValue(section, "varWithDefault", {
-                    value: "new value",
-                    defaultValue: "default",
-                });
-                return true;
+                try {
+                    await assertVariableValue(section, "varWithDefault", {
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `338a78f2` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29917435862)

misha-db (approved):

PR #2034, opened 2026-07-22
test(e2e): gate bundle_init wizard keystrokes on tab readiness

rugpanov:
## Why
`bundle_init :: should initialize new project` flakes on the Windows shard with **"Can't complete cli bundle init wizard"** (its two follow-on tests then fail as cascades). The test drives an interactive **editor-hosted terminal** wizard by firing `browser.keys()` with fixed `sleep(1000)` gaps; on the slow Windows shard the keystrokes race the terminal prompt's readiness and desync the wizard.

## What
- Before typing, gate on the **"Databricks Project Init" tab being the active editor tab** (bounded `waitUntil`) instead of a bare `sleep`.
- Widen the keystroke settle sleeps **1 s → 3 s** and the completion `waitUntil` **20 s → 40 s**.

**Caveat:** the terminal buffer text isn't reliably readable via the wdio API, so the keystroke steps remain timing-based. This **reduces rather than provably eliminates** the race. If it still recurs, a non-interactive init path would be the follow-up.

## Verification
`tsc --noEmit` exit 0; eslint + prettier clean. e2e can't run locally — the `bundle_init` Windows shard is exercised across ≥4 isolated CI runs (more if any residual failure appears).

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/bundle_init.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/bundle_init.e2e.ts
@@ -76,15 +76,33 @@ describe("Bundle Init", async function () {
         const initTab = await getTabByTitle(title);
         assert(initTab, "Can't find a tab for project-init terminal wizard");
         await initTab.select();
-        await sleep(1000);
+
+        // The init wizard runs inside an editor-hosted terminal. Keystrokes sent
+        // before the terminal prompt is ready land in the wrong place and desync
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `5efb7296` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/29917424646)

misha-db (approved):

PR #2038, opened 2026-07-23
[2/?] feat(python-setup): map setup-local error codes to actionable messages

rugpanov:
> **Stacked on #2036** — review that first. This PR's base is `rugpanov/vpex-cli-result-contract`, so the diff here is only the error-message mapping.

## What

Second task of **M4 — Productionize the VS Code extension**.

Adds `controllers/pythonSetupErrorMessage.ts` — the first code in the feature's `controllers/` layer.

- **`pythonSetupErrorMessage(result)`** maps each `PythonSetupErrorCode` (`E_NO_TARGET`, `E_ENV_UNSUPPORTED`, `E_MANAGER_UNSUPPORTED`, `E_FETCH`, `E_PROVISION`, `E_VALIDATE`, …) to a concise, actionable message.
- Wording is friendlier/shorter than the CLI's own error text (still available via "Show Logs") but stays **consistent** with it — `E_ENV_UNSUPPORTED` points at the latest LTS target and includes the env key; `E_MANAGER_UNSUPPORTED` names uv.
- Always appends a **`diskMutated`-aware suffix**: *"No changes were made to your project."* when false, or a pointer to the `pyproject.toml.bak` backup when true.
- Falls back to a generic message for an unknown code or a result with no error object.

## Why

When `setup-local` fails, the extension must show a clear, actionable message rather than a raw CLI dump — and tell the user whether their project was modified. This is the layer the setup orchestrator (later task) calls to render failures.

## Backward compatibility

No user-facing change. New file only, under `src/python-setup/controllers/`; nothing existing is imported or modified. Not wired anywhere yet.

## Verification

- `yarn test:unit` → **299 passing, 0 failing** (12 new assertions, two driven by the real CLI golden fixtures).
- `yarn test:lint` → clean.

This pull request and its description were written by Isaac.

PR #2039, opened 2026-07-23
[3/?] feat(python-setup): add setup-local CLI client (argv, path, spawn, JSON)

rugpanov:
> **Stacked on #2038** — review that first. This PR's base is `rugpanov/python-setup-error-messages`, so the diff here is only the CLI-client code.

## What

Third task of **M4 — Productionize the VS Code extension**.

The argv/path helpers and the process gateway for driving `databricks environments setup-local`.

- **`utils/setupLocalArgs.ts`** (pure):
  - `buildSetupLocalArgs(invocation)` assembles the argv — target flag (`--cluster-id` or `--serverless-version`) → `--constraints-only` (constraints-only mode) → `--profile` → the hidden `--constraint-source-url` → `--output json`. Serverless version is passed as a bare number (e.g. `5`), per the shipped contract.
  - `resolveCliPath({override, bundled})` prefers a non-empty custom CLI path over the bundled binary. Pure — the config value is read by the caller (the `databricks.experiments.cliPathOverride` setting lands in a later task).
- **`utils/venvInterpreterPath.ts`** (pure): OS-aware interpreter path — `Scripts\python.exe` on Windows, `bin/python` elsewhere; platform injectable.
- **`gateways/PythonSetupCliClient.ts`**: spawns the CLI (injectable spawn seam), captures **stdout → `parsePythonSetupResult`**, forwards **stderr → an `onLog` channel** (for "Show Logs"), resolves with the parsed result on **both success and failure exits** (the caller branches on `result.ok`/`result.error`, not the exit code), rejects on spawn error or unparseable stdout (with captured stderr appended), and terminates the child on cancellation.

## Design note: no live progress under `--output json`

Reading the CLI source (`cmd/environments/output.go`, `libs/localenv/uv.go`) confirmed that in JSON mode the CLI emits **only the final JSON object** on stdout — the per-phase text loop is skipped, and `uv sync` output is buffered (`process.Background`), not streamed. So there is **no live per-phase progress stream** to narrate. Consumers show an indeterminate spinner and read the per-phase outcomes from `result.phases` after the process exits. (The earlier plan's "narrate `=== Phase N ===` stderr lines" idea came from the old POC and does not hold; no phase-narrator is built.)

## Backward compatibility

No user-facing change. New files under `src/python-setup/{utils,gateways}/`; nothing existing is imported or modified. Not wired anywhere yet.

## Verification

- `yarn test:unit` → **325 passing, 0 failing** (argv, path, venv-path, and gateway spawn/parse/cancel cases, incl. two driven by the real CLI golden fixtures).
- `yarn test:lint` → clean.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/gateways/PythonSetupCliClient.ts
+++ b/packages/databricks-vscode/src/python-setup/gateways/PythonSetupCliClient.ts
@@ -1,0 +1,263 @@
+import {
+    spawn as nodeSpawn,
+    ChildProcessWithoutNullStreams,
+} from "node:child_process";
+import {StringDecoder} from "node:string_decoder";
+import {
+    buildSetupLocalArgs,
+    SetupLocalInvocation,
+} from "../utils/setupLocalArgs";
+import {
+    parsePythonSetupResult,
+    PythonSetupResult,
+} from "../models/PythonSetupResult";
+
+/**
+ * Minimal cancellation signal, structurally compatible with `vscode.Cancellation
+ * Token`. Declared locally so this gateway carries no `vscode` import and stays
+ * unit-testable without the extension host.
+ */
+export interface CancellationLike {
+    readonly isCancellationRequested: boolean;
+    onCancellationRequested(listener: () => void): {dispose(): void};
+}
+
+/**
+ * Injectable spawn seam. The real implementation is Node's `child_process.
+ * spawn`; tests pass a fake that emits scripted stdout/stderr.
+ */
+export type SpawnFn = (
+    command: string,
+    args: string[],
+    options: {cwd: string; env: NodeJS.ProcessEnv; detached: boolean}
+) => ChildProcessWithoutNullStreams;
+
+/**
+ * Injectable process-tree terminator. `setup-local` spawns `uv sync` as a
+ * grandchild, and killing only the direct `databricks` child orphans it: on
+ * Windows a plain `SIGTERM` doesn't even reach the CLI, and on POSIX a signal
+ * to the direct child is not propagated to grandchildren. The default kills the
+ * whole tree on both — `taskkill /T /F` on Windows, and (because the child is
+ * spawned {@link https://nodejs.org/api/child_process.html detached}, i.e. its
+ * own process-group leader) a `SIGTERM` to the negated pid on POSIX, which
+ * signals the entire group. Injectable so cancellation is assertable in tests
+ * without spawning real processes.
+ */
+export type TerminateFn = (child: ChildProcessWithoutNullStreams) => void;
+
+/**
+ * OS primitives used by {@link terminateProcessTree}, injectable so the
+ * termination logic can be unit-tested without touching real processes.
+ */
+export interface TerminatePrimitives {
+    platform: NodeJS.Platform;
+    /** Send `signal` to `pid` (a negative pid targets the process group). */
+    kill(pid: number, signal: NodeJS.Signals): void;
+    /** Fire-and-forget helper spawn (e.g. `taskkill`). */
+    spawnHelper(command: string, args: string[]): void;
+}
+
+const realTerminatePrimitives: TerminatePrimitives = {
+    platform: process.platform,
+    kill: (pid, signal) => process.kill(pid, signal),
+    spawnHelper: (command, args) => {
+        // Swallow the async spawn error (e.g. taskkill not on PATH) so it can't
+        // become an uncaught exception in the extension host during cancel.
+        nodeSpawn(command, args).on("error", () => {});
+    },
+};
+
+/**
+ * Kill the whole `setup-local` process tree, injectable primitives and all. On
+ * Windows a plain `SIGTERM` doesn't reach the CLI, so force-kill the tree with
+ * `taskkill /T /F`; on POSIX the child is a process-group leader (spawned
+ * detached), so signalling the negated pid tears down the whole group —
+ * including the `uv` grandchild a direct-child SIGTERM would orphan — with a
+ * fallback to a direct kill if the group is already gone.
+ */
+export function terminateProcessTree(
+    child: ChildProcessWithoutNullStreams,
+    prims: TerminatePrimitives = realTerminatePrimitives
+): void {
+    const pid = child.pid;
+    if (prims.platform === "win32") {
+        if (pid) {
+            prims.spawnHelper("taskkill", ["/pid", String(pid), "/T", "/F"]);
+            return;
+        }
+    } else if (pid) {
+        try {
+            prims.kill(-pid, "SIGTERM");
+            return;
+        } catch {
+            // Group already gone / not a leader — fall through to a direct kill.
+        }
+    }
+    child.kill("SIGTERM");
+}
+
+const defaultTerminate: TerminateFn = (child) => terminateProcessTree(child);
+
+/**
+ * Rejection raised when a {@link PythonSetupCliClient.run} is aborted via its
+ * cancellation token, so callers can distinguish a user cancellation from a
+ * genuine CLI failure (which surfaces as a `PythonSetupParseError` or the
+ * spawn error).
+ */
+export class PythonSetupCancelledError extends Error {
+    constructor() {
+        super("Python setup was cancelled");
+        this.name = "PythonSetupCancelledError";
+    }
+}
+
+export interface RunOptions {
+    /** Working directory the CLI runs in (the project root). */
+    cwd: string;
+    /** Receives raw stderr chunks for the output/"Show Logs" channel. */
+    onLog?: (chunk: string) => void;
+    /** When cancelled, the child process is terminated. */
+    token?: CancellationLike;
+}
+
+/**
+ * Gateway to the `databricks environments setup-local` CLI command.
+ *
+ * Runs the CLI with `--output json`, captures stdout (the single structured
+ * {@link PythonSetupResult}) and stderr (raw logs), and resolves with the parsed
+ * result on both success and failure exits — the caller branches on `result.ok`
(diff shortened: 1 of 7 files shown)


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `2a9b1a41`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30003240195)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `01ca7aca`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30005298916)

rugpanov:
🤖 Integration tests triggered for `b7efe481` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30009263215)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `7416432a`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30009654397)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `3ba5c9ba`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30017637145)

rugpanov:
🤖 Integration tests triggered for `293265d5` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30020561263)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `794313c2`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30020980303)

packages/databricks-vscode/src/python-setup/gateways/PythonSetupCliClient.ts
```diff
@@ -0,0 +1,263 @@
...
+ * pure helpers, and the spawn function is injected, so it is fully unit-testable.
+ */
+export class PythonSetupCliClient {
+    constructor(
+        private readonly resolvePath: () => string,
+        private readonly spawnFn: SpawnFn = nodeSpawn as unknown as SpawnFn,
+        private readonly terminateFn: TerminateFn = defaultTerminate
+    ) {}
+
+    run(
```
misha-db, line 148: We already have functions allowing running cli commands and reading results, like: 

- execFile(file, args, options?, cancellationToken?) — runs a CLI binary once and returns {stdout, stderr}. Handles Windows escaping (cmd.exe /d /c ...) via getEscapedCommandAndAgrs. This is what you want for one-shot commands that produce output.
- cancellableExecFile(...) — same, but the raw version without the platform escaping (used internally by execFile).
- waitForProcess(childProcess, onStdOut?, onStdError?) — for long-running/streamed processes you spawn yourself, collecting stdout/stderr with optional live callbacks.


Also there's a CliWrapper class.

Can they be reused/enhanced instead of coming with new implementation?

rugpanov: Good question — there's real overlap, but the three helpers can't be reused as-is here, for reasons specific to the `--output json` contract:

1. **Resolve on both exit codes.** `waitForProcess` (and `execFile` via `promisify`) reject on a non-zero exit and drop stdout. But `setup-local --output json` emits a structured result on stdout on *both* success and failure, and the caller branches on `result.ok`/`result.error`, not the exit code. Reusing either helper would discard the structured error result on the failure path.

2. **UTF-8 chunk-straddle.** `waitForProcess` does `data.toString()` per chunk; a multi-byte UTF-8 sequence split across two `data` events corrupts to U+FFFD, which would break JSON parsing. The client accumulates `Buffer`s and decodes stdout once (and uses a `StringDecoder` for streamed stderr).

3. **POSIX process-group kill.** `runBundleCommand`'s cancel is `abortController.abort()`, which signals only the direct child and orphans the `uv sync` grandchild `setup-local` spawns. The client spawns `detached` and SIGTERMs the negated pid to tear down the whole group. `runBundleCommand` doesn't spawn detached, so it can't share this path without a behavior change to bundle cancel.

There's also a deliberate constraint: the gateway carries no `vscode` import (uses a local `CancellationLike`) so it's unit-testable without the extension host, whereas `CliWrapper` is coupled to `window`/`commands`/`ExtensionContext`.

Where you're right: the Windows `taskkill /T /F` tree-kill is duplicated between `terminateProcessTree` and `runBundleCommand`. The clean unification would be a vscode-free low-level runner (buffered decode + group-aware kill, resolves regardless of exit code) that both `runBundleCommand` and this gateway sit on top of. That's the right end state, but it rewrites the plumbing of `runBundleCommand`'s 5 bundle-op callers — larger blast radius than fits this isolated-client PR. I'd rather do it as a dedicated follow-up refactor than widen this stacked PR. WDYT?

packages/databricks-vscode/src/python-setup/gateways/PythonSetupCliClient.ts
```diff
@@ -0,0 +1,263 @@
...
+ */
+export type SpawnFn = (
+    command: string,
+    args: string[],
+    options: {cwd: string; env: NodeJS.ProcessEnv; detached: boolean}
+) => ChildProcessWithoutNullStreams;
+
+/**
+ * Injectable process-tree terminator. `setup-local` spawns `uv sync` as a
+ * grandchild, and killing only the direct `databricks` child orphans it: on
```
rclarey, line 37: Can you clarify where exactly the kill disconnect is here? I can't quite follow.

Is it
- when `databricks` is killed it doesn't kill the `uv` process it spawns
- when `uv` is killed it doesn't kill some processes it spawns
- something else?

misha-db (approved):

packages/databricks-vscode/src/python-setup/gateways/PythonSetupCliClient.ts
rugpanov: It's the first one: when we kill only the direct `databricks` child, the `uv sync` process it spawned as a grandchild keeps running (orphaned). On POSIX a signal isn't propagated from a parent to its children — `child.kill()` reaches only the `databricks` PID, not `uv`. It's transitive too (anything `uv` spawns would also survive), which is why the fix is a whole-tree/group kill rather than walking children manually.

The two branches exist because the symptom differs per-OS:
- **POSIX:** `child.kill()` does kill `databricks`, just not `uv`. So we spawn `databricks` `detached` → it becomes its own process-group leader and `uv` inherits the group; `SIGTERM` to the negated pid (`-pid`) signals the entire group, reaching `databricks` + `uv` + uv's children at once.
- **Windows:** a plain `SIGTERM` doesn't reliably terminate even the `databricks` CLI itself (Node's Windows kill is unreliable for console processes), so `taskkill /T /F` force-kills the whole tree.

packages/databricks-vscode/src/python-setup/gateways/PythonSetupCliClient.ts
rclarey: It seems the conclusion is exactly what Misha is saying. IMO it's better to improve `CliWrapper` with what you need up front rather than create a new implementation which may/may not be unified later on

packages/databricks-vscode/src/python-setup/gateways/PythonSetupCliClient.ts
rclarey: > Windows: a plain SIGTERM doesn't reliably terminate even the databricks CLI itself (Node's Windows kill is unreliable for console processes)

ah, makes sense 👍 

> POSIX: child.kill() does kill databricks, just not uv

Yes that's the current behaviour, but it's completely controllable on our side. The CLI should register a SIGTERM handler and terminate the `uv` subprocesses it owns. This means we avoid a lot of extra work here working around a badly behaving CLI process

rugpanov: Confirmed the CLI doesn't do this today: `main.go` runs `root.Execute` with `context.Background()` (no `signal.NotifyContext`), and there's no signal handler on the `environments`/`setup-local` path and no `Setpgid`/`SysProcAttr` anywhere in the repo. So on SIGTERM the CLI exits immediately with no cleanup and `uv` is orphaned. Agree the right long-term fix is CLI-side — it should reap the `uv` subprocesses it owns — and I'll file that upstream.

That said, I'd keep this logic here even once the CLI handles it, for two reasons:

1. **Windows isn't a CLI-behavior problem — a handler can't help there.** Node's `child.kill()` on Windows is `TerminateProcess` on the single PID; SIGTERM/SIGINT/SIGKILL all collapse to one uncatchable hard kill. No catchable signal is ever delivered, so a CLI-side SIGTERM handler would never run. Even a perfectly-behaved CLI still orphans `uv` on Windows unless we `taskkill /T`. This branch stays regardless of the CLI.

2. **Version coupling during the experimental window.** `setup-local` isn't in the bundled CLI yet and we honor `cliPathOverride`, so we can't assume the fix is present in whatever binary runs, and we don't gate on a minimum CLI version. Dropping the POSIX group-kill on the promise of a fix would regress anyone on an unfixed binary.

So: file the CLI cleanup upstream, and once it ships the POSIX `detached`/`kill(-pid)` branch becomes redundant and can go — but the Windows tree-kill is unconditional, and version coupling means we shouldn't lean on the CLI fix yet. Net: keep it now, revisit the POSIX half as a simplification later.

rclarey: Agreed offline to do the CLI cleanup upstream, but still leave this process tree termination as a backup safeguard for orphaned processes

rclarey (approved):

PR #2032, opened 2026-07-22
test(e2e): raise waitForWorkflowWebview timeout to 60s for Windows shard

rugpanov:
## Why
`run_notebooks_ipynb :: should run a notebook.ipynb file as a workflow` flakes on the Windows e2e shard with **"Webview did not open"**. The `waitForWorkflowWebview` poll allowed only **5 s** — not enough time on Windows to submit the workflow run and materialize the "Databricks Job Run" webview panel.

## What
Raise the poll timeout **5 s → 60 s** (interval 1 s → 2 s) in the shared `waitForWorkflowWebview` helper. Its only caller is the ipynb workflow test.

Test-only change; no product code touched.

## Verification
`tsc --noEmit` exit 0; eslint + prettier clean. e2e can't run locally — the `run_notebooks_ipynb` Windows shard is exercised across ≥4 isolated CI runs.

This pull request and its description were written by Isaac.

PR #2042, opened 2026-07-23
test(e2e): harden bundle_variables revert against watcher refresh race

rugpanov:
## Why
`bundle_variables :: should revert overrides` flakes on the Windows shard — the **same FileSystemWatcher refresh race** that #2033 fixed for the sibling `should override default variable`. The test clicks **"Reset bundle variables to default values"** and expects the tree to revert `varWithDefault` to `"dev"`, but within the **10 s** window it still reads the stale `"new value"` because the watcher hadn't refreshed the tree yet.

Seen post-merge on `main` (all of #2032/#2033/#2034 already landed): [run 30004574431](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30004574431) — 31/32 green, the only failure being this test (`+ 'new value' - 'dev'`). #2033 hardened only the override test; this applies the identical pattern to its revert sibling.

## What
- Widen the assertion `waitUntil` **10 s → 30 s**.
- On each still-stale poll, **re-click the (re-fetched) "Reset bundle variables to default values" action** to re-trigger the watcher. Re-fetching guards against a stale element across tree refreshes; resetting when already-default is a harmless no-op.
- The `assertVariableValue` ground-truth assertion is unchanged.

## Verification
`tsc --noEmit` exit 0; eslint + prettier clean. e2e can't run locally — the `bundle_variables` shard (Linux + Windows) is exercised across repeated isolated CI runs.

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/bundle_variables.e2e.ts
@@ -250,14 +250,30 @@ describe("Bundle Variables", async function () {
 
         await browser.waitUntil(
             async () => {
-                await assertVariableValue(section, "varWithDefault", {
-                    value: "dev",
-                    defaultValue: "default",
-                });
-                return true;
+                try {
+                    await assertVariableValue(section, "varWithDefault", {
+                        value: "dev",
+                        defaultValue: "default",
+                    });
+                    return true;
+                } catch {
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `0ae46f31`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30008891761)

misha-db (approved):

PR #2043, opened 2026-07-23
WIP: Code conventions for new code matching existing style

misha-db:
## Changes

Describes existing code style and naming conventions. So new code can follow the same principles

## Tests

Manually

### DIFF
--- a/CODE_CONVENTIONS.md
+++ b/CODE_CONVENTIONS.md
@@ -57,6 +57,12 @@ matches the responsibility - don't invent new ones.
   handles refresh/polling.
 - **`XCommands`** — thin command handlers that call into the model/manager.
 
+**Keep `window.*` UI out of `Model` / `Manager` / `Loader`.** These surface
+data/results and fire events; let `Commands` / `Component` do the
+`window.show*` rendering. Fusing `window.showInformationMessage` /
+`showErrorMessage` with SDK or `child_process` calls in one class is exactly what
+makes it hard to unit-test (see section 8).
+
 ---
 
 ## 3. Structuring a new feature
@@ -67,14 +73,27 @@ matches the responsibility - don't invent new ones.
 2. **Add classes by role** using the suffixes described in section 2 above:
    `MyFeatureModel.ts`, `MyFeatureManager.ts`, `MyFeatureCommands.ts`, plus any
    providers/wizards the feature needs.
-3. **Only add subfolders when a feature grows large.** Precedent: a `models/`
-   subfolder for a cluster of state classes (as in `configuration/models/`,
-   `bundle/models/`), or a domain subfolder like `bundle/run/`. Don't pre-split a
-   small feature.
-4. **Put UI-heavy code under `ui/`.** Tree data providers, detail panels,
+3. **Group by role-suffix once a feature grows large.** Keep small features flat.
+   Once a feature exceeds ~6–8 files, group files into role-suffix subfolders
+   (`models/`, `managers/`, `commands/`, `providers/`…) — **but only for roles that
+   have 2+ files.** Singleton roles stay at the feature root: even `bundle/`, the
+   biggest feature, has just one Wizard, one Watcher, and one Provider, so a strict
+   folder-per-suffix would scatter logic into one-file folders. The folder name is
+   just the plural of the suffix the class already carries — **no class renames.**
+   Co-locate each role's tests inside its folder. Precedent: `configuration/models/`,
+   `bundle/models/`, `bundle/run/`. Don't pre-split a small feature.
+4. **Put UI-heavy code under `ui/`.** (unresolved) Tree data providers, detail panels,
    wizards, and configuration-view components live in `ui/<feature>/` (e.g.
    `ui/unity-catalog/`). Feature _logic_ (models, managers) stays in the feature
    folder; the tree that renders it lives under `ui/`.
+
+    > _Trade-off:_ a global `ui/` gives one index of all views, but
+    > splits a feature across trees — `bundle` today spans `bundle/`,
+    > `ui/bundle-resource-explorer/`, and `ui/bundle-variables/` (43 files, one
+    > concept, three places). The inverse — co-locating views under the feature
+    > (`bundle/ui/…`) so one feature is one subtree — is worth considering. Follow the
+    > global-`ui/` convention for now; flagged for discussion.
+
 5. **Never touch VS Code globals directly for state/config/context** — go through
    the adapter layer in `vscode-objs/` (see section 5).
 6. **Wire it up in `extension.ts`** (see section 6): construct the objects, inject
@@ -176,6 +195,19 @@ If you need a new piece of global state or a new setting accessor, **add it to t
 relevant `vscode-objs/` class** rather than sprinkling `workspace.getConfiguration`
 or `context.globalState` calls across the feature.
 
+### The SDK is a cross-cutting dependency too
+
+The Databricks SDK (`@databricks/sdk-experimental`) is the biggest unguarded
+cross-cutting dependency in the codebase — dozens of files import it, and many
+reach the `WorkspaceClient` / `apiClient` directly.
+
+- **Reach the workspace client through the connection seam** (`ConnectionManager`),
+  not by constructing your own client in a feature.
+- **Never import through deep `/dist/...` paths** (`.../dist/apis/…`,
+  `.../dist/retries/…`). They're not a stable entry point — import from the package
+  root. Keeping SDK access behind one seam is also what turns a future SDK migration
+  into a bounded change instead of a repo-wide edit.
+
 ---
 
 ## 6. Wiring in `extension.ts`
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests triggered for `550887e6` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30016804611)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `0244fc9c`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30017226107)

CODE_CONVENTIONS.md
```diff
@@ -0,0 +1,254 @@
...
+### Logging & telemetry
+
+- Log through the named loggers in `logger/` (`Loggers.Extension`, …) — **never
+  `console.log`** (`no-console` is an ESLint error outside tests).
+- Define new telemetry events in `telemetry/constants.ts`. User-facing commands are
+  instrumented automatically by the `telemetry.registerCommand` wrapper (section 6).
+
+---
+
+## 5. Go through the adapter layer (`vscode-objs/`)
```
rugpanov, line 182: §5 covers state/config/context via `vscode-objs/`, but the **SDK is the biggest unguarded cross-cutting dependency** and isn't mentioned: 68 non-test files import `@databricks/sdk-experimental` (20 touch `WorkspaceClient`/`apiClient` directly), many through deep `/dist/apis/...` / `/dist/retries/...` paths. Worth a rule here: reach the workspace client through the connection seam, and avoid deep `/dist/` imports (they're not a stable entry point). This is also what would make the planned SDK migration a bounded change rather than a 68-file edit.

CODE_CONVENTIONS.md
```diff
@@ -0,0 +1,254 @@
...
+| Unit test                                                   | `<Name>.test.ts`, co-located                                  | `ClusterModel.test.ts`                        |
+
+**Rule of thumb for file casing:** does the file's _primary_ export have a name
+starting with a capital (a class/type)? Name the file to match it exactly
+(`ClusterModel.ts` exports `ClusterModel`). Is it a bag of functions? Use camelCase
+(`fileUtils.ts`).
+
+---
+
+## 2. Choose the right class role, then name for it
```
rugpanov, line 31: Could we add one line to the triad: **keep `window.*` UI out of `Model`/`Manager`/`Loader`** — surface data/results and let `Commands`/`Component` render it? Audit found 9 classes fusing `window.show*` with SDK/`child_process` in one file (e.g. `ConnectionManager`, `AuthProvider`, `AzureCliCheck`), which is exactly what makes them hard to unit-test — reinforces §8.

CODE_CONVENTIONS.md
```diff
@@ -0,0 +1,254 @@
...
+**The core triad.** Most features are built from three cooperating classes:
+
+- **`XModel`** — owns the data and fires events when it changes. No VS Code UI.
+- **`XManager`** — orchestrates: constructs collaborators, reacts to events,
+  handles refresh/polling.
+- **`XCommands`** — thin command handlers that call into the model/manager.
+
+---
+
+## 3. Structuring a new feature
```
rugpanov, line 68: The "keep it flat" rule leaves medium/large features as an undifferentiated pile — business logic, presentation, and I/O side-by-side with tests (`run/` 14 files, `bundle/` 12, `language/` 11 at one level). Suggestion: once a feature exceeds ~6–8 files, **group by role-suffix into subfolders** (`models/`, `managers/`, `commands/`, `providers/`…), **but only for roles with 2+ files** — singleton roles stay at the root (even `bundle/`, the biggest feature, has just 1 Wizard / 1 Watcher / 1 Provider, so a folder-per-suffix would scatter into one-file folders). Tests co-locate inside each role folder, which also unburies logic from test files. No class renames — the folder is just the plural of the suffix the class already carries.

rugpanov (commented):
Solid, accurate onboarding doc — the suffix taxonomy, DI, event/disposal idioms, `vscode-objs/` seam, decorators, and test suffixes all match the codebase. A few additions from a recent structural audit that fill gaps here, all consistent with the "match the current setup" framing (no new abstractions):

CODE_CONVENTIONS.md
```diff
@@ -0,0 +1,254 @@
...
+   (lowercase, hyphenated if multi-word): `src/my-feature/`. Keep it **flat** —
+   put files directly in the folder.
+2. **Add classes by role** using the suffixes described in section 2 above:
+   `MyFeatureModel.ts`, `MyFeatureManager.ts`, `MyFeatureCommands.ts`, plus any
+   providers/wizards the feature needs.
+3. **Only add subfolders when a feature grows large.** Precedent: a `models/`
+   subfolder for a cluster of state classes (as in `configuration/models/`,
+   `bundle/models/`), or a domain subfolder like `bundle/run/`. Don't pre-split a
+   small feature.
+4. **Put UI-heavy code under `ui/`.** Tree data providers, detail panels,
```
rugpanov: Discussion point (not a blocker): putting a feature's UI under a global `ui/<feature>/` while its logic stays in the feature folder splits one feature across trees. `bundle` today spans **three** homes — `bundle/` (21 files), `ui/bundle-resource-explorer/` (18), `ui/bundle-variables/` (4) — 43 files, one concept, three places to jump between when editing it. Worth considering the inverse: co-locate a feature's views under the feature folder (`bundle/ui/…`) so one feature = one subtree. Trade-off: the global `ui/` gives a single index of all views, but in practice you navigate "this feature," not "all views." Flagging for discussion since it reverses a deliberate convention.

CODE_CONVENTIONS.md
```diff
@@ -0,0 +1,254 @@
...
+
+---
+
+## 7. Imports & exports
+
+- **Double quotes** for imports (Prettier-enforced).
+- Order: external packages first (`vscode`, `@databricks/sdk-experimental`, …),
+  then local relative imports.
+- Use **`import type { … }`** for type-only imports and to break dependency cycles.
+- For utility folders, follow the **namespace-barrel** pattern —
```
rugpanov, line 245: Worth noting this rule is currently inverted in practice: **10 of 12 `index.ts` use `export *`** (only 2 are selective). Blanket `export *` barrels actively hurt navigation — "go to definition" and grep-for-usage land on the barrel rather than the real file, adding an indirection hop with no encapsulation benefit. Suggest the doc either (a) hold the line on selective re-exports and note the existing barrels should migrate, or (b) drop under-used barrels entirely. Right now §7 says one thing and most barrels do the opposite.

rugpanov (commented):
Two more, focused on navigation (finding where a feature lives / where to add code):

CODE_CONVENTIONS.md
```diff
@@ -0,0 +1,254 @@
...
+- `{a: 1}` — **no** space inside braces (`bracketSpacing: false`).
+- `(x) => …` — always parenthesize arrow params.
+- Trailing commas where ES5 allows.
+- Prefer `===` / `!==` and always use curly braces (`curly`).
+
+---
+
+## Issues
+
+- extension.ts is becoming too big
```
rugpanov, line 291: Can we add another entry here? A large feature folder becomes hard to navigate when many files with different role-suffixes sit flat at one level — business logic, presentation, and I/O interleaved (and doubled by co-located tests). Concretely today: `run/` 14 files, `bundle/` 12, `language/` 11, `ui/unity-catalog/` 10 at the top level, across 17 distinct suffixes repo-wide. Proposed bullet:

- large feature folders mix many role-suffixes (logic / presentation / I/O) flat at one level, which is hard to navigate — need a grouping convention (see the §3 discussion)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `b5f727b5` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30102474661)

CODE_CONVENTIONS.md
```diff
@@ -0,0 +1,296 @@
...
+| File exporting a **class**                                  | `PascalCase.ts`                                               | `ClusterModel.ts`                             |
+| File exporting **functions** (util / helper / registration) | `camelCase.ts`                                                | `fileUtils.ts`, `registerDetailPanel.ts`      |
+| Class / interface / type / enum                             | `PascalCase`                                                  | `ConnectionManager`, `RunState`               |
+| Method / local variable / function                          | `camelCase`                                                   | `refresh()`, `activeCluster`                  |
+| Module-level constant                                       | `UPPER_SNAKE_CASE`                                            | `SCHEME`, `PROD_APP_INSIGHTS_KEY`             |
+| Private field                                               | `camelCase`, `_`-prefix only for event-emitter backing fields | `private disposables`, `private _onDidChange` |
```
rugpanov: This rule is contradicted by the codebase and I think it should be fixed before merge (the risk with a style guide is that a wrong *rule* gets applied — someone cites it to reject `private _state`, or "fixes" existing code to match).

"`_`-prefix only for event-emitter backing fields" — but there are ~37 non-emitter private fields using the `_` prefix today, e.g. `_state` / `_serverlessEnabled` / `_workspaceClient` / `_clusterManager` in `ConnectionManager`, `_clusters` in `ClusterModel`, `_target` / `_authProvider` in `ConfigModel`, `_activeProjectUri` in `WorkspaceFolderManager`, `_value` / `_dirty` in the state-cache base. So the guide as written fails its own reference classes.

The actual pattern is closer to: `_`-prefix a private backing field that has a public getter/accessor of the same name (`_state` → `get state()`, `_onDidChange` → `onDidChange`). Suggested cell:

> `camelCase`; `_`-prefix a private field backing a public getter/accessor of the same name

with the example column becoming e.g. `private disposables`, `private _state` (→ `get state()`).

rugpanov:
🤖 Integration tests triggered for `6d95e0be` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30274655158)

rugpanov (approved):
looks strong!

rugpanov:
🤖 Integration tests triggered for `939ce893` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30275531260)

PR #2044, opened 2026-07-24
[4/?] feat(python-setup): package-manager gating & live detection

rugpanov:
## Why

The upcoming uv-native "set up Python environment" feature must be offered only
when it fits the project, and never alongside the legacy pip checklist. This PR
adds the two decision pieces — a live package-manager classification of the
active project, and the pure gate predicate the config view will consult — plus
the refactor that lets the detector and the existing detection telemetry share
one signal-collection layer instead of drifting.

This is additive: nothing here is wired into the extension yet (that lands with
the orchestrator + config-view dispatch), so there is no user-facing behavior
change. The whole feature stays behind a flag; this PR only supplies the
building blocks.

## What

- **Extract the shared collector** — move the private disk+interpreter signal
  collection out of `PackageManagerTelemetry` into
  `language/packageManagerSignals.ts` (`collectPackageManagerSignals`), with an
  injectable debug-log sink so callers without a logger (tests) can omit it.
  `PackageManagerTelemetry` now calls it; behavior is unchanged and its existing
  tests pass untouched.
- **Live detector** — `python-setup/utils/PythonSetupManagerDetector.ts`: a thin
  async shell over the pure `detectPackageManagers`, driven by an injected
  signal collector, returning `{primary, managers}`. Any collection failure
  degrades to `{unknown, []}` so detection never disrupts the flow.
- **Gate predicate** — `python-setup/utils/pythonSetupGate.ts`
  (`shouldShowPythonSetup`): show only when the flag is on, the primary manager
  is `uv` or `unknown`, and no competing manager (pip/poetry/conda) fired. This
  is the single point that keeps uv-native setup and the legacy pip checklist
  mutually exclusive.

## Verification

- `yarn --cwd packages/databricks-vscode test:unit` — 319 passing (14 new:
  5 detector, 9 gate; the unchanged `PackageManagerTelemetry` suite included).
- `yarn --cwd packages/databricks-vscode test:lint` — clean.
- `yarn --cwd packages/databricks-vscode build` — clean.

### DIFF
--- a/packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
+++ b/packages/databricks-vscode/src/language/PackageManagerTelemetry.ts
@@ -1,5 +1,3 @@
-import fs from "node:fs";
-import path from "node:path";
 import {NamedLogger} from "@databricks/sdk-experimental/dist/logging";
 import {Loggers} from "../logger";
 import {Telemetry} from "../telemetry";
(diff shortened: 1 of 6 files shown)


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `71122d7e`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30079473633)

misha-db (approved):

PR #2045, opened 2026-07-24
[5/?] feat(python-setup): cliPathOverride setting & opt-in feature flag

rugpanov:
## Why

The uv-native "set up Python environment" feature must ship hidden: its CLI
command (`environments setup-local`) is available only in custom CLI builds for
now, so the whole flow stays behind an opt-in flag until the command is
generally available. This PR wires the two settings and the feature-flag
registration that gate it.

This is additive and inert: nothing consumes the flag or the override yet (the
config-view dispatch + orchestrator wiring is a later ticket), so there is no
user-facing behavior change. The feature is registered as **disabled by
default**.

## What

- **Single source of truth** — `PYTHON_SETUP_FEATURE_ID = "environment.pythonSetup"`
  exported from `FeatureManager`, and the `FeatureId` union extended. This
  matters because `FeatureManager` unlocks a disabled feature only when the
  opt-in string equals the feature id **verbatim** — so the disabled-list entry,
  the `package.json` opt-in enum value, and any future flag reader must all use
  this one constant. (The original plan used two different strings, `setup.vpex`
  vs `environment.vpex`, which would have made the feature impossible to unlock.)
- **`extension.ts`** — register `PYTHON_SETUP_FEATURE_ID` in the `FeatureManager`
  `disabledFeatures` list, hiding the feature by default.
- **`package.json`** — add `"environment.pythonSetup"` to the
  `databricks.experiments.optInto` enum (+ description), and add the temporary
  `databricks.experiments.cliPathOverride` string setting (documented as
  temporary, to be removed at GA).
- **`WorkspaceConfigs.ts`** — add the `pythonSetupCliPathOverride` getter (empty
  string ⇒ use the bundled CLI). Deliberately **no** `pythonSetupEnabled` getter
  here: it would make `WorkspaceConfigs` import `FeatureManager`, which already
  imports `workspaceConfigs` — a circular dependency. The one-line opt-in check
  belongs at the `extension.ts` wiring site, and nothing in this ticket consumes
  it yet (YAGNI).
- **Drift guard** — `pythonSetupFeatureFlag.test.ts` pins the coupling that a
  type can't: the `package.json` opt-in enum contains the id, enum and
  `enumDescriptions` stay aligned, the `cliPathOverride` setting is contributed,
  and the feature is disabled by default. If the three coupled strings ever
  drift, CI fails instead of silently shipping a feature nobody can turn on.

## Verification

- `yarn --cwd packages/databricks-vscode build` — clean (confirms the FeatureId
  union extension + JSON `require` type-check).
- `yarn --cwd packages/databricks-vscode test:lint` — clean.
- `yarn --cwd packages/databricks-vscode test:unit` — 309 passing (4 new
  feature-flag tests).

### DIFF
--- a/packages/databricks-vscode/src/feature-manager/FeatureManager.ts
+++ b/packages/databricks-vscode/src/feature-manager/FeatureManager.ts
@@ -11,15 +11,14 @@ export type FeatureEnableAction = (...args: any[]) => Promise<void>;
 /**
  * Feature id for the uv-native Python environment setup (python-setup).
  *
- * This exact string is the single source of truth used in three coupled
- * places, and they must stay identical or the feature can never unlock:
+ * This exact string is the single source of truth used in two coupled places,
+ * and they must stay identical or the feature can never unlock:
  *  - the {@link FeatureManager} `disabledFeatures` entry that hides it by
- *    default (see extension.ts),
+ *    default (see extension.ts), and
  *  - the value a user adds to `databricks.experiments.optInto` to opt in
- *    (see the enum in package.json), which {@link FeatureManager.registerFeature}
- *    matches verbatim against the disabled id to decide whether to unlock, and
- *  - the flag reader `workspaceConfigs.pythonSetupEnabled`.
- * Exporting it as a constant keeps those three from drifting apart.
+ *    (see the enum in package.json), which {@link FeatureManager} matches
+ *    verbatim against the disabled id to decide whether to unlock.
+ * Exporting it as a constant keeps those two from drifting apart.
  */
 export const PYTHON_SETUP_FEATURE_ID = "environment.pythonSetup";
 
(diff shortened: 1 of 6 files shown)


rugpanov:
🤖 Integration tests triggered for `0bf73abd` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30091425271)

rugpanov:
🤖 Integration tests triggered for `33acef93` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30093641096)

packages/databricks-vscode/src/feature-manager/FeatureManager.ts
```diff
@@ -86,7 +116,7 @@ export class FeatureManager<T = FeatureId> implements Disposable {
         if (!feature) {
             throw new Error(`Feature ${id} has not been registered`);
         }
-        if (this.disabledFeatures.includes(id)) {
+        if (this.isDisabled(id)) {
```
misha-db, line 119: new call `isDisabled` isn't equivalent to previous 'this.disabledFeatures.includes(id)' check as it adds additional `&& !workspaceConfigs.experimetalFeatureOverides.includes(id as string)` condition, is it intentional?

rugpanov: Yes, intentional — that added condition *is* the fix. Before this, the two methods disagreed on what "disabled" means:

- `registerFeature` gated on `disabledFeatures.includes(id) && !experimetalFeatureOverides.includes(id)` — so it honored the opt-in and handed back the real feature factory.
- `isEnabled` gated on `disabledFeatures.includes(id)` alone — so it ignored the opt-in and reported `available: false` regardless.

Since every consumer reads state through `isEnabled`, a user who added the id to `databricks.experiments.optInto` would get the real feature registered but `isEnabled` would still report it unavailable — i.e. the opt-in never actually unlocked the feature. Extracting `isDisabled` and using it in both places makes them agree, so opt-in works.

This was latent until now because `disabledFeatures` was empty; `environment.pythonSetup` is the first entry. `environment.dependencies` isn't in the list, so `isDisabled` short-circuits to `false` for it and its behavior is unchanged. The two new tests in `FeatureManager.test.ts` / `pythonSetupFeatureFlag.test.ts` pin the unlock path (they fail without this change).

misha-db (approved):

PR #2046, opened 2026-07-24
[6/?] feat(python-setup): setup orchestrator (first CLI-client caller)

rugpanov:
> **Stacked on #2039** (base = `rugpanov/python-setup-cli-client`). Review after #2039; the diff of interest is the single `controllers/` commit on top. Once #2039 merges, GitHub retargets this to `main` automatically.

## Why

#2039 adds `PythonSetupCliClient` with no call site — as raised in review, the API is hard to judge without seeing it used. This PR adds the orchestrator that actually drives the client end-to-end, so the client's usage ships alongside it.

## What

`PythonSetupEnvironmentSetup` (`python-setup/controllers/`) — the first real caller of `PythonSetupCliClient`:

1. gate → resolve compute → `cli.run(invocation, {cwd, onLog})` under a progress indicator,
2. on success: adopt the provisioned venv interpreter + persist `{envKey, pythonVersion}` for drift detection,
3. on failure: surface the mapped `getPythonSetupErrorMessage` copy,
4. `PythonSetupCancelledError` → treated as a silent user action (no error toast).

It consumes the **real** client surface (`run(invocation{compute}, {cwd, onLog})`, `isPythonSetupSuccess`, `getPythonSetupErrorMessage`), and depends on the client **structurally** (`CliRunner`) so the flow is unit-testable without spawn/`vscode`.

### The two not-yet-built pieces are injected seams, not hard deps

Per the agreed approach, the gate and serverless-version selection are stubbed behind injected functions with explicit TODOs, so this PR stacks only on the CLI-client change and doesn't pull in unbuilt tickets:

- `isVisible()` — **TODO(DECO-27781 / #2044):** `shouldShowPythonSetup(...)`. The wiring stub returns `false`, so the feature is **inert end-to-end** for now.
- `resolveCompute()` — **TODO(DECO-27782):** serverless-version picker (`scoreServerlessVersions` + `pickServerlessVersion`). The wiring stub returns `undefined`, so a serverless setup is a no-op until that ticket; the cluster case can already resolve.

The real gate and picker drop in through these seams in their own tickets (and the extension.ts wiring lands in DECO-27786). Nothing is wired into the extension in this PR.

## Verification

- `yarn --cwd packages/databricks-vscode test:unit` — 344 passing (6 new orchestrator cases: success→ready+adopt+saveState, gate-closed→no run, no-compute→no run, CLI failure→mapped error, cancel→silent, `onDidChangeState` fires).
- `test:lint` — clean · `build` — clean.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
@@ -1,0 +1,212 @@
+import {Disposable, Event, EventEmitter} from "vscode";
+import {
+    CancellationLike,
+    PythonSetupCancelledError,
+    RunOptions,
+} from "../gateways/PythonSetupCliClient";
+import {
+    isPythonSetupSuccess,
+    PythonSetupResult,
+} from "../models/PythonSetupResult";
+import {getPythonSetupErrorMessage} from "../utils/errorMessages";
+import {SetupLocalInvocation} from "../utils/setupLocalArgs";
+
+/**
+ * The one method the orchestrator needs from {@link PythonSetupCliClient}, typed
+ * structurally so the flow is unit-testable without a spawn/`vscode` dependency.
+ */
+export interface CliRunner {
+    run(
+        invocation: SetupLocalInvocation,
+        options: RunOptions
+    ): Promise<PythonSetupResult>;
+}
+
+export type SetupCompute = SetupLocalInvocation["compute"];
+
+/** Persisted after a successful setup, for later drift detection. */
+export interface PythonSetupPersistedState {
+    envKey: string;
+    pythonVersion: string;
+}
+
+type ReadyLocalEnvironmentResult = PythonSetupResult &
+    Required<Pick<PythonSetupResult, "venvPath" | "target" | "resolved">>;
+
+function isLocalEnvironmentReady(
+    r: PythonSetupResult
+): r is ReadyLocalEnvironmentResult {
+    return (
+        isPythonSetupSuccess(r) &&
+        r.venvPath !== undefined &&
+        r.target !== undefined &&
+        r.resolved !== undefined
+    );
+}
+
+/**
+ * Runs under a progress indicator, forwarding `log` to the client's `onLog` and
+ * `token` to its `RunOptions.token` so a user "Cancel" tears down the CLI. The
+ * production wrapper is `window.withProgress` + an output channel.
+ */
+export type ProgressTask<T> = (
+    log: (chunk: string) => void,
+    token: CancellationLike
+) => Promise<T>;
+
+/**
+ * Injected collaborators for {@link PythonSetupEnvironmentSetup}: seams so the
+ * flow is tested without a VS Code host, and so the not-yet-built pieces
+ * (visibility gate, serverless-version selection) can be stubbed until their
+ * own tickets wire the real implementations in.
+ */
+export interface PythonSetupSetupDeps {
+    cli: CliRunner;
+
+    /** Absolute path of the open project the CLI runs against, if any. */
+    projectRoot: () => string | undefined;
+
+    /**
+     * Whether the uv-native setup should run for the current project.
+     *
+     * TODO(#2044): replace the wiring-site stub with
+     * `shouldShowPythonSetup({flagOn: workspaceConfigs.pythonSetupEnabled,
+     * detection: await detector.detect(projectRoot)})`. Until then the extension
+     * passes a stub that returns `false`, so the feature is inert end-to-end.
+     */
+    isVisible: () => Promise<boolean>;
+
+    /**
+     * The compute target to provision for, or `undefined` to abort silently
+     * (nothing selected / the user dismissed the picker).
+     *
+     * TODO(#2052 / #2053): for a serverless session this must resolve the version
+     * via the serverless-version picker (`scoreServerlessVersions` +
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests triggered for `3508259b` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30093974809)

rugpanov:
🤖 Integration tests triggered for `b48d2ae8` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30095295165)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `fa992f32` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30100804733)

packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
```diff
@@ -0,0 +1,218 @@
...
+ * Orchestrates the uv-native "set up Python environment" flow: decide whether
+ * to run, resolve the compute target, invoke the CLI under a progress
+ * indicator, then adopt the provisioned interpreter and persist state on
+ * success — or surface a mapped error on failure.
+ *
+ * This is the first real caller of {@link PythonSetupCliClient}. The visibility
+ * gate and serverless-version selection are injected seams (see
+ * {@link PythonSetupSetupDeps}) that the extension currently wires to stubs, so
+ * shipping this alongside the client shows the client's usage without depending
+ * on tickets that are not built yet.
```
rclarey: nit: I don't think this is useful as code comments
```suggestion
```

packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        // A real-run success must carry everything the success path consumes:
+        // the provisioned venv to adopt, and the target/resolved fields we
+        // persist as the drift-detection baseline. If any is missing (a dry-run
+        // shape, or CLI/schema drift), we can't deliver a usable, tracked
+        // environment — treat it as a failure rather than reporting a hollow
+        // "ready" that either skips interpreter adoption or leaves drift
+        // detection with no baseline.
+        if (!result.venvPath || !result.target || !result.resolved) {
```
rclarey: Can we merge this into `isPythonSetupSuccess` since it's an error case?

packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
```diff
@@ -0,0 +1,218 @@
...
+            await this.deps.showError(getPythonSetupErrorMessage(result));
+            return;
+        }
+
+        await this.deps.adoptInterpreter(result.venvPath);
```
rclarey: Do we intentionally not show an error message when this fails? What is the consequence of the Python extension not being pointed at the provisioned venv interpreter?

rugpanov:
🤖 Integration tests triggered for `1c238f1b` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30296696631)

rugpanov:
🤖 Integration tests triggered for `5f6604a9` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30297095220)

packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
rugpanov: Agreed — trimmed the class doc down to just the flow summary and dropped the PR-narrative prose. Also did a pass over the rest of the file removing comments that only restated self-explanatory names. Done in `f84e6ff`.

packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
rugpanov: I kept `isPythonSetupSuccess` as the CLI-contract predicate (`ok === true`) because it is legitimately true for a successful `--dry-run` (which has no `venvPath`), and folding the stricter check in would break that contract and its existing consumers/test. Instead I extracted the check into a named orchestrator-level type guard, `isLocalEnvironmentReady()`, so the "this is an error case" intent is explicit without changing the shared predicate. `f84e6ff`.

packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
rugpanov: Good catch — that was an oversight, not intentional. Consequence: the venv is provisioned on disk but the Python extension is not pointed at it, so imports/dbconnect will not resolve — and previously it failed silently. Fixed by wrapping `adoptInterpreter()`: a failure now surfaces the error and keeps the flow not-ready. Added a test (`surfaces an error and stays not-ready when interpreter adoption fails`). `f84e6ff`.

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `f84e6ff3`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30297500283)

rclarey (approved):

PR #2052, opened 2026-07-27
[7/?] feat(python-setup): serverless version resolution & picker

rugpanov:
## Why

When a Databricks Connect setup targets serverless, the extension must decide which serverless environment version to provision. The signal comes from several places of differing trust (a bundle YAML declaration, a notebook's recorded environment, the workspace default), so we score candidates by weighted provenance and let the user confirm — rather than guessing or silently auto-selecting. This is the piece the setup orchestrator's compute-resolution seam calls for a serverless target.

Additive: nothing is wired into the extension yet (that's the config-view/orchestrator wiring ticket, DECO-27786), so there is no user-facing change.

## What

Three pure/near-pure units under `python-setup/utils/`:

1. **`serverlessVersionScoring.ts`** — `scoreServerlessVersions(observations)`: deterministic ranking by total per-source weight (`bundleYaml` 100 > `notebook` 50 > `workspaceDefault` 20 > `fallback` 1), ties broken by higher numeric version. A built-in fallback candidate is always present and *merges* into a matching observed version (no duplicate row); repeated `(version, source)` pairs aren't double-counted.
2. **`serverlessVersionPicker.ts`** — `buildVersionPickItems` (pure: pre-picks + stars the top candidate, human-readable provenance in the description) + `pickServerlessVersion` (thin `window.showQuickPick` wrapper). The user must confirm — the recommendation is pre-selected, never auto-applied.
3. **`serverlessVersionResolver.ts`** — `resolveServerlessVersion(deps)`: the flag-guarded entry point that ties collect → score → confirm together, plus `isPythonSetupEnabled()`.

### Constraint 1 — backward compatibility

The new serverless-environment version is a **bare integer** (`4`, `5` — the value the CLI's `--serverless-version` flag accepts, verified against the CLI's own acceptance goldens and `--help`). This is a **distinct namespace** from the existing `databricks.connect.serverlessDbconnectVersion` setting (a DBR/dbconnect version like `17.3`, consumed by the legacy pip flow). The resolver never reads or writes that setting, so existing users are completely unaffected. (Note: the JIRA's `--serverless-version <vN>` phrasing is the CLI's *display* form; the *input* flag is bare — the picker emits bare integers accordingly.)

### Constraint 2 — feature-flag guard (#2045)

[...]

PR #2053, opened 2026-07-27
[8/?] feat(python-setup): merge serverless-version selection into the compute picker

rugpanov:
> **Stacked on #2052** (base = `rugpanov/python-setup-serverless-picker`). Review after #2052; the diff of interest is the commits on top. GitHub retargets to `main` once #2052 merges.

## Why

Selecting **Serverless** in the compute picker only ever recorded a boolean — the serverless *environment version* was then either an invisible setting (legacy pip flow reads `serverlessDbconnectVersion`) or would have to be asked on every setup run. This merges the version choice **into the compute picker**: the user confirms it once, at compute-selection time, and it's persisted with the selection so setup never re-prompts.

Part of DECO-27786 (see the split note below).

## What

1. **Persist** — `OverrideableConfigState.serverlessVersion?: string` (routes through the existing `ConfigModel` overrideable path automatically).
2. **ConnectionManager** — `enableServerless(version?)` persists it (also when serverless is already on, so re-picking the version alone saves); `disableServerless()` clears it; `updateServerless()` loads it on reload **and re-validates it** (a hand-edited / out-of-range value is dropped to the scored default rather than exposed as a valid `--serverless-version`); new `get serverlessVersion()`.
3. **Bundle collector** — `collectBundleServerlessVersions(bundle)`: defensively walks the parsed bundle and harvests every serverless `environment_version` (a bare integer, same form `--serverless-version` takes) as a `bundleYaml` scoring observation, so the picker is seeded with real candidates rather than a lone fallback.
4. **Notebook collector** — `collectNotebookServerlessVersions(notebooks)` + `collectProjectNotebookVersions(projectRoot)`: harvest the serverless version recorded in `.ipynb` metadata (`metadata["application/vnd.databricks.v1+notebook"].environmentMetadata.environment_version`) as a `notebook` scoring observation. The project scan excludes dependency/venv/build trees (`.venv`, `site-packages`, `node_modules`, `.git`, `.databricks`, `dist`, `build`) so a package's sample notebook can't inject a spurious signal, is capped (200 notebooks) and runs bounded-concurrency batches with a cheap substring pre-filter to skip the JSON parse of notebooks that don't declare a version — so it can't stall the compute picker on a large repo.
5. **Picker wiring** — `ConnectionCommands`: choosing Serverless routes through `selectServerless()`, which (when opted in) collects the bundle + notebook observations, ranks them, shows #2052's picker (top candidate pre-selected), and persists via `enableServerless(version)`. Each source is collected independently and guarded, so one failing source never blocks the other or compute selection.

## Constraint 1 — backward compatibility

- The persisted field is **optional and independent of `serverless`**. Existing configs with `serverless: true` and no version stay valid — a version-less selection falls back to the scored default (`5`). No migration writes; writing an unrelated key never fabricates a version (tested). A persisted version is re-validated on load, so a downgrade/upgrade that leaves an out-of-range value can't leak it to the CLI.
- **Flag OFF ⇒ the compute picker enables plain serverless, unchanged** (`enableServerless()` with no version). The new field is just an unused optional.
- The pre-existing `databricks.connect.serverlessDbconnectVersion` setting (a DBR/dbconnect version like `17.3`, used by the legacy pip rail) is a **separate namespace** — this code never reads or writes it.
- A bundle or notebook read/parse failure falls back to scoring with no observations rather than blocking compute selection.

## Constraint 2 — feature-flag guard (#2045)

The version sub-step is gated on `isPythonSetupEnabled()` (the `PYTHON_SETUP_FEATURE_ID` opt-in from #2045). Only opted-in users ever see the version picker; everyone else gets the unchanged serverless selection.

## Verification

- `yarn --cwd packages/databricks-vscode test:unit` — 364 passing (config persistence incl. legacy-config-untouched, bundle collector, notebook collector, and project-notebook scan incl. exclusions / parse-skip / batch boundary). ConnectionManager + picker wiring type-checked by `build`.
- `test:lint` — clean · `build` — clean.
- Integration tests — ✅ all 35 jobs passed on the latest commit.

## Note on the ticket split

DECO-27786 (extension wiring & config-view dispatch) is being split into two diffs. **This is Diff A** — the compute-picker merge (shared compute code, releasable on its own). **Diff B** (construct detector/client/orchestrator in `activate()`, `EnvironmentComponent` dispatch, read `serverlessVersion` from `ConnectionManager` in the setup flow) will follow once the orchestrator (#2046) and client (#2039) merge, so it can go cleanly off `main`. Nothing in this diff consumes `serverlessVersion` yet — the picker persists it for Diff B to read.

### DIFF
--- a/packages/databricks-vscode/src/configuration/ConnectionCommands.ts
+++ b/packages/databricks-vscode/src/configuration/ConnectionCommands.ts
@@ -23,6 +23,13 @@ import {
 } from "../ui/configuration-view/AuthTypeComponent";
 import {ManualLoginSource} from "../telemetry/constants";
 import {onError} from "../utils/onErrorDecorator";
+import {isPythonSetupEnabled} from "../python-setup/utils/serverlessVersionResolver";
+import {
+    scoreServerlessVersions,
+    VersionObservation,
+} from "../python-setup/utils/serverlessVersionScoring";
+import {pickServerlessVersion} from "../python-setup/utils/serverlessVersionPicker";
+import {collectBundleServerlessVersions} from "../python-setup/utils/bundleServerlessVersions";
 
 function formatQuickPickClusterSize(sizeInMB: number): string {
     if (sizeInMB > 1024) {
@@ -187,7 +194,11 @@ export class ConnectionCommands implements Disposable {
                     const cluster = selectedItem.cluster;
                     await this.connectionManager.attachCluster(cluster.id);
                 } else if (selectedItem.label === "$(cloud) Serverless") {
-                    await this.connectionManager.enableServerless();
+                    // Dispose the compute QuickPick before opening the version
+                    // sub-picker so they don't stack visually.
+                    disposables.forEach((d) => d.dispose());
+                    await this.selectServerless();
+                    return;
                 } else {
                     await UrlUtils.openExternal(
                         `${
@@ -206,6 +217,46 @@ export class ConnectionCommands implements Disposable {
                 quickPick.dispose();
             });
         };
+    }
+
+    /**
+     * Enable serverless compute. When the uv-native python-setup feature is
+     * opted into, first ask the user to confirm the serverless environment
+     * version (ranked from the project's bundle) and persist it with the
+     * selection, so setup need not re-prompt. If they dismiss the version
+     * picker, no compute change is made. With the feature off this is the
+     * plain, unchanged serverless enable.
+     */
+    private async selectServerless() {
+        if (!isPythonSetupEnabled()) {
+            await this.connectionManager.enableServerless();
+            return;
+        }
+        const version = await this.pickServerlessVersion();
+        if (version === undefined) {
+            // User dismissed the version picker -- don't switch compute.
+            return;
+        }
+        await this.connectionManager.enableServerless(version);
+    }
+
+    /**
+     * Rank the serverless environment versions declared in the project's bundle
+     * and let the user confirm one (top candidate pre-selected). Returns the
+     * confirmed bare version, or undefined if dismissed.
+     */
+    private async pickServerlessVersion(): Promise<string | undefined> {
+        let observations: VersionObservation[];
+        try {
+            const validateConfig = await this.configModel.get("validateConfig");
+            observations = collectBundleServerlessVersions(validateConfig);
+        } catch {
+            // Bundle not available/parseable -- fall back to scoring with no
+            // observations (yields the default candidate), never block compute
+            // selection on it.
+            observations = [];
+        }
+        return pickServerlessVersion(scoreServerlessVersions(observations));
     }
 
     /**
--- a/packages/databricks-vscode/src/configuration/ConnectionCommands.ts
+++ b/packages/databricks-vscode/src/configuration/ConnectionCommands.ts
@@ -29,6 +29,9 @@ import {
 } from "../python-setup/utils/serverlessVersionResolver";
 import {pickServerlessVersion} from "../python-setup/utils/serverlessVersionPicker";
 import {collectBundleServerlessVersions} from "../python-setup/utils/bundleServerlessVersions";
+import {collectProjectNotebookVersions} from "../python-setup/utils/projectNotebookVersions";
+import {VersionObservation} from "../python-setup/utils/serverlessVersionScoring";
+import {WorkspaceFolderManager} from "../vscode-objs/WorkspaceFolderManager";
 
 function formatQuickPickClusterSize(sizeInMB: number): string {
     if (sizeInMB > 1024) {
@@ -69,7 +72,8 @@ export class ConnectionCommands implements Disposable {
         private connectionManager: ConnectionManager,
         private readonly clusterModel: ClusterModel,
         private readonly configModel: ConfigModel,
-        private readonly cli: CliWrapper
+        private readonly cli: CliWrapper,
+        private readonly workspaceFolderManager: WorkspaceFolderManager
     ) {}
 
     /**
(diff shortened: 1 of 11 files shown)


rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `81c66cb1` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30253279069)

rugpanov:
🤖 Integration tests triggered for `81a23384` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30262508473)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `52ac5a3e` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30263205090)

rugpanov:
🤖 Integration tests triggered for `30adfe61` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30270801500)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `aa0d1239`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30271639221)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `20e8f291`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30296683257)

packages/databricks-vscode/src/configuration/ConnectionCommands.ts
```diff
@@ -187,7 +197,11 @@ export class ConnectionCommands implements Disposable {
                     const cluster = selectedItem.cluster;
                     await this.connectionManager.attachCluster(cluster.id);
                 } else if (selectedItem.label === "$(cloud) Serverless") {
-                    await this.connectionManager.enableServerless();
+                    // Dispose the compute QuickPick before opening the version
+                    // sub-picker so they don't stack visually.
+                    disposables.forEach((d) => d.dispose());
```
misha-db, line 202: Dispose happens already onDidHide (next statement after if/else block), why do we need to do it onDidAccept as well?

packages/databricks-vscode/src/python-setup/utils/projectNotebookVersions.ts
```diff
@@ -0,0 +1,79 @@
...
+    projectRoot: string
+): Promise<VersionObservation[]> {
+    const allFiles = await glob(path.join(projectRoot, "**", "*.ipynb"), {
+        // path.join uses "\" on Windows; glob wants "/", so opt out of escaping.
+        windowsPathsNoEscape: true,
+        ignore: IGNORED_NOTEBOOK_DIRS,
+        nodir: true,
+    });
+
+    const files = allFiles.slice(0, MAX_NOTEBOOKS_SCANNED);
```
misha-db, line 76: What's the performance of `collectProjectNotebookVersions` functions when parsing MAX_NOTEBOOKS_SCANNED (200) notebooks? How does it influence UX

misha-db (approved):

misha-db (approved):

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `d86a3cbc` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30358261196)

packages/databricks-vscode/src/configuration/ConnectionCommands.ts
rugpanov: Good question — it's needed here specifically because this is a `window.createQuickPick()` instance, not `showQuickPick()`. `createQuickPick` doesn't auto-hide when the user accepts, so at `onDidAccept` time the compute picker is still visible on screen. The serverless branch then opens the *version* sub-picker inside `selectServerless()`, so we hide the compute picker first or the two stack visually.

The other two dispose sites don't cover this moment: the dispose at the end of the callback is skipped by the early `return`, and `onDidHide` hasn't fired yet (nothing has hidden the picker). Disposal is idempotent, so the later `onDidHide` dispose remains the safety net for the dismiss path. The other accept branches fall through to the end-of-callback dispose and don't open a sub-picker, so they don't need the early dispose.

packages/databricks-vscode/src/python-setup/utils/projectNotebookVersions.ts
rugpanov: Fair concern — addressed in d86a3cb. The original code read + fully `JSON.parse`d every match via `Promise.all`, and `.ipynb` files embed cell outputs (plots/images) so they can be tens of MB — a real latency/memory spike on the interactive path in the 200-notebook worst case.

Two changes:
- **Substring pre-filter** — skip `JSON.parse` (and its large-object allocation) for any notebook whose raw text doesn't even contain `environment_version`. That's the common case, since most notebooks declare no serverless environment, so we only parse the handful that could actually contribute.
- **Bounded concurrency** (batches of 20) instead of reading all matches at once, keeping peak memory flat regardless of repo size.

Combined with the earlier `.venv`/`site-packages`/`node_modules`/build-dir exclusions and the 200 cap, the typical project (a handful of notebooks) is negligible, and the pathological large-repo case is bounded. New tests cover the parse-skip and the batch boundary (>20 notebooks).

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `8e57467a` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30364172818)

PR #2056, opened 2026-07-28
test(e2e): harden run_files cancel against toast-action race and unbounded poll

rugpanov:
## Why

`Run files :: should cancel a run during deployment` is the **top e2e flake** in recent `vscode-isolated-pr` runs — 3 hits in ~24h, all on the Windows shard — with two distinct failure signatures that both trace to unguarded spots in the test:

- **`Can't call elementClick ... title='Cancel' ... element wasn't found`** — `takeAction("Cancel")` runs *inside* a `browser.waitUntil` condition. wdio rejects a condition that **throws** rather than retrying it, so a Cancel button that hasn't rendered yet kills the test instead of being retried on the next poll.
- **bare `Error: Timeout`** — the follow-up `while (true)` poll for `"Cancelled"` in the debug console has **no timeout**, so a missed cancel spins until mocha's per-test timeout with zero diagnostic context.

## What

- Wrap the notification message read and `takeAction("Cancel")` in `try/catch` inside the `waitUntil` condition, so a not-yet-rendered button (or a toast that vanished between listing and click) simply retries on the next 500ms poll. A `cancelled` flag is asserted afterward, with an explicit `timeoutMsg`, so a genuinely absent toast fails **closed** with a clear reason (60s budget).
- Replace the unbounded `while (true)` `"Cancelled"` poll with a bounded `browser.waitUntil` (120s, 2s interval) that fails with a diagnostic `timeoutMsg` instead of a context-free mocha timeout.

No production code changes. The test still requires a **real** Cancel click and a **real** `"Cancelled"` debug-console line, so a broken cancel path cannot pass silently.

## Verification

- `prettier --check`: clean
- `eslint`: clean
- `tsc --noEmit` (e2e tsconfig): no errors
- Isolated `vscode-isolated-pr` runs dispatched against this branch.

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/run_files.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/run_files.e2e.ts
@@ -49,30 +49,76 @@ describe("Run files", async function () {
     it("should cancel a run during deployment", async () => {
         const workbench = await driver.getWorkbench();
         await executeCommandWhenAvailable("Databricks: Upload and Run File");
-        await browser.waitUntil(async () => {
-            const notifications = await workbench.getNotifications();
-            console.log("Notifications:", notifications.length);
-            for (const notification of notifications) {
-                const message = await notification.getMessage();
-                console.log("Message:", message);
-                if (message.includes("Uploading bundle assets")) {
-                    await notification.takeAction("Cancel");
-                    return true;
+
+        // Catching the in-flight "Uploading bundle assets" toast and clicking
+        // its Cancel action is a tight race on the slow Windows shard: the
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `8062285a`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30360593353)

misha-db (approved):

rugpanov:
🤖 Integration tests triggered for `07b04039` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30452988147)

PR #2074, opened 2026-08-04
[10/?] feat(python-setup): setup attempt/result telemetry

rugpanov:
## Why

The VPEX Python-environment setup flow is uninstrumented: we cannot see how often setup runs, where it breaks, or how long it takes.

## What

Two events emitted from the setup orchestrator, following the existing `PackageManagerTelemetry` / `packageManagerExtensions` split — module augmentation for the emit half, injected deps for collection:

- **`python_env.setup.attempt`** — package manager, target type, serverless version, mode, is-greenfield. Recorded once the compute target resolves and a CLI run is about to start.
- **`python_env.setup.result`** — outcome (`ok` / `failed` / `cancelled` / `not_started`), failure phase, error code, env key, disk-mutated, duration.

`recordPythonSetupAttempt` returns the result reporter, so the 1:1 attempt/result pairing is structural rather than a convention.

Together with the existing `python_env.setup.detected` (`explicit_command`) this gives a three-stage funnel: click → attempt → result.

### Two requested fields are not taken from the CLI result

Because it has no producer for either:

- **Duration is measured in-extension.** The CLI documents `durationMs` as reserved and always emits `0` (`libs/localenv/result.go`). Measuring here also captures the latency the user actually experiences, including process spawn and interpreter adoption.
- **The merge-conflict warning count is omitted.** Nothing in the CLI ever appends to `Result.Warnings` (`NewResult()` seeds it to `[]`; only the text renderer reads it), and merge conflicts are not a modelled concept there. The field would be a permanent `0` — a dashboard on it would read "merge quality is perfect" rather than "unmeasured". It can be added once the CLI has a producer.

### Notes

- `adopt` is added as a synthetic seventh failure phase for the extension-side interpreter-adoption step, which happens *after* the CLI exits and so cannot appear in its own `phases[]`.
- Categorical/enum data only — no cluster IDs/names, paths, or package names. Absent optionals are omitted rather than stringified to `"undefined"`.
- `isGreenfield` is only emitted for uv/unknown projects: for pip/conda users a missing `pyproject.toml` says nothing about greenfield-ness, so reporting it would inflate the rate.
- Emission is fully best-effort — a detection, emit, or report failure can never break the setup run it measures.
- `mode` is constant `default` until the `--constraints-only` picker ships; the field is in the schema now so no migration is needed then.
- `telemetry.json` is generated from `EventTypes` (`scripts/generateTelemetry.ts`), so it needs no manual edit.

Schema and privacy docs: `src/telemetry/PYTHON_SETUP_TELEMETRY.md`.

## Verification

- `yarn build` — clean; `yarn test:lint` — 0 errors, prettier clean.
- Unit tests: **461 passing, +14 new** — attempt/result for each outcome path, no-attempt for each early return, the pairing invariant, greenfield conditionality, no-cluster-id, and telemetry-failure resilience.
- The 6 failures in `cli/CliWrapper.test.ts` are pre-existing on the base commit (7 there, one flaky) and untouched by this change.
- Regenerated `telemetry.json` and confirmed both events serialize.

## Backward compatibility

Additive only: two new events, and new required fields on the internal `PythonSetupSetupDeps` / `PythonSetupWiringDeps` seams (both internal to the extension). No change to existing event schemas, settings, persisted state, or the CLI contract. Behaviour for users who have not opted into the feature flag is unchanged.


rugpanov:
🤖 Integration tests triggered for `8bf51268` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30892664713)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `bef83830`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30893067423)

packages/databricks-vscode/src/telemetry/PYTHON_SETUP_TELEMETRY.md
```diff
@@ -0,0 +1,178 @@
+# Telemetry: Python environment setup attempt / result
```
misha-db: Why do we need this markdown files under src folder?

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `e4236807`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30896236691)

packages/databricks-vscode/src/telemetry/PYTHON_SETUP_TELEMETRY.md
rugpanov: It aims to help understanding what exact telemetry are we collecting. Should it be README.md instead?

misha-db: Ideally we should have separate `docs` folder (in project/package root?) for things like that. mixing source code and documentation makes navigation less straightforward

rugpanov:
🤖 Integration tests triggered for `a87085a6` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30899710648)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `cc87c310` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30900096748)

misha-db (approved):

PR #2075, opened 2026-08-04
docs: add a convention for where module documentation goes

rugpanov:
## Why

We have no rule for where a module's prose documentation lives, and the result drifts. `src/telemetry/` picked up a feature-specific `PACKAGE_MANAGER_DETECTION.md` that nothing links to, and a second such file was about to land beside it. A folder that accumulates one markdown file per feature is a folder nobody reads.

## What

Adds `CODE_CONVENTIONS.md` section 4a with three rules:

- **When a module needs a prose doc, it is `README.md` in the module's folder** — the name tooling and humans already look for, and what GitHub renders when browsing. Not a per-feature descriptive filename.
- **Document the *why*, not the *what*.** Schemas, field lists and signatures belong in the code or in generated output; prose that restates them becomes a second source of truth that drifts.
- **No top-level `docs/`** — it's gitignored in this repo, so anything committed lives beside the code or in an existing root file.

Also states explicitly (here and in the section 4 telemetry bullet) that `EventTypes` in `telemetry/constants.ts` is the only place an event schema may be maintained, since `telemetry.json` is generated from it.

## The survey behind it

The section records the evidence rather than just asserting a preference, so the premise can be argued with:

| Project | Practice |
|---|---|
| microsoft/vscode | Colocated docs, **no top-level `docs/`** — ~29 architecture docs under `src/` (e.g. `src/vs/sessions/LAYOUT.md`) |
| facebook/react | Colocated `README.md` per package |
| rust-lang/rust | Colocated `README.md` per crate, pointing at the external dev guide |
| vscode-pull-request-github | Centralized `documentation/` |
| kubernetes, eslint | Centralized `docs/` + external enhancement proposals |

Both colocation and centralization are well-established; there is no universal answer. We colocate because `docs/` is gitignored here, and standardize on `README.md` rather than VS Code's descriptive-filename style because the latter only stays navigable when a folder is one subsystem — `src/telemetry/` serves every feature.

Worth noting for telemetry specifically: VS Code documents events *inline in TypeScript* (GDPR classification and comment in a type beside the emit call), with no per-event markdown. Our `EventTypes` `comment` fields are the same idea, which is why prose docs must not duplicate them.

## Notes

- Split out of #2074 on review feedback — a repo-wide rule shouldn't ride in on a feature PR. That PR keeps only its own `src/telemetry/README.md`.
- The existing `PACKAGE_MANAGER_DETECTION.md` is left in place; the section says to fold such files in when next touched, rather than churning a file this PR has no other reason to open.
- This documents a convention, it doesn't enforce one. No lint rule.

## Verification

Docs only, no code touched. I verified every path and claim the section cites exists on `main` (`PACKAGE_MANAGER_DETECTION.md`, `python-setup/README.md`, `scripts/generateTelemetry.ts`, `docs/` in `.gitignore`) and that the external references resolve. Prettier clean with the repo-pinned 3.1.1.

### DIFF
--- a/CODE_CONVENTIONS.md
+++ b/CODE_CONVENTIONS.md
@@ -100,6 +100,8 @@ makes it hard to unit-test (see section 8).
    dependencies, register commands.
 7. **Add a barrel (`index.ts`) only if the feature has a clear public surface**
    other modules import — and make it a _selective_ re-export, not `export *`.
+8. **If the feature needs a prose doc, add `README.md` to its folder** — see
+   section 4a. Most features don't.
 
 A typical new feature:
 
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests triggered for `1e946688` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30900084815)

rugpanov:
🤖 Integration tests triggered for `28ca1b60` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/30900480708)

misha-db (approved):

PR #2077, opened 2026-08-04
fix(shellUtils): generate terminal commands for the shell that runs them (#1822)

### LINKED ISSUE #1822: [BUG] 'clear' is not recognized as an internal or external command
**Describe the bug**
When trying to create a new Databricks project I get the following error after authentication:
```
C:\Users\USER~1.NAM\AppData\Local\Temp>clear; echo "Executing: databricks bundle init --output-dir "c:\Users\user.name\Documents\databricks-test"
'clear' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\USER~1.NAM\AppData\Local\Temp>Follow the steps below to create your new Databricks project.
'Follow' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\USER~1.NAM\AppData\Local\Temp>"; 'c:\Users\user.name\.vscode\extensions\databricks.databricks-2.9.4-win32-x64\bin\databricks' bundle init --output-dir "c:\Users\user.name\Documents\databricks-test"; echo "
The filename, directory name, or volume label syntax is incorrect.

C:\Users\USER~1.NAM\AppData\Local\Temp>Press any key to close the terminal and continue ..."; read; exit
'Press' is not recognized as an internal or external command,
operable program or batch file.
```

**To Reproduce**

1. Install the extension and restart extensions
2. Click on the Databricks extension on the left panel
3.Click on **Create a new Databricks project**
4. Complete on-screen instructions to authenticate to Databricks (OAuth in my case)
5. Select a folder where you want your project to be
6. See error

**Screenshots**
```
C:\Users\USER~1.NAM\AppData\Local\Temp>clear; echo "Executing: databricks bundle init --output-dir "c:\Users\user.name\Documents\databricks-test"
'clear' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\USER~1.NAM\AppData\Local\Temp>Follow the steps below to create your new Databricks project.
'Follow' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\USER~1.NAM\AppData\Local\Temp>"; 'c:\Users\user.name\.vscode\extensions\databricks.databricks-2.9.4-win32-x64\bin\databricks' bundle init --output-dir "c:\Users\user.name\Documents\databricks-test"; echo "
The filename, directory name, or volume label syntax is incorrect.

C:\Users\USER~1.NAM\AppData\Local\Temp>Press any key to close the terminal and continue ..."; read; exit
'Press' is not recognized as an internal or external command,
operable program or batch file.
```

**System information:**

1. Version: 1.107.0 (user setup)
Commit: 618725e67565b290ba4da6fe2d29f8fa1d4e3622
Date: 2025-12-10T07:43:47.883Z
Electron: 39.2.3
ElectronBuildId: 12895514
Chromium: 142.0.7444.175
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200
2. Databricks Extension Version: 2.10.4

**Databricks Extension Logs**
[databricks-cli-logs.json](https://github.com/user-attachments/files/24184265/databricks-cli-logs.json)
[sdk-and-extension-logs.json](https://github.com/user-attachments/files/24184272/sdk-and-extension-logs.json)

abdelrahmanomar1515:
Fixes #1822.

## Why

On Windows with cmd.exe as the terminal shell, "Create a new Databricks project"
was broken. The wizard built one POSIX command line and sent it to whatever shell
the terminal was running, so cmd rejected it token by token and the CLI never
started ([#1822](https://github.com/databricks/databricks-vscode/issues/1822)):

```
C:\...\Temp>clear; echo "Executing: databricks bundle init --output-dir "c:\...\databricks-test"
'clear' is not recognized as an internal or external command,
'Follow' is not recognized as an internal or external command,
'Press' is not recognized as an internal or external command,
```

`clear`, `;` as a separator, and `read` are all POSIX-only. cmd needs `cls`, ` & `
and `pause`.

The root cause is structural, not a missing branch. `shellUtils` read `env.shell`
inside every helper and split only "PowerShell" vs "everything else", using a
substring match on the full path. cmd.exe fell into the POSIX bucket. Because the
branching sat behind a module-level VS Code read, none of it could be reached from
a unit test — so this class of bug had no way to be caught. Fixing only the
`clear`/`read` symptoms would have left the next shell-specific bug just as
invisible, which is why this is a refactor rather than a two-line patch.

## What changed

**One classifier instead of scattered predicates.**
`detectShellKind(shell, platform) -> "cmd" | "powershell" | "posix"` is a pure
function; every helper takes that kind. `env.shell` is now read in exactly one
place, `currentShellKind()`, matching the existing `venvInterpreterPath`
injection pattern in this repo. Detection matches on the shell's **basename**, so
`C:\cmder\...\bash.exe` is no longer mistaken for cmd.

**Terminals pin their shell.** The two terminals we create pass `shellPath`, so
the shell that parses the command is the one it was generated for, instead of
whatever `terminal.integrated.defaultProfile` happens to be. This matters more
now: a cmd-shaped line (` & ` separators) is a *parse* error in PowerShell, so
nothing would run — not even the trailing `exit` — and the wizard would hang
awaiting a terminal-close event.

**Command construction extracted.** `bundleInitCommand.ts` holds the assembled
`databricks bundle init` line, deliberately free of `vscode` imports so it is a
pure function of (cli path, output dir, shell) and can be asserted for all three
shells.

Per file:

| File | Change |
|---|---|
| `utils/shellUtils.ts` | `ShellKind` + `detectShellKind`; all helpers parameterised; per-shell quoting; `clearCmd`/`commandSeparator`/`echoCmd`/`hasCmdUnsafeChars` added |
| `bundle/bundleInitCommand.ts` | New — pure builder for the bundle-init command line |
| `bundle/BundleInitWizard.ts` | Uses the builder; pins `shellPath` |
| `configuration/auth/AzureCliCheck.ts` | Same hardcoded `;`/`echo` problem fixed; pins `shellPath`; quotes `azBinPath` |
| `cli/CliWrapper.ts` | `escapedCliPathFor(kind)`; drops duplicated inline escaping |
| `run/RunCommands.ts` | Resolves the shell kind once and threads it through |

### Bugs found once the logic became testable

Each is a real defect on `main`, not a hypothetical:

- **`pwsh[.exe]` was unhandled** — PowerShell 6+/7, the default on modern
  Windows. It matched neither `"powershell"` nor cmd, so those users got POSIX
  `read`, which does not exist in PowerShell.
- **PowerShell paths were double-quoted, which interpolates.**
  `C:\Users\me\$RECYCLE.BIN` expanded to `C:\Users\me\.BIN`, and a directory
  named `C:\a$(whoami)b` would **execute** `whoami`. Both PowerShell and POSIX
  now use single quotes, which are literal.
- **`escapedCliPath` used `.replace`, not `.replaceAll`** — every quote after the
  first was left unescaped.
- **POSIX used `echo`**, whose builtin form interprets backslash escapes in zsh
  and dash *even inside single quotes*: `\t` printed a tab, and `\c` truncated
  the rest of the line. Now `printf '%s\n'`, which is verbatim everywhere. (The
  message is a printf *argument*, not the format string, so a literal `%` is
  safe.)
- **Bare `read` is a usage error in dash** (`read: arg count`, exit 2). The
  hold-open step failed and the following `exit` closed the terminal, discarding
  exactly the CLI error it exists to keep readable. Now `read _`.
- **`%VAR%` cannot be escaped in cmd** — it expands even inside double quotes.
  Added `hasCmdUnsafeChars` so callers can refuse rather than silently pass the
  wrong directory.
- **Empty `env.shell`** (shell-less environments) now falls back to the platform
  default rather than unconditionally POSIX, since `shellPath` cannot be pinned
  to `""` and Windows would launch PowerShell.

## How this was tested

Manual tests in a windows VM, setting `terminal.integrated.defaultProfile.windows` to "Command Prompt" (default is Powershell).
`yarn test:unit` — **554 passing, 0 failing** (was 505 before this branch), plus
`yarn test:lint` clean. 105 of those cases come from the two new suites.

Three layers, chosen so the weaknesses of each are covered by another:

**1. Table-driven classification.** `detectShellKind` is checked across 11
shell/platform pairs plus explicit regression cases: the cmder false positive,
`cmd` on Linux, `pwsh` on macOS, and the empty-shell fallback per platform.
Because both inputs are parameters, these are deterministic on any host OS — no
`process.platform` patching, and no mocking anywhere in the suite.

**2. Exact command strings.** The full bundle-init line is asserted verbatim for
cmd, PowerShell and POSIX, so a reviewer can read the expected output for each
shell directly in the test.

**3. Round-trip execution against real shells.** String equality only proves the
code built what the author intended — not that the shell agrees. So the generated
commands are executed through `/bin/sh`, `zsh`, `bash` and `dash` (each gated on
`existsSync`), asserting stdout matches the input byte for byte across 10 awkward
messages and 8 awkward paths: embedded quotes of both kinds, `$VAR`, `` `cmd` ``,
`$(cmd)`, `&`/`|`/`>`, backslashes and multi-line input. This layer is what caught
the `echo` and `read` bugs above; the string assertions alone had passed.

**Verifying the tests actually bite.** Since a test that passes either way proves
nothing, I reverted each fix individually in the compiled output and recorded the
failures:

| Fix reverted | Tests failing |
|---|---|
| PowerShell single-quoting | 9 |
| `printf` instead of `echo` | 12 (in `sh`/`zsh` only — `bash` passes, matching the real defect) |
| `read _` instead of `read` | 3 (incl. the `dash` round-trip) |
| `%` guard | 1 |
| empty-shell fallback | 1 |

Every fix has at least one test that fails without it.

## Reviewer notes

- **`escapedCliPath` is retained** alongside the new `escapedCliPathFor(kind)`
  because `SshCommands.ts` still uses it. Behaviour there is unchanged (it
  defaults to `currentShellKind()`), but it is the one remaining place a path
  could be escaped for a different shell than the command it lands in.
- `echoCmd` now splits on **real** newlines. The previous call site passed a
  literal `\n`; if you add a caller, pass an actual newline.

This pull request and its description were written by Isaac, then edited and reviewed by me.

### DIFF
--- a/packages/databricks-vscode/src/bundle/bundleInitCommand.ts
+++ b/packages/databricks-vscode/src/bundle/bundleInitCommand.ts
@@ -1,0 +1,41 @@
+import {
+    clearCmd,
+    commandSeparator,
+    echoCmd,
+    escapePathArgument,
+    readCmd,
+    ShellKind,
+} from "../utils/shellUtils";
+
+/**
+ * The full `databricks bundle init` line sent to the wizard terminal: clear the
+ * screen, explain what is running, run the CLI, then hold the terminal open
+ * until the user acknowledges so any CLI error stays readable.
+ *
+ * Kept in its own module, free of `vscode` imports, so it stays a pure function
+ * of (cli path, output dir, shell) and can be asserted for every shell family.
+ * `escapedCliPath` is passed in because it is derived from CLI state.
+ */
+export function buildBundleInitCommand(
+    escapedCliPath: string,
+    outputDir: string,
+    kind: ShellKind
+): string {
+    const args = [
+        "bundle",
+        "init",
+        "--output-dir",
+        escapePathArgument(outputDir, kind),
+    ].join(" ");
+    return [
+        clearCmd(kind),
+        echoCmd(
+            `Executing: databricks ${args}\nFollow the steps below to create your new Databricks project.\n`,
+            kind
+        ),
+        `${escapedCliPath} ${args}`,
+        echoCmd("\nPress any key to close the terminal and continue ...", kind),
+        readCmd(kind),
+        "exit",
+    ].join(commandSeparator(kind));
+}
--- a/packages/databricks-vscode/src/bundle/bundleInitCommand.ts
+++ b/packages/databricks-vscode/src/bundle/bundleInitCommand.ts
@@ -3,9 +3,34 @@ import {
     commandSeparator,
     echoCmd,
     escapePathArgument,
+    hasCmdUnsafeChars,
     readCmd,
     ShellKind,
 } from "../utils/shellUtils";
+
+/**
+ * Why `outputDir` cannot be passed through `kind` faithfully, or undefined when
+ * it can.
+ *
+ * cmd.exe expands `%VAR%` — and `!VAR!` under delayed expansion — even inside
+ * double quotes, and an interactive command line has no escape for either. The
+ * CLI would scaffold into a directory the user never chose, after which
+ * `getSubProjects` reports "we haven't detected any Databricks projects in …":
+ * a silent wrong-directory, so callers must refuse rather than send.
+ */
+export function unsafeOutputDirReason(
+    outputDir: string,
+    kind: ShellKind
+): string | undefined {
+    if (kind === "cmd" && hasCmdUnsafeChars(outputDir)) {
+        return (
+            `The folder path "${outputDir}" contains a "%" or "!" character, which cmd.exe always expands ` +
+            `as a variable, so the project would be created somewhere else. Choose a folder without those ` +
+            `characters, or set terminal.integrated.defaultProfile.windows to PowerShell.`
+        );
+    }
+    return undefined;
+}
 
 /**
  * The full `databricks bundle init` line sent to the wizard terminal: clear the
@@ -15,12 +40,21 @@ import {
  * Kept in its own module, free of `vscode` imports, so it stays a pure function
  * of (cli path, output dir, shell) and can be asserted for every shell family.
  * `escapedCliPath` is passed in because it is derived from CLI state.
+ *
+ * Throws when `outputDir` cannot be represented in `kind` — see
+ * {@link unsafeOutputDirReason}. Callers should check that first and surface the
+ * reason; the throw is a backstop so no path can reach a shell that would
+ * silently rewrite it.
  */
 export function buildBundleInitCommand(
     escapedCliPath: string,
     outputDir: string,
     kind: ShellKind
 ): string {
+    const reason = unsafeOutputDirReason(outputDir, kind);
+    if (reason !== undefined) {
+        throw new Error(reason);
+    }
     const args = [
         "bundle",
         "init",
--- a/packages/databricks-vscode/src/run/RunCommands.ts
+++ b/packages/databricks-vscode/src/run/RunCommands.ts
@@ -7,6 +7,7 @@ import {MsPythonExtensionWrapper} from "../language/MsPythonExtensionWrapper";
 import path from "path";
 import {FeatureManager, FeatureState} from "../feature-manager/FeatureManager";
 import {
+    currentShellKind,
     escapeExecutableForTerminal,
     escapePathArgument,
 } from "../utils/shellUtils";
@@ -244,10 +245,13 @@ export class RunCommands {
             path.join("resources", "python", "dbconnect-bootstrap.py")
         );
         terminal.show();
+        const kind = currentShellKind();
         terminal.sendText(
-            `${escapeExecutableForTerminal(executable)} ${escapePathArgument(
-                bootstrapPath
-            )} ${escapePathArgument(targetResource.fsPath)}`
+            [
+                escapeExecutableForTerminal(executable, kind),
+                escapePathArgument(bootstrapPath, kind),
+                escapePathArgument(targetResource.fsPath, kind),
+            ].join(" ")
         );
 
         this.telemetry.recordEvent(Events.DBCONNECT_RUN, {
--- a/packages/databricks-vscode/src/run/RunCommands.ts
+++ b/packages/databricks-vscode/src/run/RunCommands.ts
@@ -7,9 +7,9 @@ import {MsPythonExtensionWrapper} from "../language/MsPythonExtensionWrapper";
 import path from "path";
 import {FeatureManager, FeatureState} from "../feature-manager/FeatureManager";
 import {
-    currentShellKind,
     escapeExecutableForTerminal,
     escapePathArgument,
+    terminalShellKind,
 } from "../utils/shellUtils";
 import {CustomWhenContext} from "../vscode-objs/CustomWhenContext";
 import {WorkspaceFolderManager} from "../vscode-objs/WorkspaceFolderManager";
@@ -245,7 +245,10 @@ export class RunCommands {
             path.join("resources", "python", "dbconnect-bootstrap.py")
         );
         terminal.show();
-        const kind = currentShellKind();
+        // Classify the terminal we are actually sending to. This may be a
+        // pre-existing tab running any profile, so `env.shell` (the *default*
+        // profile) is not necessarily the shell that parses this line.
+        const kind = terminalShellKind(terminal);
         terminal.sendText(
             [
                 escapeExecutableForTerminal(executable, kind),
--- a/packages/databricks-vscode/src/utils/shellUtils.ts
+++ b/packages/databricks-vscode/src/utils/shellUtils.ts
@@ -1,4 +1,8 @@
 import {env} from "vscode";
+import {
+    TerminalProfileConfig,
+    workspaceConfigs,
+} from "../vscode-objs/WorkspaceConfigs";
 
 /**
  * The shell families we generate terminal command strings for. Quoting and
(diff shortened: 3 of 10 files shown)


packages/databricks-vscode/src/run/RunCommands.ts
```diff
@@ -244,10 +245,13 @@ export class RunCommands {
             path.join("resources", "python", "dbconnect-bootstrap.py")
         );
         terminal.show();
+        const kind = currentShellKind();
```
rugpanov: **This is a behaviour regression, and I think it needs fixing before merge.**

The quoting is derived from `env.shell` (the *default* shell) but the text is sent to `window.activeTerminal` — whichever terminal the user happens to have focused, running any profile. That's the same generate-for-one-shell/run-in-another mismatch this PR fixes for the two terminals it creates; here it can't be fixed by pinning because the terminal is reused.

What makes it a regression rather than a pre-existing flaw: `main`'s `escapePathArgument` was `` `"${arg}"` `` — double quotes, which parse in **both** cmd and POSIX. So the old line survived a shell mismatch by accident. The new posix branch emits single quotes, which cmd doesn't treat as quoting at all.

Concretely, Windows user whose default profile is Git Bash (`env.shell` = `bash.exe` → `posix`) with a cmd tab active:

```
main : "C:\Python\python.exe" "C:\ext\dbconnect-bootstrap.py" "C:\ws\f.py"   # works in cmd
PR   : 'C:\Python\python.exe' 'C:\ext\dbconnect-bootstrap.py' 'C:\ws\f.py'   # cmd fails
```

cmd looks for a program named literally `'C:\Python\python.exe'`, quote characters included, and fails. The reverse direction (a PowerShell `& ` prefix arriving in cmd) becomes newly possible too.

Options: only reuse `activeTerminal` when it's known to match and otherwise create a pinned one; or — probably cleaner for this command — skip the shell entirely and run via `ProcessExecution`/a task with an argv array, which needs no quoting.

packages/databricks-vscode/src/utils/shellUtils.ts
```diff
@@ -1,23 +1,214 @@
...
+
+/**
+ * Whether `arg` contains a character cmd.exe expands even inside double quotes.
+ *
+ * cmd has no quoting that makes `%` literal in an interactive command line, so
+ * a path such as `C:\p%TEMP%q` cannot be passed through faithfully. Callers that
+ * must not silently corrupt a user-chosen path should check this and surface an
+ * error rather than generating a command that does the wrong thing.
+ */
+export function hasCmdUnsafeChars(arg: string): boolean {
```
rugpanov, line 334: **`hasCmdUnsafeChars` has no callers, so the corruption it's documented to prevent is still reachable.**

The comment directly above says callers "should check this and surface an error rather than generating a command that does the wrong thing" — but grepping the repo, the only references are this definition and one assertion in `shellUtils.test.ts`. `buildBundleInitCommand`, the single caller that hands a user-chosen directory to cmd.exe, never calls it.

So the failure is unchanged from `main`. Output dir `C:\p%TEMP%q\proj` generates:

```
"C:\bin\databricks.exe" bundle init --output-dir "C:\p%TEMP%q\proj"
```

cmd expands `%TEMP%` inside double quotes (there's no quoting that makes `%` literal on an interactive command line), so the CLI scaffolds into a directory the user never chose — and `getSubProjects(parentFolder)` then reports "we haven't detected any Databricks projects in …". Silent wrong-directory, exactly the outcome the doc comment promises to prevent.

Either wire it in — in `bundleInitInTerminal`, `kind === "cmd" && hasCmdUnsafeChars(parentFolder.fsPath)` → surface an error instead of sending — or drop the function along with the claim. As it stands the test reads as coverage but doesn't constrain the shipped path.

Minor, same area and conditional so I'd treat it as optional: the predicate only checks `%`, but under delayed expansion (`cmd /V:ON`, or `DelayedExpansion` set in the registry — not the default, though some corporate images enable it) `!VAR!` expands inside double quotes too, so `/[%!]/` would be more complete. `echoLine`'s cmd branch doesn't escape `%`/`!` either, which means the `Executing:` banner can print a different path than the CLI actually receives — harmless on its own, but it would mislead anyone debugging the above.

rugpanov (commented):
Reviewed the shell-quoting core closely and verified the main claims by executing the generated commands rather than reading them. The POSIX fixes hold up: `'/tmp/it\'s/db'` (old) is a genuine syntax error in `sh` — `unexpected EOF while looking for matching `''` — and the new `'/tmp/it'\''s/db'` runs correctly. `read _` vs bare `read` in dash, the `printf` vs builtin-`echo` difference, the PowerShell interpolation fix, and basename-not-substring matching all check out too. The three-layer test approach is genuinely good, particularly running the *shipped* command string with stdin closed instead of a rewritten variant.

Two things I think should be resolved before merge, both inline. One is a behaviour regression the refactor introduces in `RunCommands`; the other is that `hasCmdUnsafeChars` doesn't actually guard anything yet.

Two smaller notes:

**`shellPath` pinning silently drops the default profile's `args`.** `env.shell` is backed by `getDefaultShell()`, which returns the profile's *path* only. In `resolveShellLaunchConfig`, passing an explicit `executable` makes VS Code build a profile from `{path: e.executable, args: e.args}` — and `e.args` is undefined here, with `_resolveProfile` only expanding variables, never supplying defaults. So a default profile of `{"path": "wsl.exe", "args": ["-d", "Ubuntu-22.04"]}` collapses to bare `wsl.exe` and lands in the **default** distro: `bundle init` scaffolds into a different filesystem than the user configured, and `getSubProjects()` then reports no projects found. Git Bash similarly loses `--login`, though that one looks cosmetic since VS Code's startup `resolveShellEnv` (`["-i","-l","-c"]`) already put the login `PATH` into the inherited env. Forwarding `shellArgs` from the resolved profile would close this.

**Empty `env.shell` makes the kind and the launched shell disagree.** At `shellUtils.ts:53-56` an empty shell resolves to `powershell` on Windows, but the callers deliberately pass `shellPath: undefined` in exactly that case — so VS Code launches `terminal.integrated.defaultProfile.windows`, which may be cmd.exe or Git Bash. When they disagree it's #1822 with the shells swapped: a PowerShell-shaped line sent to cmd fails to parse, so nothing runs — not even the trailing `exit` — and `bundleInitInTerminal` awaits `terminalDidClosePromise` until the user closes the tab manually. Narrow (empty `env.shell` is rare), but deriving the kind from `ComSpec`/the configured default profile would make the two consistent by construction.

No objection to the direction — the classifier-plus-parameterised-helpers shape is the right call, and the diff is easy to follow.

anton-107:
The diagnosis is right and most of the fix is right. #1822 is real, the root-cause framing (branching hidden behind a module-level `env.shell` read) is correct, and `detectShellKind(shell, platform)` as a pure function is the right structural call.

I verified the substantive sub-bugs against real shells rather than taking the description's word, and they hold up:

- Bare `read` in dash → `read: arg count`, exit 2. In the real pattern (`echo ERROR; read; exit`) the error is emitted and the terminal closes, exactly as described.
- zsh's and `/bin/sh`'s builtin `echo` mangle backslashes **inside single quotes**: `echo 'a\class'` prints `a` (truncated), while bash prints it literally. `printf '%s\n'` is verbatim in all three. This is a genuinely non-obvious bug and the round-trip layer earning its keep.
- The POSIX `'…'\''…'` quoting round-trips correctly through zsh, dash and bash with `/tmp/My Dir's "odd" $HOME`.
- Quoting a bare command name (`azBinPath` defaults to `"az"`) still resolves via `PATH` in sh/zsh/dash, so that change is safe.

Both unit-test jobs (Linux + Windows) pass. Six findings below — **1 and 2 I'd want resolved before merge**; the rest can be follow-ups or a description fix.

---

### 1. Pinning `shellPath` silently drops the profile's shell args (blocking)

`BundleInitWizard.ts:163-168` and `AzureCliCheck.ts:375-378` pass `shellPath: shell` with no `shellArgs`. `env.shell` is documented as a *path* only. VS Code's own defaults attach args to it — in the shipped workbench bundle: `path:"zsh",args:["-l"]`, and the auto-detect heuristic `os===2 && name.match(/(zsh|bash)/) ? ["--login"] : []`.

So on macOS these terminals change from a **login shell** to a non-login one, and `~/.zprofile`/`~/.profile` stop being sourced. The exposed case is `AzureCliCheck`: it invokes `az` by bare name and therefore depends on `PATH`, which on macOS is very commonly assembled in `~/.zprofile` (`brew shellenv`). That's a plausible "az: command not found" regression on a path that works today, on a platform the bug being fixed doesn't even affect.

The description presents pinning as pure upside. Either pass the resolved profile's `shellArgs` too, or scope the pin to Windows, where the parse-failure hazard actually lives.

### 2. `hasCmdUnsafeChars` is dead code — the `%` bug it documents is still live (blocking)

The description says it was added "so callers can refuse rather than silently pass the wrong directory." It is referenced **only from tests** (`shellUtils.ts:167` — no production caller). Nothing refuses. On cmd.exe, `C:\Users\me\%USERNAME%proj` still expands to the wrong directory: the exact failure the PR claims to have addressed.

It's also not just unwired — `echoLine`'s cmd branch (`shellUtils.ts:213`) escapes `^&|<>()"` but **not** `%`. Since the banner embeds the output dir, a `%`-containing path corrupts the message too.

Either wire the guard into `BundleInitWizard` (surface an error) or drop it and state the limitation plainly. Shipping an unused guard reads as fixed when it isn't.

### 3. Zero round-trip coverage for cmd.exe and PowerShell — the two shells this PR exists to fix

`shellUtils.test.ts:350` gates every round-trip on `process.platform !== "win32"`, so on the Windows CI runner all ~58 generated round-trip cases **skip**. No test ever executes a generated command through cmd.exe or PowerShell; those two shells have string assertions only, plus one manual VM run.

That's precisely the standard the PR argues against itself: "String equality only proves the code built what the author intended — not that the shell agrees." The `%` gap in finding 2 is the kind of thing a cmd round-trip would have caught. CI already runs `windows-server-latest`, where both shells are present, so this is cheap and available.

Relatedly: `+49` net passing versus "105 of those cases come from the two new suites" implies ~56 don't run where they were counted. Worth a sentence so a reviewer isn't left reconciling it.

### 4. "Behaviour there is unchanged" for `SshCommands` isn't accurate

The reviewer note claims the retained `escapedCliPath` leaves `SshCommands.ts:435` unchanged. It doesn't. Via `CliWrapper.ts:379` the quoting changed for both families:

- PowerShell: `& "path"` → `& 'path'`
- POSIX: the previously broken `'a\'b'` (a backslash is literal inside POSIX single quotes, so the quote closed) → correct `'a'\''b'`

Both are improvements, but SSH tunnel launch is now on a new code path and needs re-testing — the opposite of what the note tells a reviewer.

### 5. The default parameters undercut the PR's central thesis

Every helper is `kind: ShellKind = currentShellKind()`. The claim "`env.shell` is now read in exactly one place" holds only for the function body; at call sites any helper still reaches into VS Code state implicitly. That default is exactly how the `SshCommands` divergence in finding 4 survived the refactor. Dropping the defaults would force each call site to name its shell and would have made that impossible to miss.

### 6. fish is classified `posix` — please verify `read _`

The test suite asserts `/usr/bin/fish → posix`, so fish is a claimed-covered case. In fish, `_` is a read-only variable (the current command name), so `read _` would likely fail, reintroducing the dash bug class for fish users. **I could not verify this** — fish isn't installed in my environment. Worth a check before merge, since the test asserts the classification without exercising it.

---

None of this undermines the core work: the classifier, the `printf` and `read _` fixes, and the POSIX quoting are correct and well-evidenced.

anton-107 (changes requested):
Requesting changes to gate on the two blocking findings from my [detailed review above](https://github.com/databricks/databricks-vscode/pull/2077#issuecomment-5190979927):

1. **Pinning `shellPath` drops the profile's shell args.** `BundleInitWizard.ts:163-168` and `AzureCliCheck.ts:375-378` pass `shellPath` with no `shellArgs`, so on macOS these terminals stop being login shells and `~/.zprofile` is no longer sourced. `AzureCliCheck` invokes `az` by bare name and depends on `PATH`, which `brew shellenv` commonly sets there — a plausible "az: command not found" regression on a platform #1822 doesn't affect. Either pass the profile's `shellArgs` or scope the pin to Windows.

2. **`hasCmdUnsafeChars` is dead code, so the `%` bug it documents is still live.** No production caller (`shellUtils.ts:167`), so nothing refuses; `C:\Users\me\%USERNAME%proj` still expands to the wrong directory on cmd.exe. `echoLine`'s cmd branch (`shellUtils.ts:213`) also doesn't escape `%`, so the banner is corrupted too. Either wire the guard into `BundleInitWizard` or drop it and state the limitation.

Findings 3–6 in that comment (no cmd/PowerShell round-trip coverage, the inaccurate `SshCommands` "unchanged" note, the `currentShellKind()` defaults, and the fish `read _` question) are not blocking — happy for those to be follow-ups or a description fix.

To be clear about the overall assessment: the diagnosis and the core of this fix are correct. The classifier, the `printf` and `read _` fixes, and the POSIX quoting all hold up — I verified them against real shells. This is a gate on two specific items, not on the approach.

abdelrahmanomar1515:
Thanks — all four hold up. I verified each against VS Code's own source rather than
just reasoning about it, and pushed fixes in 80e4cb19.

**1. `RunCommands` regression.** Confirmed, and the point about *why* it's a regression
rather than a pre-existing flaw is the key one: `main`'s `"${arg}"` parsed in both cmd
and POSIX, so the mismatch survived by accident. Single quotes don't — `& '/bin/echo' 'hi'`
is a hard syntax error, and cmd looks for a program named literally
`'C:\Python\python.exe'`, quotes included.

Added `terminalShellKind()`, which classifies from the terminal's own
`creationOptions.shellPath` (public API) and falls back to the ambient shell only for
inherited-profile and extension pseudoterminals, which genuinely have no shell of their
own. I did *not* take the `ProcessExecution`/argv suggestion: it's the cleaner long-term
shape, but it changes user-visible behaviour (output would stop landing in the terminal
the user is looking at), so that felt like a separate decision rather than something to
fold into this PR.

**2. `hasCmdUnsafeChars` unwired.** Correct — grep confirmed the only references were the
definition and one assertion, so the failure was unchanged from `main`. Now wired via
`unsafeOutputDirReason()`, which `bundleInitInTerminal` surfaces as an error *and*
`buildBundleInitCommand` enforces by throwing, so the check can't be skipped by a future
caller. Also broadened to `!` per your addendum.

**3. `shellPath` pinning drops the profile's `args`.** This was the subtlest one and it
checks out exactly as described. From VS Code source: `env.shell` is
`getDefaultShell()` → `profile?.path` only; `mainThreadTerminalService` maps
`shellPath`→`executable`; and `_resolveProfile` only expands variables, never supplying
args. So `{path: "wsl.exe", args: ["-d", "Ubuntu-22.04"]}` really does collapse to the
default distro. Both call sites now forward `defaultProfileShellArgs()`, gated on the
profile path matching the shell being pinned — a renamed or `source`-based profile can
point at a different executable, and passing another shell's args would be worse than
passing none.

**4. Empty `env.shell` disagreement.** Agreed it's narrow, but the fix was cheap:
`currentShellKind()` now resolves through `effectiveShellPath()`
(`env.shell` → configured default profile → `ComSpec`), so the kind and the shell VS Code
actually launches agree by construction.

The terminal settings reads go through `vscode-objs/WorkspaceConfigs` per
CODE_CONVENTIONS §5, which also keeps the new logic pure and unit-testable.

### Do the new tests actually bite?

Reverted each fix in turn and recorded the failures:

| Fix reverted | Tests failing |
|---|---|
| `terminalShellKind` → `env.shell` | 3 |
| cmd `%`/`!` guard unwired | 2 |
| `shellArgs` forwarding dropped | 3 |
| empty-shell profile fallback | 2 |
| `!` detection | 2 |

`terminalShellKind` takes `platform` as a parameter so the cmd.exe cases run on
macOS/Linux CI rather than being skipped off Windows.

**572 passing, 0 failing**, `test:lint` clean. Worth noting for anyone reproducing: the
6 `CliWrapper` failures I initially saw were only a missing `bin/databricks` in a fresh
worktree — `yarn package:cli:fetch` fixes them, and they were unrelated to these changes.

packages/databricks-vscode/src/bundle/bundleInitCommand.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    return undefined;
+}
+
+/**
+ * The full `databricks bundle init` line sent to the wizard terminal: clear the
+ * screen, explain what is running, run the CLI, then hold the terminal open
+ * until the user acknowledges so any CLI error stays readable.
+ *
+ * Kept in its own module, free of `vscode` imports, so it stays a pure function
```
rclarey, line 40: nit: I don't think this needs to be in it's own file

packages/databricks-vscode/src/bundle/bundleInitCommand.ts
```diff
@@ -0,0 +1,75 @@
...
+ * silently rewrite it.
+ */
+export function buildBundleInitCommand(
+    escapedCliPath: string,
+    outputDir: string,
+    kind: ShellKind
+): string {
+    const reason = unsafeOutputDirReason(outputDir, kind);
+    if (reason !== undefined) {
+        throw new Error(reason);
```
rclarey, line 56: Is this properly caught anywhere?

rclarey (approved):
A couple comments but otherwise LGTM

abdelrahmanomar1515:
Closing this PR in favor of #2092

PR #2080, opened 2026-08-05
fix(python-setup): pass workspace auth to the setup-local CLI

rugpanov:
## Changes

`environments setup-local` resolves its own authentication via `MustWorkspaceClient`, but the extension spawned it with a bare `process.env` and never told it which workspace we are connected to. The CLI therefore fell back to its own default-profile resolution.

Two user-visible consequences:

- **Wrong-workspace provisioning, silently.** A user on a non-default profile provisioned against a different workspace's cluster, so the environment was matched to the wrong DBR and constraints — with no error.
- **Hard failure on a relocated config.** A user who moved `.databrickscfg` (via `databricks.databrickscfgLocation`) failed outright, because `DATABRICKS_CONFIG_FILE` was never propagated, even though the extension itself was connected fine.

### Commits

**1. Forward the auth env** — adds `CliWrapper.getSetupLocalEnvVars`, alongside the existing `getBundleInitEnvVars` / `getSshConnectEnvVars`: `getEnvVarsForCli` (with the configured `.databrickscfg` location) + proxy + logging + `authProvider.toEnv()`. `PythonSetupCliClient` takes an injected `EnvFn` rather than reading `process.env` at the spawn site, keeping the gateway free of a `vscode` import and making the wiring assertable.

**2. Drop the unused `--profile` field** — `SetupLocalInvocation.profile` had no caller. With auth on the environment, a flag would be a redundant second source of truth that could disagree with it, so it is removed and a test pins that `--profile` is never emitted.

### Notes for reviewers

Three details are deliberate rather than incidental:

- **The `process.env` spread lives at the call site, not in the method.** Every `*EnvVars` method in `CliWrapper` returns only the curated Databricks vars and lets the caller decide (`BundleInitWizard` spreads the ambient env; `SshCommands` deliberately does not). This follows that convention.
- **Databricks vars are applied last.** `setup-local` shells out to `uv`, which needs the real OS environment (platform paths, `UV_*`) as its base — but overlaying the Databricks vars on top means a stale ambient `DATABRICKS_CONFIG_PROFILE` cannot redirect the run.
- **`DATABRICKS_OUTPUT_FORMAT` stays `json`.** The bundle-init and ssh-connect flows override it to `text` because they render into a terminal for a human; this run's stdout is parsed as the single JSON result object, so it must not.

The env is resolved per run rather than at construction, so switching workspace or profile mid-session is picked up without a reload.

## Tests

- `yarn tsc --noEmit` clean; `yarn test:lint` clean (eslint + prettier).
- `yarn test:unit`: **474 passing, 6 failing.** Those 6 (`CliWrapper` `listProfiles`) also fail on unmodified `main` — confirmed by stashing and re-running the baseline, which gives 471 passing / the same 6 failing. They shell out to a real CLI binary that is absent from the worktree. So all new tests pass and nothing regressed.
- New coverage: the env from the injected function reaches `spawn`; it is re-read on every run; `getSetupLocalEnvVars` forwards profile + host; and `--profile` is never emitted.

Not covered by automated tests: an end-to-end run against a real workspace on a non-default profile. That needs a bundled CLI that can fetch constraints, and is tracked separately.

### DIFF
--- a/packages/databricks-vscode/src/cli/CliWrapper.test.ts
+++ b/packages/databricks-vscode/src/cli/CliWrapper.test.ts
@@ -347,6 +347,29 @@ token = dapitest5678
             throw e;
         }
     });
+
+    it("should forward auth to the setup-local env vars", async () => {
+        const logFilePath = getTempLogFilePath();
+        const cli = createCliWrapper(logFilePath);
+        const authProvider = new ProfileAuthProvider(
+            new URL("https://test.com"),
+            "PROFILE",
+            cli,
+            true
+        );
+
+        const env = cli.getSetupLocalEnvVars(authProvider);
+
+        // The two vars this exists for: the CLI resolves auth itself, so the
+        // profile and host must arrive via the environment.
+        assert.equal(env.DATABRICKS_CONFIG_PROFILE, "PROFILE");
+        assert.equal(env.DATABRICKS_HOST, "https://test.com/");
+        // Inherited from getEnvVarsForCli and left alone: it agrees with the
+        // explicit `--output json` on the argv that the result parser needs.
+        // The bundle-init/ssh-connect flows override this to "text" because they
+        // render CLI output to a terminal; this flow must not.
+        assert.equal(env.DATABRICKS_OUTPUT_FORMAT, "json");
+    });
 });
 
 describe("waitForProcess", () => {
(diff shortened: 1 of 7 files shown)


rugpanov:
🤖 Integration tests triggered for `9c36bff8` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31009292741)

misha-db (approved):

rugpanov:
🤖 Integration tests triggered for `5a0dfbd9` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31017377194)

PR #2093, opened 2026-08-06
Fix show bundle logs in Cursor

misha-db:
## Changes

Problem

LoggerManager.showOutputChannel() called LogOutputChannel.show(), which silently did nothing on hosts that register a log channel under a different id than the extension host derives for it. The extension host computes the channel id locally (<extensionId>.<sanitized name>) and never round-trips it through the workbench, so the two can disagree — some hosts append a workspace scope segment (.workspaceId-<id>). When they disagree, .show() targets an id that isn't registered and the channel never appears.

User-visible effect: Databricks: Show Bundle Logs (databricks.bundle.showLogs), the databricks.internal.showOutput command behind the AI-tools "Show Logs" affordance, and the "Show Error Logs" button in CliWrapper's error notification all appeared to do nothing.

Fix

New src/logger/outputChannelReveal.ts recovers the real reveal command by probing the registered command list for workbench.action.output.show.<channelId>:

- getLogChannelId() derives the id the way the extension host does, stripping the same illegal characters.
- pickRevealCommand() returns a scoped command only when the host registered the channel under a different id — if the exact id is present, it returns undefined so plain .show() stays in charge. Candidates must differ by exactly one dot-segment, which keeps a longer channel name whose id merely prefixes ours out of the set. On ambiguity it prefers the .workspaceId- shape and otherwise gives up rather than reveal an unrelated channel.

This is a capability probe rather than a host/fork-name check, so it degrades to existing behavior on hosts where ids already agree.

LoggerManager changes:

- showOutputChannel() is now async and resolves the reveal command before falling back to channel.show(). It never rejects — any failure is logged at debug level and falls through to .show().
- Resolved command ids are cached per channel; a failed executeCommand evicts the cache entry so a stale id doesn't break reveals for the rest of the session.
- When a channel was just created, one 200 ms retry covers createOutputChannel registering with the workbench asynchronously.
- The ad-hoc "Databricks Logs" | "Databricks Bundle Logs" union is extracted to an exported LogChannelName type and now types the outputChannels map.

CliWrapper awaits the newly-async call.

## Tests

outputChannelReveal.test.ts
LoggerManager.test.ts

### DIFF
--- a/packages/databricks-vscode/src/logger/LoggerManager.test.ts
+++ b/packages/databricks-vscode/src/logger/LoggerManager.test.ts
@@ -6,7 +6,11 @@ import {tmpdir} from "os";
 import path from "path";
 import {instance, mock, when} from "ts-mockito";
 import {ExtensionContext, Uri} from "vscode";
-import {LoggerManager, Loggers} from "./LoggerManager";
+import {LoggerManager, Loggers, RevealCommandApi} from "./LoggerManager";
+
+const EXTENSION_ID = "databricks.databricks";
+const CHANNEL = "Databricks Logs" as const;
+const SCOPED_REVEAL = `workbench.action.output.show.${EXTENSION_ID}.${CHANNEL}.workspaceId-abc123`;
 
 describe(__filename, function () {
     let tempDir: string;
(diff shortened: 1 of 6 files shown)


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `3bc805ff`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31163771142)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `837779e6` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31583455425)

rclarey (approved):

packages/databricks-vscode/src/logger/outputChannelReveal.ts
```diff
@@ -0,0 +1,76 @@
+import {commands} from "vscode";
+
+/**
+ * Recovers the command that reveals a `LogOutputChannel` on hosts that register
+ * the channel under a different id than the extension host believes it has.
+ *
+ * See ./README.md for why this is needed and why the check is shaped as a
```
rugpanov: **[non-blocking]** Dangling doc reference — this points to `./README.md`, but no README is added in this PR and `src/logger/` doesn't have one. Either add the file or drop/inline the explanation.

packages/databricks-vscode/src/logger/outputChannelReveal.ts
```diff
@@ -0,0 +1,76 @@
...
+            candidates.push(command);
+        }
+    }
+
+    if (candidates.length === 1) {
+        return candidates[0];
+    }
+    // Ambiguous: prefer the known workspace-scope shape, and otherwise give up
+    // rather than reveal some unrelated channel of ours.
+    const workspaceScoped = candidates.filter((command) =>
```
rugpanov, line 66: **[non-blocking]** Case-handling inconsistency: candidate matching is case-insensitive (the prefix is lowercased on line 36), but this workspace-scope tiebreaker runs `startsWith(".workspaceId-")` against the original-case slice. If a host registered the scope with different casing (e.g. `.WorkspaceId-`) *and* there were multiple candidates, the real match would be filtered out and the function would give up. Doubly-rare — the single-candidate path returns early above, and real Cursor uses lowercase — but lowercasing the slice here would make it uniform.

packages/databricks-vscode/src/logger/LoggerManager.ts
```diff
@@ -157,8 +172,57 @@ export class LoggerManager {
...
+        const extensionId = this.context.extension.id;
+        let command = await findRevealCommand(extensionId, name);
+        if (command === undefined && justCreated) {
+            await new Promise((resolve) =>
+                setTimeout(resolve, REVEAL_COMMAND_RETRY_DELAY_MS)
+            );
+            command = await findRevealCommand(extensionId, name);
+        }
+
+        if (command !== undefined) {
```
rugpanov, line 257: **[non-blocking]** `undefined` (ids agree — the normal VS Code case) is never cached, so every `showOutputChannel` call re-runs `findRevealCommand` → `commands.getCommands(true)` and enumerates all registered commands. It's a user-initiated, infrequent action so the cost is negligible; caching a 'no command needed' sentinel would avoid re-scanning each time.

rugpanov (commented):
Multi-engine review (Isaac Review + Codex + a Claude reviewer, then cross-checked against the source by hand). No confirmed correctness or security bugs — the probe logic is sound, the exact-id short-circuit correctly avoids shadowing `.show()`, and the whole thing fails safe. **Every finding below is non-blocking** (nits + one test-coverage gap).

One thing worth confirming before merge that can't be verified statically: the fix assumes the host registers the channel id as `${extensionId}.${name}` with spaces preserved (`getLogChannelId`, outputChannelReveal.ts:22). If Cursor/VS Code derives the id differently (e.g. transforms spaces), `pickRevealCommand` matches nothing and silently falls back to the same broken `.show()` — i.e. the fix fails safe but does nothing. The added unit test can't catch this because it encodes the same assumption. Presumably already validated on a real Cursor session; just flagging it since it's the crux of whether the fix works.

packages/databricks-vscode/src/logger/LoggerManager.ts
```diff
@@ -157,8 +172,57 @@ export class LoggerManager {
         env.openExternal(this.context.logUri);
     }
 
-    showOutputChannel(name: "Databricks Logs" | "Databricks Bundle Logs") {
-        this.getLogOutputChannel(name).show();
+    /**
+     * Reveals a log output channel. Never rejects: any failure to resolve the
```
rugpanov: **[non-blocking]** Minor: "Never rejects" is slightly stronger than the code guarantees — `getLogOutputChannel()` (line 182) and the fallback `channel.show()` (line 198) both run outside the try/catch, so a throw from either would still reject. In practice these VS Code APIs don't throw and the `extension.ts` callers are fire-and-forget, so this is theoretical — just noting the comment overstates it.

packages/databricks-vscode/src/logger/LoggerManager.test.ts
```diff
@@ -17,6 +17,17 @@ describe(__filename, function () {
         tempDir = await mkdtemp(path.join(tmpdir(), "testdir-"));
     });
 
+    it("should reveal an output channel without throwing", async () => {
```
rugpanov: **[non-blocking]** The `outputChannelReveal.ts` helpers are covered thoroughly, but this is the only test touching the new `LoggerManager` orchestration and it asserts nothing beyond "does not throw" — command execution, per-channel caching, cache eviction on `executeCommand` failure, the 200 ms `justCreated` retry, and the `.show()` fallback all go unverified. Note the retry (`justCreated`) branch is only reachable in production when `loggingEnabled` is false, since otherwise `initLoggers()` pre-creates both channels at activation (extension.ts:282-283) — so that path has no coverage at all today.

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `d5b7e7bd` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31592830420)

rugpanov:
🤖 Integration tests triggered for `04a1160e` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31594866727)

PR #2094, opened 2026-08-10
fix(python-setup): show red icon when setup is not done

rugpanov:
## Why

The uv-native **Set up Python environment** entry rendered its call-to-action icon in green (`debugIcon.startForeground`) while its `contextValue` was `databricks.environment.pythonSetup.error`. The green icon read as "ready" before setup had ever run — contradicting both the contextValue and the sibling checklist entries.

This is defect #2 from the M4 bug bash (DECO-27917): *"Entry Appears Ready Before Setup Runs — Call-to-action renders green before provisioning occurs; internal inconsistency: contextValue shows `...pythonSetup.error` while icon is green."*

## What

- `pythonSetupEntry.ts`: color the not-ready rocket icon with `errorForeground` (red) instead of `debugIcon.startForeground` (green). This matches the convention already used by the other Python Environment and cluster entries, where required-but-not-done steps are red and only ready/available states are green. The ready state (green `check`) is unchanged.
- `pythonSetupEntry.test.ts`: assert the not-ready icon color is `errorForeground` so it can't silently regress to green.

## Verification

- Unit suite green: **507 passing, 0 failing**.
- ESLint + Prettier clean on both files.

### DIFF
--- a/packages/databricks-vscode/src/ui/configuration-view/pythonSetupEntry.test.ts
+++ b/packages/databricks-vscode/src/ui/configuration-view/pythonSetupEntry.test.ts
@@ -1,5 +1,5 @@
 import {expect} from "chai";
-import {ThemeIcon} from "vscode";
+import {ThemeColor, ThemeIcon} from "vscode";
 import {buildPythonSetupEntry} from "./pythonSetupEntry";
 
 describe("buildPythonSetupEntry", () => {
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `b412e36b`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31375899415)

rclarey (approved):

PR #2095, opened 2026-08-10
test(e2e): win the sub-second cancel race in run_files

anton-107:
## Changes

The nightly `run_files > should cancel a run during deployment` spec fails
intermittently on Linux (most recently in `vscode-isolated-nightly` run
31358950776, and on ~1 night in 3).

The automated triage read this as "the notification never appeared", but the wdio
log in the job's `test-logs` artifact shows the opposite — the toast appeared, and
Cancel was found *and clicked*:

```
05:42:54.105  RESULT Uploading bundle assets            <- toast found
05:42:54.106  findElement(".//a[@role='button' and @title='Cancel']")
05:42:54.127  elementClick(element_158)                 <- Cancel found and clicked
05:42:54.373  request failed: stale element reference   <- toast disposed mid-click
```

`bundle sync` completed in **0.93s** (CLI log, pid 4988: `05:42:53.26` ->
`05:42:54.19`, exit 0), so VS Code tore the toast down between the `findElement`
and the `elementClick`. After that the poll loop had nothing left to retry and
spun out the remaining 59s on an empty notification list (`Notifications: 0`).
The recorded video confirms the run then completed normally
(`Done (took 10516ms)`).

The cancellable window is short because the upload is *fast*, not because the
test is slow — so longer timeouts and click retries (#2056) can't fix it. Instead:

- **Find and click in a single `browser.execute`**, so the lookup and the click
  are one synchronous JS turn that cannot go stale in between. This also drops
  two round-trips (the toast message read, then a separate find — the message
  read alone cost ~135ms of a sub-second window), and it returns immediately when
  the button is absent, whereas `element.click()` blocks on wdio's 10s implicit
  wait and would make a tight poll interval a no-op.
- **Retry the whole run, not just the click**: if the upload wins the race, start
  another run and race the next upload (3 attempts).
- **Clear the debug console between attempts.** A lost race completes a run and
  prints `hello world` — exactly what the next test in this spec waits for, on
  the same shared console.

Two guards, since a test that always fails would be worse than a flake:

- The final attempt falls back to wdio's element click (a real input event), so
  if a synthetic DOM click ever stops reaching VS Code's handler this stays a
  race we can lose rather than a hard failure.
- `clearReplAction` is a `ViewAction` resolved via `getActiveViewWithId`, so it
  silently no-ops unless the console is the active view — the helper focuses the
  panel first, and only logs (never throws) if clearing fails.

## Tests

Verified against the pinned VS Code 1.86.0 source that the DOM assumptions hold:

- The action button is an `<a>` with `role="button"` and `title="Cancel"`
  (`base/browser/ui/button/button.ts`) — same element the wdio-vscode-service
  locator targets, just scoped to `.notifications-toasts` and expressed in CSS.
- `.notifications-toasts` is the outermost toast container, with each
  `.notification-toast` nested inside
  (`browser/parts/notifications/notificationsToasts.ts`).
- The button binds a plain `click` listener, which `HTMLElement.click()`
  dispatches — so the synthetic click does reach the handler.
- `workbench.debug.panel.action.clearReplAction` exists in 1.86 and is a
  `ViewAction` (`contrib/debug/browser/repl.ts:973`), hence the focus-first
  ordering above.

`prettier --check` and `tsc -p src/test/e2e/tsconfig.json --noEmit` are clean for
this file. (Two pre-existing `tsc` errors in `utils/unityCatalogUtils.ts` are
untouched by this change.)

Not yet exercised by a real run — the e2e suite can't run in my dev environment,
so this needs a CI run on the `run_files` shard. Since the failure reproduces on
roughly one night in three, a single green run won't be conclusive; I'd suggest a
couple of `vscode-isolated-pr` runs, and ideally one confirming the retry path
itself works (the `Cancel attempt N of 3` logging makes that visible).

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/run_files.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/run_files.e2e.ts
@@ -14,6 +14,30 @@ import {
     writeRootBundleConfig,
 } from "./utils/dabsFixtures.ts";
 
+/**
+ * Empty the debug console. The whole spec shares one console, so output from an
+ * earlier run would otherwise satisfy a later wait.
+ */
+async function clearDebugConsole() {
+    try {
+        await browser.executeWorkbench(async (vscode) => {
+            // clearReplAction is a ViewAction: it resolves the REPL through
+            // getActiveViewWithId and silently does nothing when the console
+            // isn't the active view. Focus it first so the clear really happens.
+            await vscode.commands.executeCommand(
+                "workbench.panel.repl.view.focus"
+            );
+            await vscode.commands.executeCommand(
+                "workbench.debug.panel.action.clearReplAction"
+            );
+        });
+    } catch (e) {
+        // Hygiene, not an assertion — don't fail the run over it, but say so,
+        // because a stale console can make a later wait pass on old output.
+        console.log("Failed to clear the debug console:", e);
+    }
+}
+
 describe("Run files", async function () {
     let projectDir: string;
     this.timeout(6 * 60 * 1000);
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests ❌ **3 of 35 test jobs failed** for `2fb77289` (32 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31381485524)

anton-107:
## CI: the target spec passes; the 3 red shards are unrelated infra flakes

`run_files` — the spec this PR fixes — **passed on both OSes**, each winning on the
first attempt:

```
LNX  ✓ should cancel a run during deployment    3 passing (1m 13.2s)   Cancel attempt 1 of 3
WIN  ✓ should cancel a run during deployment    3 passing (1m 27.2s)   Cancel attempt 1 of 3
```

The 3 failing shards in [run 31381485524](https://go/deco-tests/31381485524) are all
environmental, and each of the three specs **also passed on the other OS in the same
run** — the signature of infra flake rather than a code defect. This PR touches exactly
one file (`src/test/e2e/run_files.e2e.ts`) and cannot affect any of them:

| Spec | OS | Failure |
|---|---|---|
| `refresh_resource_explorer_on_yml_change` | Win | `curl: (52) Empty reply from server` -> exit 52. Died in setup while downloading a dependency; **zero tests ran**. |
| `bundle_init` | Win | Browser session lost mid-run: `no such window: target window already closed` / `Connection closed. Code: 1006`. 3 tests failed, all after the crash. |
| `run_dbconnect.ucws` | Linux | `should select serverless compute`: `waitUntil condition timed out after 10000ms`. Other 4 passed, incl. the actual dbconnect run. |

Two notes on reading these:
- The `bundle_init` follow-on error (`Unable to locate the bundle root: databricks.yml
  not found`) is teardown failing *after* the session died, not a real assertion.
- The `run_dbconnect.ucws` shard also logged the known `[dbconnect-fallback]` path.

### Caveat on the green result

Worth stating plainly: this run is *consistent with* the fix working, but it is not
proof. The old test also usually passed — it failed on roughly 1 night in 3. Both
shards won on `Cancel attempt 1 of 3`, so the retry path and the attempt-3
real-input-event fallback are still unexercised, and because `test-logs` artifacts are
only uploaded on failure I can't time the click against the upload window on this run.

A couple of extra `vscode-isolated-pr` runs would firm this up: either it stays green,
or a `Cancel attempt 2` appears and demonstrates the retry actually works.

anton-107:
## Re-run is fully green — the 3 earlier failures were infra noise

[Run 31422992256](https://go/deco-tests/31422992256) (same commit `2fb7728`,
verified via the checkout job's `PULL_REQUEST_NUMBER: 2095` / `COMMIT_SHA`):
**0 failing shards out of 32.** The `Integration Tests` check is now green.

All three specs that failed in the first run passed this time without any change to
them, confirming the earlier read that they were environmental:

| Spec | First run | Re-run |
|---|---|---|
| `refresh_resource_explorer_on_yml_change` (Win) | `curl: (52)`, exit 52 | pass |
| `bundle_init` (Win) | `no such window: target window already closed` | pass |
| `run_dbconnect.ucws` (Linux) | `waitUntil timed out after 10000ms` | pass |

`run_files` passed again on both OSes:

```
LNX  ✓ should cancel a run during deployment    3 passing (1m 30.3s)   Cancel attempt 1 of 3
WIN  ✓ should cancel a run during deployment    3 passing (1m 28.1s)   Cancel attempt 1 of 3
```

### What is and isn't proven

4/4 shards across two runs won on **attempt 1**, so the single-`browser.execute`
find-and-click is reliably catching the toast inside the sub-second window — that is
the actual fix, and it is now the part with real evidence behind it.

Still unexercised, and I'd rather say so than imply full coverage:

- the **retry loop** (no run has needed attempt 2),
- the **attempt-3 real-input-event fallback**,
- **`clearDebugConsole()`** as a correctness guard — it runs every attempt, and no
  `Failed to clear the debug console` appeared in any shard, so it isn't throwing; but
  since no attempt was ever repeated, it never had stale output to suppress.

Those paths are defensive: they exist so that if the upload ever wins the race, the
test recovers instead of failing. Their value is precisely that they should stay dormant
when the primary path works. Worth keeping in mind if this spec is ever touched again —
a green run is not evidence they work.

For context on the earlier flakes: a neighbouring PR's run in the same window
(31383452231, PR #2097) failed on `unity_catalog` Windows with `Children of "system" did
not load and stabilise` — a fourth distinct spec. E2E infra was broadly noisy this
morning; the last three nightlies before it were clean at 0 failures.

anton-107:
## CI status: all checks green

Summarising where this landed, since the earlier red `Integration Tests` X may still be
in reviewers' notifications:

| Check | State |
|---|---|
| Integration Tests ([run 31422992256](https://go/deco-tests/31422992256)) | **pass** — 0 failing shards / 32 |
| Run unit tests (linux_amd64) | pass |
| Run unit tests (windows_amd64) | pass |
| Package Arm64 VSIX | pass |
| Check secrets access | pass |

Nothing is failing on this PR. The one earlier red `Integration Tests` run
(31381485524) failed on three specs — `refresh_resource_explorer_on_yml_change`,
`bundle_init`, `run_dbconnect.ucws` — all of which passed on re-run with no change to
them, and each of which also passed on the *other* OS within that same run. Details in
the two comments above.

`run_files`, the spec this PR fixes, has now passed **4/4 shards across two runs** (both
OSes each time), every one winning on `Cancel attempt 1 of 3` — i.e. the
single-`browser.execute` find-and-click is catching the toast inside the sub-second
window that the previous implementation kept losing.

Ready for review. Two caveats I'd rather restate than let a green tick paper over:

1. The retry loop, the attempt-3 real-input-event fallback, and `clearDebugConsole()`'s
   stale-output suppression are all still unexercised — no run has needed a second
   attempt. They're deliberately dormant safety nets, so green CI is not evidence they
   work.
2. The underlying flake was intermittent (~1 night in 3 on `vscode-isolated-nightly`),
   so two green runs reduce but do not eliminate the possibility of a rarer failure mode.
   The nightly is the long-run signal worth watching after merge.

misha-db (approved):

PR #2096, opened 2026-08-10
fix(cli): warn when the bundled CLI is stale in a dev checkout

anton-107:
## Changes

The extension shells out to a Databricks CLI bundled at
`packages/databricks-vscode/bin/databricks`, pinned by `cli.version` in
`packages/databricks-vscode/package.json`. `bin/` is gitignored, so the binary is
fetched once by `package:cli:fetch` and a later `git pull` that bumps
`cli.version` leaves the old one in place.

The resulting failure is opaque. A stale CLI rejects a subcommand the extension
expects — `unknown command "aitools"` for a checkout predating #2025 — which
throws inside `AiToolsManager.initialize` and aborts activation before
`databricks.context.initialized` is set. The configuration view then sits on
"Initializing..." indefinitely, with nothing pointing at the CLI as the cause.
This cost real debugging time on a 4-month-old checkout.

This PR compares the bundled binary's reported version against the pinned one at
startup and warns, naming the command to fix it:

> The bundled Databricks CLI is v0.297.2, but this checkout pins v1.11.0. Run
> "yarn workspace databricks run package:cli:fetch" and reload the window.

Design notes:

- **Dev-only** (`isDevExtension()`). A packaged extension fetches its CLI during
  the build, so the two versions can't diverge there.
- **Warn-only, not blocking.** A mismatched CLI is usually still usable; failing
  activation would be worse than the warning.
- **Not awaited** in `activate()`, so activation doesn't wait on spawning the CLI.
- Lives beside `checkArchCompat` in `packageJsonUtils.ts`, which is the same shape
  of startup guard.

Also documents the devloop in `AGENTS.md` (re-fetching the CLI after a pull, lazy
activation, the marketplace extension ID collision, and where the logs live), and
notes in `CONTRIBUTING.md` that the CLI fetch is not a one-time step.

## Tests

Added unit tests in `packageJsonUtils.test.ts` covering: reading the reported
version, a missing binary, a stale CLI, a matching CLI, a non-dev checkout, and an
unknown pinned version. The stub-CLI cases are skipped on Windows, where a shell
script isn't executable. Also updated the existing `getMetadata` test, which
asserts with `deepEqual` and so needed the new `cliVersion` field.

Verified manually that the warning fires for the real case that prompted this —
a v0.297.2 binary in a checkout pinning v1.11.0 — and that it stays silent once
`package:cli:fetch` has run.

⚠️ **Not yet run through the repo's test runner.** My local checkout has stale
`node_modules` (TypeScript 5.3.3 where the repo now pins ^6.0.3, and a mocha whose
bundled yargs breaks on Node 26), and npmjs is unreachable from this machine, so
`yarn workspace databricks run test:unit` could not execute. I verified the logic
by compiling the module and exercising it directly — all six cases pass — but the
suite should be run on fresh dependencies before merging.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,13 +37,7 @@ yarn workspace databricks run package:cli:fetch
 ```
 
 Re-run this whenever `cli.version` in `packages/databricks-vscode/package.json`
-changes — after every `git pull`, in practice. `bin/` is gitignored, so a pull
-that bumps the pinned version never updates the binary you already have, and a
-stale CLI rejects subcommands the extension expects. That failure is easy to
-misread: the rejected subcommand throws during activation, so the configuration
-view sits on "Initializing..." indefinitely instead of reporting an error. In a
-development checkout the extension compares the two versions on startup and
-warns when they differ, naming this command.
+changes — after every `git pull`, in practice.
 
 After that you're ready to build, run, and test the extension.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,12 @@ Fetch the Databricks CLI that the extension bundles:
 yarn workspace databricks run package:cli:fetch
 ```
 
+Re-run this whenever `cli.version` in `packages/databricks-vscode/package.json`
+changes: `bin/` is gitignored, so a `git pull` never updates the binary, and a
+stale CLI rejects subcommands the extension expects — which surfaces as a
+configuration view stuck on "Initializing..." rather than as an error. The
+extension warns about a version mismatch on startup in a dev checkout.
+
 After that you're ready to build, run, and test the extension.
 
 ## Building and running
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,10 +37,13 @@ yarn workspace databricks run package:cli:fetch
 ```
 
 Re-run this whenever `cli.version` in `packages/databricks-vscode/package.json`
-changes: `bin/` is gitignored, so a `git pull` never updates the binary, and a
-stale CLI rejects subcommands the extension expects — which surfaces as a
-configuration view stuck on "Initializing..." rather than as an error. The
-extension warns about a version mismatch on startup in a dev checkout.
+changes — after every `git pull`, in practice. `bin/` is gitignored, so a pull
+that bumps the pinned version never updates the binary you already have, and a
+stale CLI rejects subcommands the extension expects. That failure is easy to
+misread: the rejected subcommand throws during activation, so the configuration
+view sits on "Initializing..." indefinitely instead of reporting an error. In a
+development checkout the extension compares the two versions on startup and
+warns when they differ, naming this command.
 
 After that you're ready to build, run, and test the extension.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,53 +7,18 @@ links — in code, comments, commit messages, PR titles, or PR descriptions. The
 direction is one-way: link the PR from the internal ticket, never the ticket from the
 PR. A handful of `DECO-*` mentions predate this rule and are not precedent.
 
-## Running the extension from source
+## Re-fetch the CLI after pulling
 
-Full setup and test instructions live in [CONTRIBUTING.md](CONTRIBUTING.md). The
-points below are the ones that cost time when they're missed.
-
-**Re-fetch the CLI after pulling.** The extension shells out to a Databricks CLI
-bundled at `packages/databricks-vscode/bin/databricks`, pinned by `cli.version` in
-`packages/databricks-vscode/package.json`. `bin/` is gitignored, so a `git pull`
-that bumps `cli.version` leaves the old binary in place:
+Setup, build, and test instructions live in [CONTRIBUTING.md](CONTRIBUTING.md). The
+step easiest to miss: the extension shells out to a Databricks CLI bundled at
+`packages/databricks-vscode/bin/databricks`, and `bin/` is gitignored, so a `git pull`
+that bumps `cli.version` leaves the old binary in place. Re-fetch it:
 
 ```sh
 yarn workspace databricks run package:cli:fetch
 ```
 
-A stale CLI rejects subcommands the extension expects and aborts activation, which
-looks like a configuration view stuck on "Initializing..." rather than an error.
-The extension warns about this on startup in a dev checkout.
-
-**Launching.** Press `F5` in VS Code (`Run and Watch Extension`) — it builds and sets
-`EXTENSION_DEVELOPMENT` for you. To launch from a terminal instead:
-
-```sh
-cd packages/databricks-vscode
-EXTENSION_DEVELOPMENT=true code --disable-extension databricks.databricks \
-    --extensionDevelopmentPath="$PWD" --new-window <a-bundle-project>
-```
-
-`--disable-extension databricks.databricks` avoids an ID collision with an installed
-marketplace build. Note that `code` may fold `--new-window` into an existing window
-instead of opening the folder you asked for; check the window's title bar says
-`[Extension Development Host]` and that it's on the folder you intended.
-
-**Activation is lazy.** `activationEvents` has no `*`, so the extension only starts
-for a folder containing a `databricks.yml` (or an open `.py` file) — an empty folder
-activates nothing. Point it at a project whose target sets `workspace.host`.
-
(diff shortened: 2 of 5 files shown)


rugpanov:
🤖 Integration tests triggered for `3cf99e99` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31381876691)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `e75e737a`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31383064758)

anton-107:
Reviewing my own PR here, so treat this as self-review notes for whoever picks it up rather than an approval. I went back over it with fresh eyes and verified the risky assumptions; the logic holds up, but **the description is now wrong and needs rewriting before merge.**

### The description no longer matches the code

It says:

> The stub-CLI cases are skipped on Windows, where a shell script isn't executable.

There is no stub CLI and no Windows skip in the final diff — commit e75e737 rewrote the tests to exercise the real bundled binary, and the only `win32` reference left is the `.exe` path branch. It also still carries the "⚠️ Not yet run through the repo's test runner" caveat. Both need updating so a reviewer isn't reading about a different implementation.

### What I did verify

The assumption most likely to make this whole feature silently useless is the `version --output json` contract, so I checked both ends:

- **Current CLI:** ran `databricks version --output json` on v1.11.0 — key is `Version` (capital V), as the code assumes.
- **The stale CLI this exists to catch:** read `cmd/version/version.go` at tag `v0.297.2` in databricks/cli. It renders the same `build.GetInfo()` struct through `cmdio.Render`, so `--output json` and the `Version` key both work there too. This matters because if the old CLI didn't support the flag, `getBundledCliVersion` would return `undefined`, `checkBundledCliVersion` would return `true`, and **no warning would fire for exactly the scenario that motivated the PR.** It does fire.

Also confirmed:
- `isDevExtension()` reads `process.env` at call time, not module load, so the tests' `beforeEach` mutation is actually effective (otherwise they'd pass vacuously).
- `cliPath` is hardcoded to the bundled binary and honors no user setting, so a user-configured CLI path can't produce a false warning.
- `.github/workflows/unit-tests.yml:66-70` runs `package:cli:fetch` at native arch before `test:cov` on both the Linux and Windows matrix legs, so the real-binary tests have a binary at the pinned version. (`push.yml` fetches `darwin_amd64`, but that job never runs unit tests, so it's harmless.)
- `__dirname/../../bin` resolves to `bin/` from both `src/utils` and compiled `out/utils` under `rootDir: src` / `outDir: out`.

**Still unverified:** I could not execute the suite — no `node_modules` and npm unreachable from this machine, the same wall described in the PR. Someone should run `yarn workspace databricks run test:unit` on fresh deps before merge. That's the one open item.

### Design note for the reviewer

This warns rather than prevents, and prevention is probably the better fix. `scripts/fetch-databricks-cli.sh` reads `cli.version` from `package.json` itself and `test:unit` already chains `yarn run build` — so the fetch could be wired into `build` or the F5 prelaunch task and the staleness could simply never happen.

One correction to my own earlier reasoning: the script has **no** version check, so it re-downloads the zip and SHA256SUMS unconditionally on every call. Wiring it into `build` as-is would add a network round-trip to every build. The right order is: add a version guard to the script first (reusing the version-reading logic this PR introduces), then wire it into `build`, and this warning becomes unnecessary. Reasonable to land this now and file that as follow-up — the warning has standalone value while the bundled CLI is fetched out-of-band.

*Reviewed with Claude Code.*

rugpanov (changes requested):
Thanks for this — the fix itself looks correct and I don't see any blockers in the code. My only ask is about the docs scope.

### Keep AGENTS.md lean

AGENTS.md was created deliberately small in #2079 as an agent-facing pointer to `CODE_CONVENTIONS.md`. The new section's own preamble says *"Full setup and test instructions live in CONTRIBUTING.md"* — but then inlines launching/activation/logs prose, which is exactly the "setup instructions" it's deferring to CONTRIBUTING.md for.

### One canonical home for the stale-CLI note

The "re-fetch after a `cli.version` bump / stale CLI → stuck on Initializing…" explanation now appears three times, near-verbatim:

- `AGENTS.md`
- `CONTRIBUTING.md`
- the `checkBundledCliVersion` doc-comment

A single canonical home (CONTRIBUTING.md) plus short pointers elsewhere would be easier to keep in sync.

### Suggestion

In this PR, I'd keep only the **"Re-fetch the CLI after pulling"** bullet in AGENTS.md — it's the one the new warning actually references — or even reduce it to a one-liner pointing at CONTRIBUTING.md. The launching / activation / `--extensions-dir` / logs material could move into CONTRIBUTING.md, or land as its own docs PR. That keeps this PR matched to its title.

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `51b468c0` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31480011012)

anton-107:
Thanks — agreed on all three points. Addressed in 51b468c.

**AGENTS.md is back to a pointer.** It now carries only the `package:cli:fetch`
command (the one the warning names) plus a line saying CONTRIBUTING.md explains
the symptom — 13 lines instead of 48. You were right that the section
contradicted its own preamble.

**CONTRIBUTING.md is the canonical home.** The full explanation — why a pull
doesn't update the binary, and why the failure reads as "Initializing..." rather
than an error — lives there and nowhere else now. The `checkBundledCliVersion`
doc-comment is down to one sentence plus a pointer to it, keeping only the
dev-only rationale, which is a fact about the code rather than about the devloop.

**The launching / activation / `--extensions-dir` / logs material is out of this
PR.** I'll open it as a docs-only PR into CONTRIBUTING.md rather than drop it —
it's the same class of "presents as nothing happening" problem, so I'd like it
written down, just not here. This PR is now the code change plus a re-fetch
pointer, which matches the title.

Two other things while I was in here, both flagged in my own self-review above:

- **The description was stale** and has been rewritten. It described a stub CLI
  and a Windows skip that e75e737 had already replaced with tests against the
  real bundled binary, and it still carried the "not yet run through the test
  runner" caveat.
- **That caveat is now resolved.** `test:unit` passes — 671 passing, 0 failing,
  including all six new cases. Two local-only workarounds were needed and are
  described in the updated Tests section; briefly, my `node_modules` predate the
  TypeScript ^6.0.3 bump, so I relaxed `ignoreDeprecations` to build (nothing
  committed) and couldn't run `test:lint` at all, since the stale install pins
  eslint 8 and can't read the flat config. `prettier -c` passes on every file
  touched here. CI covers lint properly.

I also left the prevention idea as a **Follow-up** section rather than folding it
in: wiring the fetch into `build` needs a version guard added to
`fetch-databricks-cli.sh` first, since it currently re-downloads unconditionally
and would put a network round-trip in every build.

*Replied with Claude Code.*

rugpanov (approved):

CONTRIBUTING.md
```diff
@@ -36,6 +36,15 @@ Fetch the Databricks CLI that the extension bundles:
 yarn workspace databricks run package:cli:fetch
 ```
 
+Re-run this whenever `cli.version` in `packages/databricks-vscode/package.json`
+changes — after every `git pull`, in practice. `bin/` is gitignored, so a pull
```
rclarey: I think just the first sentence is enough, we want this to be human readable as well

rclarey (commented):
Code changes look good, one comment about `CONTRIBUTING.md`

CONTRIBUTING.md
anton-107: Done in 9a21f600 — CONTRIBUTING.md now stops at the first sentence:

> Re-run this whenever `cli.version` in `packages/databricks-vscode/package.json` changes — after every `git pull`, in practice.

One knock-on worth flagging: the failure-mode prose you cut was the canonical copy that @rugpanov asked for earlier in this review, and both AGENTS.md and the `checkBundledCliVersion` doc-comment pointed at it. Rather than delete the symptom outright, I moved it to AGENTS.md and repointed the doc-comment there. So it's still written down exactly once, just in the agent-facing file instead of the human-facing setup guide — which is arguably where it belonged anyway, since an agent debugging a stuck "Initializing..." is the reader who needs it.

*Replied with Claude Code.*

rugpanov:
🤖 Integration tests triggered for `9a21f600` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31493911892)

PR #2097, opened 2026-08-10
fix(python-setup): show vN versions and offer the full supported range

rugpanov:
## Why

Two defects in the serverless version picker, found during the M4 bug bash ([DECO-27917](https://databricks.atlassian.net/browse/DECO-27917) defects #4 and #5):

- **#4 — Version picker shows bare integers instead of `vN`.** It displayed "5", "4" rather than "v5", "v4", inconsistent with the compute picker, the CLI output, and the docs. Display-only issue; the payload should stay a bare integer.
- **#5 — Only observed versions are offered.** Lower versions the project never used (v1, v2, v3) were absent and unreachable, contradicting the ERD, which specifies that all known versions are listed.

## What

- **`serverlessVersionScoring.ts`**: seed the whole supported range `[MIN_SUPPORTED_VERSION, MAX_SUPPORTED_VERSION]` at score 0 before applying observations, so every CLI-accepted version is offered. Unobserved versions keep score 0 (no sources) and sort below any observed candidate; the fallback still wins when nothing was observed, and observed versions still outrank the rest.
- **`serverlessVersionPicker.ts`**: render the QuickPick labels and placeholder in the `vN` display form via a new `displayVersion` helper. **Display-only** — each item's `version` field (the value forwarded to `--serverless-version`) stays the bare integer.
- **Tests**: assert the `vN` labels keep bare-integer payloads, that the full v1–v5 range is offered with unobserved versions scored 0, and update the resolver end-to-end expectation to the full ranked range.

## Verification

- Unit suite green: **666 passing, 0 failing, 10 pending**.
- ESLint + Prettier clean on all changed files.

[DECO-27917]: https://databricks.atlassian.net/browse/DECO-27917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/databricks-vscode/src/python-setup/utils/serverlessVersionPicker.test.ts
+++ b/packages/databricks-vscode/src/python-setup/utils/serverlessVersionPicker.test.ts
@@ -19,16 +19,20 @@ describe("buildVersionPickItems", () => {
         expect(items[1].picked).to.equal(false);
     });
 
-    it("labels each item with the bare version and only stars the picked one", () => {
+    it("labels each item with the vN display form and only stars the picked one", () => {
         const items = buildVersionPickItems([
             {version: "5", score: 100, sources: ["bundleYaml"]},
             {version: "4", score: 50, sources: ["notebook"]},
         ]);
 
-        // The picked item is visually marked; others are the plain version.
-        expect(items[0].label).to.contain("5");
+        // Labels use the `vN` form (matching the CLI/docs/compute picker); the
+        // picked item is additionally starred, others are just `vN`.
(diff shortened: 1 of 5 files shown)


rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `46ba99be` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31383452231)

misha-db (approved):

PR #2098, opened 2026-08-11
fix(python-setup): surface the CLI failure detail in the setup log

rugpanov:
## Why

On a failed `environments setup-local` run, the error popup shows a concise
message mapped from the error *code* and drops the CLI's own `error.message`.
Under `--output json` the CLI streams almost nothing to stderr, so that message
— e.g. uv's dependency version-conflict explanation — reaches neither the popup
nor the **Databricks Python Environment Setup** output channel. The
`E_PROVISION` copy nonetheless tells the user to "review the conflict in the
logs", which were effectively empty, so the failure was undiagnosable from the
UI.

## What

- **`errorMessages.ts`** — add `formatSetupFailureDetail`, a pure helper that
  builds the log detail block: the failing phase + error code, the raw
  `error.message`, and the per-phase status line.
- **`PythonSetupEnvironmentSetup.ts`** — widen the `showError` seam to
  `(message, detail?)` and pass `formatSetupFailureDetail(result)` at the
  failed-result site, so the CLI's explanation reaches the channel.
- **`pythonSetupDeps.ts`** — `showError` writes `detail` to the channel and, in
  place of force-revealing it, offers a **"Show Logs"** action that opens the
  channel only when the user asks. A self-opening, possibly-empty log is worse
  than one opened on demand.

Net: the popup stays concise, gains a **Show Logs** button, and the log it opens
now contains the actual conflict.

## Testing

- Unit suite (VS Code host): **668 passing**, 10 pending. The 7 failures are
  pre-existing `CliWrapper` tests that require the bundled `bin/databricks`
  (absent in a fresh checkout — `spawn … ENOENT`) and are unrelated to this
  change.
- New tests: `formatSetupFailureDetail` (phase/code, raw message, phase table,
  no-error → `undefined`); the real `showError` (writes detail, offers
  "Show Logs", reveals only when picked, writes nothing when there is no
  detail); the orchestrator passes the detail on a failed result.
- ESLint + Prettier clean on all changed files.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
@@ -310,6 +310,28 @@ describe("PythonSetupEnvironmentSetup.setup", () => {
         );
     });
 
+    it("passes the CLI's raw failure detail to the log, not just the popup copy", async () => {
+        const shown: {message: string; detail?: string}[] = [];
+        const setup = new PythonSetupEnvironmentSetup(
+            makeDeps({
+                cli: makeCli({resolve: ERROR_NO_TARGET}),
+                showError: async (message, detail) => {
+                    shown.push({message, detail});
+                },
+            })
+        );
+
+        await setup.setup();
+
+        expect(shown).to.have.length(1);
+        // The popup stays the concise mapped copy (no raw CLI flag noise)...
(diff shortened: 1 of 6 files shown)


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `59d611f8`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31485249913)

misha-db (approved):

PR #2100, opened 2026-08-11
feat(python-setup): add "Re-run Python setup" button on the ready row

rugpanov:
## Why

Once the uv-native Python setup completes, the configuration view shows a **"Python environment ready"** row with no discoverable way to run setup again — the only affordance is clicking the label, which is not obvious. This adds a button, like *Serverless → Configure Compute*, to re-trigger setup.

## What

- Add an **inline icon** and a **right-click "Re-run Python setup"** entry on the ready row (contextValue `databricks.environment.pythonSetup.success`), wired exactly like the existing `selectPythonInterpreter` entry (`inline@0` + `navigation@0`).
- Back it with a thin `databricks.environment.rerunPythonEnv` command that delegates to the existing, re-entrancy-guarded `setup` handler. A distinct id lets the menu show a "Re-run Python setup" title instead of the initial CTA's.
- **Telemetry:** run vs re-run is *one event with an enum label*, not two command-execution events. `python_env.setup.attempt` gains a required `trigger` (`initial | rerun`), derived from state (`readyRoots.has(root)`) rather than the command — so every entry point (CTA, row click, inline button, palette) labels the same event correctly.

Only the ready state gets the button; the error-state row is already a "Set up…" call-to-action.

## Verification

- `tsc --build` clean; `eslint` + `prettier -c` clean.
- 48 unit tests pass, including a new test that a second setup over an already-ready project labels the attempt `rerun`.

### DIFF
--- a/packages/databricks-vscode/package.json
+++ b/packages/databricks-vscode/package.json
@@ -511,6 +511,13 @@
                 "category": "Databricks"
             },
             {
+                "command": "databricks.environment.rerunPythonEnv",
+                "title": "Re-run Python setup",
+                "icon": "$(sync)",
+                "enablement": "databricks.context.activated && databricks.context.loggedIn && !databricks.context.remoteMode",
+                "category": "Databricks"
+            },
+            {
                 "command": "databricks.environment.selectPythonInterpreter",
                 "title": "Change Python environment",
                 "icon": "$(gear)",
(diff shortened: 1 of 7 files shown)


rugpanov:
🤖 Integration tests triggered for `0ed1e3f2` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31494804029)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `1e7959d7`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31495719698)

misha-db (approved):

PR #2104, opened 2026-08-12
Promote uv Python setup entry to the top level of the Configuration view

rugpanov:
## Why

When the uv-native Python setup is enabled, the **Python Environment** group in the Configuration view holds a single child — the **Set up Python environment** entry. The wrapping group adds an extra expand arrow and a level of nesting with no content of its own, so the entry is easier to find promoted to the top level.

## What

- `EnvironmentComponent.getRoot()` now checks the uv entry's visibility first. When visible (feature enabled + uv-suitable project) it returns the setup entry directly as a top-level row; otherwise it returns the **Python Environment** group unchanged for the legacy pip checklist.
- Removed the now-dead uv branch from `getChildren()`: the group only renders on the legacy path, so its children are always the checklist.
- Updated tests to assert the uv entry is a top-level leaf (no group, no children) when visible, and that the group still appears when it is not.

**Net effect:**
- uv setup enabled → a single top-level "Set up Python environment" row (no empty group, no expand arrow).
- uv setup disabled → the "Python Environment" group + legacy checklist, exactly as before.

## Verification

- `tsc --build --force`: clean
- `yarn test:unit`: 667 passing, 0 failing
- `eslint` + `prettier`: clean

No persisted or collapse state is keyed to the group, so there is nothing to migrate.

### DIFF
--- a/packages/databricks-vscode/src/ui/configuration-view/EnvironmentComponent.test.ts
+++ b/packages/databricks-vscode/src/ui/configuration-view/EnvironmentComponent.test.ts
@@ -7,7 +7,10 @@ import {ConfigModel} from "../../configuration/models/ConfigModel";
 import {ConfigurationTreeItem} from "./types";
 import {PythonSetupEntry} from "./pythonSetupEntry";
 
-const ENVIRONMENT_ROOT = {id: "ENVIRONMENT"} as ConfigurationTreeItem;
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests triggered for `7c9a7cd4` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31591607854)

misha-db (approved):

PR #2105, opened 2026-08-12
feat(python-setup): auto-reveal the setup output channel on completion

rugpanov:
## Why

When local uv Python setup finishes, the "Databricks Python Environment Setup" output channel was only revealed if the user clicked the notification's button ("View Details" on success, "Show Logs" on failure). The full log — the details on success and the CLI's own explanation on failure — stayed hidden behind that extra click. We want to land the user on the output automatically, in both the success and error cases.

## What

- `showSuccess` and `showError` now call `wiring.log.show()` unconditionally after appending the log, so the output channel is revealed as soon as setup finishes — before the notification is even dismissed.
- The notification is unchanged: same info/warning/error toast and the same jump-to-logs button, which still re-reveals the channel when picked.
- `log.show()` maps to `.show(true)` (`preserveFocus: true`), so the panel is revealed without stealing focus from the editor.
- Updated the surrounding comments to reflect the new behavior.

## Tests

- Reworked the `showError` reveal tests: the channel now reveals regardless of what the user clicks, and the button reveal stacks on top of the automatic one.
- Added a `showSuccess` describe block asserting the automatic reveal (info vs warning toast) and the button re-revealing on top.
- `yarn workspace databricks run test:unit` — 706 passing.
- `yarn workspace databricks run fix` — clean.

## Note

This is a deliberate behavior change: the output panel now appears on every setup run, where before it stayed collapsed unless the user asked for it.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/controllers/pythonSetupDeps.test.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/pythonSetupDeps.test.ts
@@ -8,6 +8,10 @@ import {
 } from "./pythonSetupDeps";
 import {PythonSetupState} from "../../vscode-objs/StateStorage";
 import {Telemetry} from "../../telemetry";
+import {
+    SUCCESS_DEFAULT,
+    SUCCESS_WITH_WARNINGS,
+} from "../models/fixtures/setupLocalResults";
 
 describe("makePythonSetupVisibility", () => {
     const uvDetection = {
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `d504929b`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31600435151)

misha-db (approved):

PR #2112, opened 2026-08-13
fix(python-setup): don't let an open notification wedge the setup entry

rugpanov:
## Changes

Fixes a re-entrancy bug in the uv-native Python-setup orchestrator where the config-view row (**"Set up Python environment"** / **"Python environment ready"** / the new **"Python environment is drifted"** re-run) would stop responding to clicks until the window was reloaded.

Root cause: `setup()` wraps the run in an `inFlight` re-entrancy guard (to coalesce concurrent clicks onto a single project-mutating run), but `runSetup()` **`await`ed its terminal notification** — `showError` / `showSuccess` / `notify`, each of which awaits `window.show*Message(...)`, and that promise stays pending until the user acts on the toast. So once a run finished, if the user left the notification open (e.g. dismissed the log panel but not the toast, or after a failed first-time setup), `inFlight` never cleared and every subsequent click returned the same still-pending promise — nothing happened.

Fix: present terminal notifications **fire-and-forget** through a small `present()` helper (`void notification.catch(() => {})`), so the guarded run settles when the *work* (CLI run + interpreter adoption + state write) completes, not when the toast is dismissed. Telemetry (`reportResult`) is still recorded **before** presenting, so the setup-time `duration` metric is unchanged; a rejection from the notification is logged at debug (not silently swallowed, not surfaced) — `showSuccess`/`showError` do real work before the toast.

## Tests

- New unit test in `PythonSetupEnvironmentSetup.test.ts`: a second `setup()` starts a fresh CLI run even while the prior run's notification is still open. It **times out on the old code** (the guard never releases) and passes with the fix.
- Existing re-entrancy tests (coalesce concurrent calls; clear the guard on rejection) still pass.
- Full unit suite: **733 passing, 0 failing, 10 pending**; `test:lint` clean.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
@@ -560,6 +560,29 @@ describe("PythonSetupEnvironmentSetup.setup", () => {
         expect(setup.ready).to.equal(true);
     });
 
+    it("clears the guard after the work even if the notification stays open", async () => {
+        // Model a user who leaves the success toast up (never dismisses it):
+        // showSuccess never resolves. The re-entrancy guard must release when the
+        // mutating work (CLI run + adopt + state) finishes, NOT when the toast is
+        // dismissed -- otherwise the entry stays wedged (every later click returns
+        // the still-pending promise) until the window is reloaded.
+        const cli = makeCli({resolve: SUCCESS_REAL_RUN});
+        const setup = new PythonSetupEnvironmentSetup(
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests triggered for `bbd1d987` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31707104914)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `0f6db33a` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31707584355)

misha-db (approved):

PR #2118, opened 2026-08-14
feat(python-setup): route the setup command to the uv flow when active

rugpanov:
## Why

When the uv-native Python setup is the active surface for a project — the `environment.pythonSetup` opt-in plus a uv-suitable project, the same gate the **configuration view** already dispatches on — the `databricks.environment.setup` command still ran the **legacy checklist**. Every "set up my Python env" trigger funnels through that command:

- the **Databricks Connect status bar** button,
- the **config-view legacy rows**,
- the **command palette** entry,
- the **Run/Debug current file** gate's own `executeCommand("databricks.environment.setup")`.

So all of them drove the wrong flow for uv projects. Separately, the legacy interpreter-change listener would pop a **second** install prompt on top of the uv flow.

## What

- **`pythonSetupRouting.ts`** — a narrow `UvPythonSetup` seam plus `routeEnvironmentSetup`, which sends `databricks.environment.setup` to the uv flow when it is visible, else the legacy checklist. One change reroutes **every** trigger surface (the palette command is **not** hidden — it routes).
- **Interpreter-change guard** (`EnvironmentDependenciesVerifier`) — suppress the legacy auto-install prompt when uv is active, so the two flows don't both prompt.
- **`extension.ts`** — wire both.

### On readiness (deliberate design choice)

The **run/debug gate** and the **status bar** keep reading the legacy `environment.dependencies` check for *readiness*. That check reads **real installed state** (interpreter version + `databricks-connect` presence/version), so it is **reload-safe** and **interpreter-safe** — unlike the uv flow's session-only `ready` flag, which is empty after a window reload and is never invalidated when the user switches interpreters. Routing only the setup **trigger** (not the readiness signal) drives the uv setup from every surface while avoiding a reload / interpreter-staleness regression. **No cross-fallback**: when uv is active it is the only setup path.

All new dependencies are optional and default to legacy behavior, so nothing changes for users who have not opted in (the uv gate returns `false` → every surface behaves exactly as before). No new settings, persisted state, telemetry events, or when-clause flags.

## Verification

- Unit tests for the command routing and the interpreter-change guard.
- `yarn test:unit`: **790 passing, 0 failing** (10 pre-existing pending).
- `eslint` + `prettier -c` clean on all touched files.
- Reviewed with Isaac Review (0 findings), Codex, a Claude code-review agent, a devil's-advocate pass, and a CODE_CONVENTIONS.md compliance pass; this design incorporates their feedback (the earlier readiness-routing approach was dropped after review flagged the reload/interpreter regressions).

### DIFF
--- a/packages/databricks-vscode/src/extension.ts
+++ b/packages/databricks-vscode/src/extension.ts
@@ -17,6 +17,7 @@ import {ClusterModel} from "./cluster/ClusterModel";
 import {ClusterCommands} from "./cluster/ClusterCommands";
 import {ConfigurationDataProvider} from "./ui/configuration-view/ConfigurationDataProvider";
 import {composePythonSetupEntry} from "./ui/configuration-view/pythonSetupEntry";
+import {routeEnvironmentSetup} from "./language/pythonSetupRouting";
 import {COPY_COMMAND_IDS} from "./ui/configuration-view/copyActions";
 import {AiToolsManager} from "./aitools/AiToolsManager";
 import {AiToolsCommands} from "./aitools/AiToolsCommands";
@@ -910,7 +911,10 @@ export async function activate(
                 pythonExtensionWrapper,
                 environmentDependenciesInstaller,
                 configureAutocomplete,
-                packageManagerTelemetry
+                packageManagerTelemetry,
+                // Constructed lazily (first isEnabled call is well after
+                // pythonSetupEnvironment is wired), so this reference is safe.
+                () => pythonSetupEnvironment.isVisible()
             )
     );
     // uv-native Python environment setup (python-setup). Constructed always,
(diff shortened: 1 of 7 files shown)


rugpanov:
🤖 Integration tests triggered for `02532f27` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31799157805)

rugpanov:
🤖 Integration tests triggered for `9d331a2a` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31799902820)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `3fc96370` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31800292938)

misha-db (approved):

PR #2119, opened 2026-08-14
[E2E] Add timestamp to unique name

rclarey:
## Changes

Add a timestamp to the unique name given to E2E directories. This allows a cleanup job to run and delete directories older than a given threshold

## Tests

None


rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `c3b0f1b0`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31800728163)

rugpanov (approved):
LGTM — reviewed the change to `getUniqueResourceName`. Name-length and charset constraints are fine (worst case well under the 4096-byte job-name limit), collision resistance is unchanged/improved, and no existing test parses the name format, so this is backward-compatible.

One non-blocking suggestion: the `vscode_integration_test_<ms>_<uuid>` format is now effectively a contract consumed by the external cleanup job. A short comment pointing at that job (and, ideally, a tiny test pinning the format) would guard against a future refactor silently breaking cleanup. Legacy UUID-only directories created before this merge won't carry a parseable timestamp, so the cleanup job will need to handle those separately.

rugpanov:
**Review suggestions (non-blocking)** — no blocking issues found across the review.

1. The `vscode_integration_test_<ms>_<uuid>` format is now effectively a contract for the external cleanup job. A comment referencing that job (and a small test pinning the format) would prevent a future refactor silently breaking cleanup.
2. Legacy UUID-only directories created before this change carry no parseable timestamp, so the cleanup job needs to handle them separately.

PR #2120, opened 2026-08-14
test(e2e): retry the dbconnect pip fallback through transient Windows file locks

rugpanov:
## Why

The `run_dbconnect` e2e test's **should setup virtual environment** step flakes on the Windows shard. When the extension's install path leaves `databricks-connect` missing, the test's own `ensureVenvHasDbConnect` fallback runs `pip install databricks-connect`. While pip overwrites the venv's existing numpy it can hit:

```
ERROR: Could not install packages due to an OSError: [WinError 32]
The process cannot access the file because it is being used by another process:
'...\.venv\Lib\site-packages\numpy\_core\_add_newdocs_scalars.py'
```

— a **transient** lock (a scanner or a still-running install holding the file open). The fallback swallowed that error and moved on, so `databricks-connect` stayed missing and the downstream `"Databricks Connect:"` tree-item check timed out with `Setup confirmation failed`. `ensureVenvHasKernelDeps` had the same single-shot-then-swallow shape.

## What

- **`src/test/retry.ts`** — a pure `retryOnTransientError` helper plus an `isTransientFileLockError` predicate (matches WinError 32 "used by another process" and WinError 5 "Access is denied", searching `.message` / `.stderr` / `.stdout` since Node's `execFile` hides the OSError in stderr). It lives under `src/test/` rather than `src/test/e2e/` — `tsconfig.json` excludes the e2e folder from the unit build, so a colocated unit test only runs when the module sits outside it; the e2e spec imports it via an explicit `.ts` extension (as e2e specs already do for `src/` modules).
- **`src/test/retry.test.ts`** — unit coverage: success without sleeping, retry-then-succeed, linear backoff, give-up-after-N, non-transient-rethrow, `onRetry` reporting, and the predicate's positive/negative cases.
- **`run_dbconnect.ucws.e2e.ts`** — both fallback pip installs now run through the helper (3 attempts, 5s linear backoff), logging each retry so a **real** Windows UX regression stays visible in nightly logs. The final swallow+log and the ground-truth tree-item assertion are unchanged.

Test-only: no production code, settings, persisted state, telemetry, or when-clause flags — zero user-facing impact and nothing changes for shards that never hit the lock.

## Verification

- `retry.test.ts`: **9 passing** (written red-first, then green).
- Full unit suite (`test:unit`): **818 passing, 0 failing** (10 pre-existing pending).
- e2e file type-checks under `src/test/e2e/tsconfig.json` (the only error is a pre-existing `assert {type: "json"}` in `wdio.conf.ts`, untouched here).
- `eslint` + `prettier -c` clean on all touched files.

### DIFF
--- a/packages/databricks-vscode/src/test/e2e/run_dbconnect.ucws.e2e.ts
+++ b/packages/databricks-vscode/src/test/e2e/run_dbconnect.ucws.e2e.ts
@@ -16,6 +16,7 @@ import {
     getBasicBundleConfig,
     writeRootBundleConfig,
 } from "./utils/dabsFixtures.ts";
+import {isTransientFileLockError, retryOnTransientError} from "../retry.ts";
 
 const execFile = promisify(execFileCb);
 
@@ -51,7 +52,8 @@ const DBCONNECT_VERSION_SPEC = "17.3.*";
 // tests"), so the notebook cell never runs and no output file is written. We
 // install the deps directly to remove that dependency on the flaky UI step; the
 // pip call is idempotent, so on shards that already have them it is a fast
-// no-op.
+// no-op. The install is retried on transient Windows file locks (see
+// ensureVenvHasDbConnect for why).
 async function ensureVenvHasKernelDeps(projectDir: string) {
     const python = venvPython(projectDir);
     try {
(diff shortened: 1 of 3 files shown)


rugpanov:
🤖 Integration tests triggered for `3d6176cd` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31807184691)

rugpanov:
🤖 Integration tests ❌ **1 of 35 test jobs failed** for `68a6a757` (34 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31808097611)

misha-db (approved):

PR #2121, opened 2026-08-14
fix(dbconnect-run): surface Databricks Connect import failures

rugpanov:
## Problem

The run bootstrap wrapped the Databricks Connect import and the session build in one
broad `try/except` that logged any failure at `debug` level, so a broken environment
produced no visible signal. The most common break is a standalone `pyspark` installed
alongside `databricks-connect`: they share the `pyspark` package and overwrite each
other, and the user only sees an opaque Java or protobuf gencode error from their own
script, with no hint at the cause.

## What

- Split the import from the session build. An **import** failure now logs at error
  level with actionable guidance naming the standalone-`pyspark` conflict, because the
  user's own script hits the same import.
- A failure to *build* the session (no configured compute, auth) stays quiet at `debug`
  as before — many scripts create their own session and never use the injected one, so
  that path must not become noisy.

## Testing

- `python3 -m py_compile` on the bootstrap script.

### DIFF
--- a/packages/databricks-vscode/resources/python/dbconnect-bootstrap.py
+++ b/packages/databricks-vscode/resources/python/dbconnect-bootstrap.py
@@ -45,17 +45,35 @@ db_globals = {}
 from databricks.sdk.runtime import dbutils  # noqa: E402
 db_globals['dbutils'] = dbutils
 
-# "table", "sc", "sqlContext" are missing
+# "table", "sc", "sqlContext" are missing.
+#
+# The import and the session build are handled separately on purpose. A failure to
+# import Databricks Connect (or pyspark) means the Python environment itself is
+# broken and the user's own script will hit the same error, so it is surfaced. The
+# most common cause is a standalone "pyspark" package installed alongside
+# databricks-connect: the two share the pyspark namespace and overwrite each other,
+# and the resulting Java/protobuf error otherwise gives no hint at the cause. A
+# failure to *build* the session (no configured compute, auth) is left quiet — many
+# scripts create their own session and never touch the injected one.
 try:
     from pyspark.sql import functions as udf, SparkSession
     from databricks.connect import DatabricksSession
-    spark: SparkSession = DatabricksSession.builder.getOrCreate()
-    sql = spark.sql
-    db_globals['spark'] = spark
-    db_globals['sql'] = sql
-    db_globals['udf'] = udf
 except Exception as e:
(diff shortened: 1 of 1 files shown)


rugpanov:
🤖 Integration tests triggered for `6feb9d18` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31807640142)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `2024d1b1`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31808547995)

anton-107 (approved):
Reviewed as part of the three-PR stack (this, databricks/cli#6276, and databricks/cli#6277).

## Verified empirically

I reproduced the collision by cloning a real working `databricks-connect` 18.0.6 venv and layering standalone `pyspark` 4.2.0 on top:

```
ImportError: cannot import name 'PythonUDFEnvironment' from
'pyspark.sql.connect.expressions'   ← raised at `from databricks.connect import ...`
```

Two things this confirms about the change:

1. **The split is placed correctly.** The failure lands on the *import*, not on `DatabricksSession.builder.getOrCreate()`. So the new error-level branch is the one that actually fires for this bug, and the session-build path stays quiet as intended — which matters, since many scripts build their own session and never touch the injected one.
2. **Control reaches the new block.** `from databricks.sdk.runtime import dbutils` at line 45 sits outside the try and still succeeds under collision, so the import failure is not pre-empted by an earlier crash. (That line hung on network egress in my sandbox, but a known-functional control venv hung identically — blocked network, not the collision.)

The follow-up commit making the pyspark remediation conditional is the right call: the import can fail for other reasons (`databricks-connect` not installed at all, another broken dependency), and the original unconditional "Remove the standalone pyspark dependency" would have been misleading advice in those cases.

## Notes

- The message references `'Set up Python environment'`; the actual command title is `"Set up Python environment (uv)"` (`package.json:508`). Close enough to find in the palette, but exact would be better.
- The import handler catches bare `Exception` where `ImportError` is the real case. This preserves the prior behaviour so it isn't a regression, but a narrower catch would avoid swallowing unrelated failures.

## Testing

Verification was `python3 -m py_compile`, which checks syntax only — it does not exercise either branch. The branch behaviour is simple enough to read, and I confirmed the import-vs-build split against a real collided venv as described above, so I'm not asking for a test here.

Approving.

PR #2123, opened 2026-08-14
test(e2e): defeat template-filter pollution in the bundle_init wizard

rugpanov:
## Why

`bundle_init :: should initialize new project` still flakes on the Windows shard with **"Can't complete cli bundle init wizard"** (its two follow-on tests then cascade), despite #2034 gating the first keystroke on tab readiness. Two races remain:

1. **Keystroke focus race** — the editor tab becoming *active* doesn't guarantee the xterm holds keyboard focus, so keys fired next can land in the tab chrome instead of the wizard.
2. **Template-filter pollution** — per `BundleInitWizard.bundleInitInTerminal` (see its own comment), the Python extension can inject env-setup text into the freshly opened terminal, and that text lands in the wizard's template **search filter**. The typed `default-python` then appends to that text, matches no template, and the wizard stalls until the 40s completion loop times out.

The terminal buffer isn't reliably readable via the wdio API on this editor-hosted terminal, so the fix hardens the input path rather than reading state back.

## What

- **Focus the terminal input** (`workbench.action.terminal.focus`) after the tab is active and before typing, so wizard keystrokes aren't swallowed by the editor-tab chrome.
- **Clear the template search filter** with a burst of `Backspace` keys before typing the template name — a no-op when the filter is already empty, and it removes any injected env-setup text so `default-python` matches.
- The Enter-hammering completion loop (which accepts the remaining default-python prompt defaults) and the ground-truth workspace-root gate are **unchanged**.

Test-only: no production code, settings, persisted state, telemetry, or when-clause flags. Nothing changes for shards that never hit the races.

## Verification

- `tsc -p src/test/e2e/tsconfig.json`: **no new errors** (the sole error is the pre-existing `assert {type: "json"}` in `wdio.conf.ts`, untouched here).
- `eslint` + `prettier -c`: clean on the touched file.
- e2e-only behavior is verified via **isolated CI runs of `bundle_init`** on this branch (repeated, since one green run ≠ stable).


rugpanov:
🤖 Integration tests triggered for `887af748` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/31836522000)

rugpanov:
🤖 Integration tests ✅ **all 35 test jobs passed** for `e0c08501`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32008117691)

misha-db (approved):

PR #2124, opened 2026-08-17
feat(python-setup): GA the uv-native environment setup

rugpanov:
## Why

The uv-native "Set up Python environment" flow shipped behind an opt-in feature flag (`environment.pythonSetup` in `databricks.experiments.optInto`) because its CLI command, `environments setup-local`, was only available in custom CLI builds. That command now ships in the bundled CLI (v1.12.1), so the flag — and the temporary `cliPathOverride` escape hatch — have served their purpose. This makes the flow generally available.

## What

- **Remove the opt-in flag.** Drop `environment.pythonSetup` from the `databricks.experiments.optInto` enum, delete the `PYTHON_SETUP_FEATURE_ID` constant, and construct the `FeatureManager` with an empty disabled list.
- **Remove the temporary override.** Delete the `databricks.experiments.cliPathOverride` setting, its `WorkspaceConfigs` getter, and `resolveCliPath`; the CLI client always uses the bundled CLI.
- **Visibility is now purely project-based.** The uv flow shows only for uv-suitable projects (`isUvSetupSuitable`) — clean uv/greenfield projects with no competing manager. Projects driven by pip/poetry/conda keep the legacy checklist via `routeEnvironmentSetup`, exactly as before.
- **Serverless-version prompt scoped to uv projects.** `ConnectionCommands.selectServerless` now gates the version prompt on uv-suitability (injected `isUvSetupVisible`) rather than the flag, so enabling serverless on a non-uv project keeps the plain, version-less enable — the old flow for pip users is unchanged.
- **Tests.** Delete the flag-coupling drift test; update the gate/deps/args/connection tests to the always-on behavior and add a regression test pinning the pip-path (plain serverless enable, no version prompt).

## Backward compatibility

- Users who already added `environment.pythonSetup` to `experiments.optInto` — harmless leftover; the value is simply no longer read.
- Users who set `experiments.cliPathOverride` — the setting is removed (VS Code flags it as unknown, benign) and the bundled CLI is used, which now ships the command.
- No persisted-state or telemetry-schema changes.

## Verification

- `yarn --cwd packages/databricks-vscode build` — clean
- `yarn --cwd packages/databricks-vscode test:lint` — clean
- `yarn --cwd packages/databricks-vscode test:unit` — 829 passing, 0 failing


rugpanov:
🤖 Integration tests triggered for `dbbbb746` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32032313886)

rugpanov:
🤖 Integration tests ❌ **3 of 35 test jobs failed** for `5441c0d9` (32 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32032790798)

misha-db (approved):

PR #2133, opened 2026-08-19
Avoid duplicate profile section creation in .databrickscfg during OAuth setup

### LINKED ISSUE #2129: [BUG] VS Code extension creates duplicate Databricks profile sections during OAuth setup
**Describe the bug**

When configuring Databricks authentication through the VS Code extension using OAuth, the extension creates two sections with the same profile name in `~/.databrickscfg`.

For example, after starting with no `.databrickscfg` file and creating a profile named `dev-profile`, the resulting configuration contains two `[dev-profile]` sections.

This makes `.databrickscfg` invalid for consumers using Python's `configparser`, including the Databricks SDK, and results in `configparser.DuplicateSectionError`.

**To Reproduce**

1. On Windows, delete or rename `%USERPROFILE%\.databrickscfg` so that no Databricks configuration file exists.
2. Open a Databricks Asset Bundle project in VS Code.
3. Open the Databricks extension.
4. Configure the project and select the `dev` bundle target.
5. Under **Login to Databricks**, select **OAuth – Create a profile and authenticate using OAuth**.
6. Authenticate successfully in the browser.
7. Name the new profile `dev-profile`.
8. Confirm that the Databricks VS Code extension connects successfully.
9. Open `%USERPROFILE%\.databrickscfg`.

The resulting file contains two sections named `[dev-profile]`, similar to:

```ini
[dev-profile]
host = https://<workspace-host>
account_id = <account-id>
workspace_id = <workspace-id>
auth_type = databricks-cli

;This profile is autogenerated by the Databricks Extension for VS Code
[dev-profile]
host=https://<workspace-host>/
auth_type=databricks-cli
```

Commands or Python code that subsequently read `.databrickscfg` can then fail with:

```text
configparser.DuplicateSectionError:
While reading from WindowsPath('C:/Users/<username>/.databrickscfg')
```

In my case this caused pytest tests using the Databricks SDK to fail during configuration loading.

**Expected behavior**

The OAuth setup should create only one section for the selected profile name:

```ini
[dev-profile]
...
```

Alternatively, if the profile already exists, the extension should update or reuse the existing section rather than append another section with the same name.

**System information:**

1. `Help: About` output:

```text
Version: 1.132.0 (system setup)
Commit: df53daabb18cd157bdb08c7f01c34df936cf12f4
Date: 2026-08-04T15:30:20Z
Electron: 42.7.1
ElectronBuildId: 14760073
Chromium: 148.0.7778.280
Node.js: 24.18.0
V8: 14.8.178.38-electron.0
OS: Windows_NT x64 10.0.26200
```

2. Databricks Extension Version:

```text
2.13.1
```

3. Databricks CLI version:

```text
Databricks CLI v1.11.0
```

4. OS:

```text
Windows
```

**Databricks Extension Logs**

I will attach the Databricks extension logs if required.

**Additional context**

I reproduced this from a clean state by deleting `.databrickscfg` before configuring authentication through the extension, so there was no pre-existing `dev-profile` section.

A workaround that avoids the duplicate configuration is to create the OAuth profile first using the CLI:

```powershell
databricks auth login --host <workspace-host> --profile dev-profile
```

and then configure the VS Code extension by selecting the existing `dev-profile` rather than choosing **Create a profile and authenticate using OAuth**.

With that workflow, the CLI owns the profile creation and the VS Code extension reuses the existing profile.

I have removed/redacted workspace URLs, account IDs, workspace IDs, usernames, and other organisation-specific information from this report.

misha-db:
## Root cause

Creating a profile through the **OAuth – Create a profile and authenticate using
OAuth** flow wrote **two** identically-named sections into `~/.databrickscfg`
(e.g. two `[dev-profile]` blocks). A file with duplicate sections is invalid for
consumers using Python's `configparser` — including the Databricks SDK — which
fail to load it with `configparser.DuplicateSectionError`.

Fixes #2129.

## Root cause

`saveNewProfile` (`packages/databricks-vscode/src/configuration/LoginWizard.ts`)
runs two independent writes for the `databricks-cli` (OAuth) auth type:

1. The auth check (`checkAuthProvider` → `DatabricksCliCheck.login`) runs
   `databricks auth login --host <host> --profile <name>`, and **the CLI itself
   persists the `[<name>]` section**.
2. `saveNewProfile` then **unconditionally appended a second `[<name>]` section**
   (prefixed with the `;This profile is autogenerated…` comment).

The existing duplicate-name guard in the wizard runs *before* the CLI login, so
it never sees the section the CLI just wrote. This double-write only affects the
OAuth provider — for PAT and Azure CLI the CLI does not persist the profile, so
the extension's append is still required.

## Fix

Before appending, `saveNewProfile` now parses the config file and, if a section
with the target profile name already exists, reuses it instead of appending a
duplicate. This matches the issue's suggested "reuse the existing section rather
than append another" behavior. It's safe for the other callers: the wizard
already rejects existing names at input time, and bundle migration uses random
`${authType}-${rnd}` names.

## Tests

LoginWizard.test.ts

### DIFF
--- a/packages/databricks-vscode/src/configuration/LoginWizard.test.ts
+++ b/packages/databricks-vscode/src/configuration/LoginWizard.test.ts
@@ -1,0 +1,101 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import {expect} from "chai";
+import * as tmp from "tmp";
+import path from "node:path";
+import {readFileSync, writeFileSync} from "node:fs";
+import ini from "ini";
+import {instance, mock} from "ts-mockito";
+import {saveNewProfile} from "./LoginWizard";
+import {DatabricksCliAuthProvider} from "./auth/AuthProvider";
+import {CliWrapper} from "../cli/CliWrapper";
+
+describe("saveNewProfile", () => {
+    const host = new URL("https://test.cloud.databricks.com/");
+    const cliPath = "/path/to/bin/databricks";
+    const cleanups: Array<() => void> = [];
+    let previousConfigFile: string | undefined;
+
+    beforeEach(() => {
+        previousConfigFile = process.env.DATABRICKS_CONFIG_FILE;
+    });
+
+    afterEach(() => {
+        if (previousConfigFile === undefined) {
+            delete process.env.DATABRICKS_CONFIG_FILE;
+        } else {
+            process.env.DATABRICKS_CONFIG_FILE = previousConfigFile;
+        }
+        while (cleanups.length) {
+            cleanups.pop()!();
+        }
+    });
+
+    function tempConfigFile(contents?: string): string {
+        const dir = tmp.dirSync({unsafeCleanup: true});
+        cleanups.push(dir.removeCallback);
+        const file = path.join(dir.name, ".databrickscfg");
+        if (contents !== undefined) {
+            writeFileSync(file, contents);
+        }
+        process.env.DATABRICKS_CONFIG_FILE = file;
+        return file;
+    }
+
+    function oauthProvider(profile: string) {
+        return new DatabricksCliAuthProvider(
+            host,
+            cliPath,
+            instance(mock(CliWrapper)),
+            profile
+        );
+    }
+
+    function countSections(file: string, name: string): number {
+        // ini.parse collapses duplicate sections, so count the raw headers.
+        const headerRegex = new RegExp(
+            `^\\[${name.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}\\]\\s*$`,
+            "gm"
+        );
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `74bde448`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32254132160)

rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `1f221766`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32467344424)

rugpanov (approved):

PR #2134, opened 2026-08-19
test(packaging): unit-test bundled-CLI version logic, isolate the real-binary smoke test

rugpanov:
## Why

The `bundled CLI version` tests in `packageJsonUtils.test.ts` spawned the **real ~50 MB `databricks` CLI** under mocha's **2 s default timeout**. On the slow Windows unit runner, the first cold spawn (AV scan + cold FS cache) intermittently exceeded 2 s → `Timeout of 2000ms exceeded`. It's a genuine flake — the same untouched file alternated pass/fail across recent commits (Linux always passes; it spawns faster).

Spawning the real CLI from a unit test is itself established here (`CliWrapper.test.ts:63` does the same, and it already carries a 10 s timeout), so the fix is a timeout bump plus an extraction that gives the version parse real coverage without spawning — not a relabel.

## What

- Raised the timeout on the real-binary assertions to **30 s** and grouped them into a clearly-labeled **smoke** describe — they validate the `package:cli:fetch` step (the "re-fetch the CLI" pitfall), not unit logic. Non-spawning gate paths (dev-checkout gate, unpinned version, missing binary) stay as fast unit tests.
- Extracted `parseCliVersion(stdout)` — the `databricks version --output json` → `Version` parse (undefined on malformed/missing/non-string) — and unit-tested it **without spawning**. `getBundledCliVersion` delegates to it and logs a (capped) debug line when the output is unparseable.
- `checkBundledCliVersion` keeps its behavior: an unknown version (CLI unreadable) is treated as "not stale", and the dev-checkout / unpinned gate short-circuits before any spawn.

Net: the version parse gets real, fast, deterministic unit coverage; the one honest real-binary check is isolated and no longer flakes on the 2 s default.

## Verification

- `yarn test:unit` — 862 passing, 0 failing (the new `parseCliVersion` units + the smoke describe with the fetched binary).
- `eslint` + `prettier` clean.

### DIFF
--- a/packages/databricks-vscode/src/utils/packageJsonUtils.test.ts
+++ b/packages/databricks-vscode/src/utils/packageJsonUtils.test.ts
@@ -165,10 +165,18 @@ describe(__filename, () => {
                 false
             );
         });
+
+        it("is not a mismatch when both versions are unknown", () => {
+            assert.equal(
+                isBundledCliVersionMismatch(undefined, undefined),
+                false
+            );
+        });
     });
 
-    // Logic-only checkBundledCliVersion paths that return before spawning the
-    // CLI — safe to keep in the unit suite.
+    // checkBundledCliVersion paths that never launch the real bundled CLI —
+    // they hit the dev-flag / unpinned gate, or fail fast on a missing binary —
+    // so they stay fast in the unit suite.
     describe("checkBundledCliVersion gating", () => {
         const originalDevFlag = process.env[EXTENSION_DEVELOPMENT];
         afterEach(() => {
@@ -193,10 +201,13 @@ describe(__filename, () => {
         it("does not warn when the pinned version is unknown", async () => {
             process.env[EXTENSION_DEVELOPMENT] = "true";
             assert.ok(
-                await checkBundledCliVersion("/nonexistent/databricks", {
-                    packageName: "databricks",
-                    version: "2.13.0",
-                })
+                await checkBundledCliVersion(
+                    path.join(__dirname, "nonexistent-databricks"),
+                    {
+                        packageName: "databricks",
+                        version: "2.13.0",
+                    }
+                )
             );
         });
 
--- a/packages/databricks-vscode/src/utils/packageJsonUtils.test.ts
+++ b/packages/databricks-vscode/src/utils/packageJsonUtils.test.ts
@@ -9,7 +9,6 @@ import {
     getBundledCliVersion,
     getCorrectVsixInstallString,
     getMetadata,
-    isBundledCliVersionMismatch,
     isCompatibleArchitecture,
     isEqual,
     nodeArchMap,
@@ -29,6 +28,19 @@ const cliPath = path.join(
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 const pinnedCliVersion = require("../../package.json").cli.version;
+
+// Restores EXTENSION_DEVELOPMENT after each test in the enclosing describe, so a
+// suite that toggles the dev flag can't leak it into the rest of the run.
+function restoreDevFlagAfterEach() {
+    const original = process.env[EXTENSION_DEVELOPMENT];
+    afterEach(() => {
+        if (original === undefined) {
+            delete process.env[EXTENSION_DEVELOPMENT];
+        } else {
+            process.env[EXTENSION_DEVELOPMENT] = original;
+        }
+    });
+}
 
 describe(__filename, () => {
     it("should correctly check compatibility", () => {
@@ -137,55 +149,11 @@ describe(__filename, () => {
         });
     });
 
-    describe("isBundledCliVersionMismatch", () => {
-        it("flags two known versions that differ", () => {
-            assert.equal(
-                isBundledCliVersionMismatch("0.240.0", "0.241.0"),
-                true
-            );
-        });
-
-        it("is not a mismatch when the versions match", () => {
-            assert.equal(
-                isBundledCliVersionMismatch("0.240.0", "0.240.0"),
-                false
-            );
-        });
-
-        it("is not a mismatch when the actual version is unknown", () => {
-            assert.equal(
-                isBundledCliVersionMismatch(undefined, "0.240.0"),
-                false
-            );
-        });
-
-        it("is not a mismatch when the expected version is unpinned", () => {
-            assert.equal(
-                isBundledCliVersionMismatch("0.240.0", undefined),
-                false
-            );
-        });
-
-        it("is not a mismatch when both versions are unknown", () => {
-            assert.equal(
-                isBundledCliVersionMismatch(undefined, undefined),
-                false
-            );
-        });
-    });
-
     // checkBundledCliVersion paths that never launch the real bundled CLI —
     // they hit the dev-flag / unpinned gate, or fail fast on a missing binary —
     // so they stay fast in the unit suite.
     describe("checkBundledCliVersion gating", () => {
-        const originalDevFlag = process.env[EXTENSION_DEVELOPMENT];
-        afterEach(() => {
-            if (originalDevFlag === undefined) {
-                delete process.env[EXTENSION_DEVELOPMENT];
-            } else {
-                process.env[EXTENSION_DEVELOPMENT] = originalDevFlag;
-            }
-        });
+        restoreDevFlagAfterEach();
 
         it("does not warn outside a dev checkout (no CLI spawn)", async () => {
             delete process.env[EXTENSION_DEVELOPMENT];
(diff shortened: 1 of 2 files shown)


rugpanov:
🤖 Integration tests triggered for `ab95a6e6` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32304023556)

rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `3cdb908d`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32305000479)

packages/databricks-vscode/src/utils/packageJsonUtils.ts
```diff
@@ -91,12 +91,37 @@ export async function getMetadata(
...
+        return typeof version === "string" ? version : undefined;
+    } catch {
+        return undefined;
+    }
+}
+
+// True only when both versions are known and differ — the case worth warning
+// about. An unknown actual (CLI unreadable) or unknown expected (unpinned) is
+// deliberately not a mismatch. Pure — unit-tested without spawning the CLI.
+export function isBundledCliVersionMismatch(
```
anton-107: This one feels over-extracted. The body is `a !== undefined && b !== undefined && a !== b` — a total boolean expression with no product logic in it — but it gets a name longer than its implementation, a 3-line comment, and five unit tests that largely assert the semantics of `&&`.

It also makes the call site *harder* to read rather than easier: at line 200 the reader now has to recall that `expected` can't be `undefined` there for the code to mean what it used to. Needing a parenthetical in the PR description to prove the equivalence is the tell.

`parseCliVersion` is a genuinely valuable extraction — real parsing, real edge cases, real tests. For this one I'd inline the compare back at the call site and keep the truth table in the head, or at least drop it to one or two tests.

packages/databricks-vscode/src/utils/packageJsonUtils.ts
```diff
@@ -91,12 +91,37 @@ export async function getMetadata(
     };
 }
 
+// Extracts the "Version" string from `databricks version --output json` stdout.
+// Returns undefined on malformed JSON or a missing/non-string field, so callers
+// treat an unreadable version the same as an absent one. Pure — unit-tested
+// without spawning the CLI (getBundledCliVersion is the thin process wrapper).
```
anton-107: Comment volume runs against CODE_CONVENTIONS.md §4b ("shorter than the code it guards", "explain _why_, never _what_"). Across the two new helpers it's ~16 lines of comment for ~35 lines of code, and several lines restate the signature: "Returns undefined on malformed JSON or a missing/non-string field" narrates `typeof version === "string" ? version : undefined`, which the return type already says.

The `Pure — unit-tested without spawning the CLI` notes (here and on `isBundledCliVersionMismatch`) describe the *test file*, not the function contract — precisely the kind that rots silently when tests move. Same for `Both failure modes log at debug.` added to the JSDoc below.

§4b is explicit that this matters *more* when working AI-assisted, and the description says this was written by Isaac. Suggest trimming each to the load-bearing sentence — for this one, roughly "Returns undefined so callers treat an unreadable version the same as an absent one."

packages/databricks-vscode/src/utils/packageJsonUtils.ts
```diff
@@ -107,8 +132,14 @@ export async function getBundledCliVersion(
             "--output",
             "json",
         ]);
-        const version = JSON.parse(stdout)["Version"];
-        return typeof version === "string" ? version : undefined;
+        const version = parseCliVersion(stdout);
+        if (version === undefined) {
+            logging.NamedLogger.getOrCreate(Loggers.Extension).debug(
+                "Bundled Databricks CLI version output was unparseable",
+                {stdout}
```
anton-107: Minor: `{stdout}` logs whatever the binary printed, unbounded. Harmless for `databricks version` output (nothing sensitive), but if `cliPath` ever resolves to the wrong executable this dumps arbitrary process output into the extension log. `stdout.slice(0, 200)` costs nothing and caps the blast radius.

packages/databricks-vscode/src/utils/packageJsonUtils.test.ts
```diff
@@ -166,22 +198,71 @@ describe(__filename, () => {
...
         });
+    });
+
+    // Smoke tests: these spawn the REAL bundled CLI that CI fetches at the
+    // pinned version, so they validate the `package:cli:fetch` step, not unit
+    // logic (the version parsing/compare is unit-tested above). Cold-spawning a
+    // ~50MB binary on the Windows runner exceeds the 2s mocha default, so give
+    // the suite a generous timeout — the default made this flake intermittently.
+    describe("bundled CLI (smoke — spawns the real fetched binary)", function () {
+        this.timeout(30_000);
```
anton-107, line 214: This fixes the instance but the class of flake survives, and it's order-dependent. The cold-spawn penalty attaches to whichever spawn test glob happens to run first, and glob order isn't a stable contract — add or rename a file under `src/` and the penalty migrates to `CliWrapper.test.ts`, whose 10s may then become the thing that flakes.

A default `timeout` on the Mocha instance in `src/test/suite.ts` (which currently passes only `ui` and `color`) would retire this failure mode repo-wide in one line, instead of per-file as each one flakes in turn. Fine as a follow-up if you'd rather keep this PR tight.

packages/databricks-vscode/src/utils/packageJsonUtils.test.ts
```diff
@@ -166,22 +198,71 @@ describe(__filename, () => {
...
+
+        const originalDevFlag = process.env[EXTENSION_DEVELOPMENT];
+        beforeEach(() => {
+            process.env[EXTENSION_DEVELOPMENT] = "true";
+        });
+        afterEach(() => {
+            if (originalDevFlag === undefined) {
+                delete process.env[EXTENSION_DEVELOPMENT];
+            } else {
+                process.env[EXTENSION_DEVELOPMENT] = originalDevFlag;
```
anton-107: The 8-line env save/restore block is now duplicated verbatim between this `describe` and `checkBundledCliVersion gating` above. Worth hoisting to a small helper in the file (or `src/test/utils.ts`) so the two can't drift.

anton-107 (approved):
Approving — correct diagnosis, behavior-preserving, and CI is green on both unit-test jobs. Comments are quality/convention only; none block.

Things I checked rather than took on faith:

- **Behavior really is unchanged.** `!isBundledCliVersionMismatch(actual, expected)` expands to `actual === undefined || expected === undefined || actual === expected`; the extra disjunct is already short-circuited by the earlier `metaData.cliVersion === undefined` gate, so it reduces to the old condition.
- **The flake diagnosis holds.** `src/test/suite.ts` constructs Mocha with no `timeout`, so the 2s default applies. Glob order puts `packageJsonUtils.test.ts` at index 8 and `CliWrapper.test.ts` at index 42 — so this file absorbed the cold-spawn cost (AV scan + cold FS cache) while the other was both warm *and* already carrying `this.timeout("10s")`. That explains exactly why this file flaked and that one didn't.
- **`parseCliVersion` has no crash path.** Exercised `null`, `5`, `[1,2]`, `{"Version": null}`, `""`, and non-JSON — all return `undefined` via the `catch` or the `typeof` guard, none throw.
- **No coverage dropped** — all six original assertions survive, renamed and redistributed.

One structural note beyond the inline comments: the framing in the description ("isn't a unit test to begin with — it was mislabeled `.test.ts`") doesn't match what the change does — the tests stay in `.test.ts` with a raised timeout. That's the right call and it matches precedent (`CliWrapper.test.ts:63` spawns the real CLI from a unit test, and CODE_CONVENTIONS.md §8's suffix table has no "smoke" row), but the stated rationale and the change disagree. Describing it as a timeout fix plus an extraction would be more accurate.

Nits, non-blocking: `assert.equal` is loose, so `assert.equal(x, undefined)` also passes for `null` — `strictEqual` is tighter, though loose `equal` is the established style in this file. The new "unparseable stdout" branch in `getBundledCliVersion` is itself uncovered — the extraction tests the parser, not the wiring to it. And 30s is generous next to `CliWrapper`'s 10s for the same binary.

rugpanov:
🤖 Integration tests triggered for `5ebc69bf` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32349753716)

rugpanov:
🤖 Integration tests triggered for `8e129f6e` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32351635092)

PR #2138, opened 2026-08-21
fix(python-setup): show an info toast on a successful setup that had warnings

rugpanov:
## Why

The Python environment-setup success notification raised a ⚠️ **warning** toast whenever the run carried non-fatal warnings. That code path (`showSuccess`) runs only after a genuinely successful setup — every failure is routed to the error path instead — so the warning styling misrepresented a success as a problem. A successful run should read as informational; the warning count belongs in the message text and the individual warnings behind **View Details**.

## What

- **`pythonSetupDeps.ts`** — `showSuccess` now always uses `window.showInformationMessage`, dropping the branch to `showWarningMessage`.
- **`setupSummary.ts`** — `formatSetupNotification` no longer returns `isWarning`. With that flag gone the single-field return wrapper had no reason to exist, so it now returns the message string directly. The message still names the warning count (e.g. "…with 2 warnings…").

No behavior change to the error path, persisted state, settings, or message contract — this is purely notification severity, so there is no backward-compatibility concern.

## Verification

- `yarn test:unit` — 882 passing, 0 failing (after fetching the bundled CLI).
- `yarn test:lint` — clean.
- Updated the `showSuccess` test to assert an info toast (never a warning) when the run had warnings, and the `formatSetupNotification` tests for the new string return.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/controllers/pythonSetupDeps.test.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/pythonSetupDeps.test.ts
@@ -809,14 +809,17 @@ describe("makePythonSetupDeps showSuccess", () => {
         expect(warnShownWith).to.have.length(0);
     });
 
-    it("raises a warning toast when the run had warnings", async () => {
+    it("raises an info toast (never a warning) when the run had warnings", async () => {
         const deps = makePythonSetupDeps(makeWiring());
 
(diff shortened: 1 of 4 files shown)


rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `fc494db0`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32457975217)

misha-db (approved):

PR #2139, opened 2026-08-21
feat(python-setup): narrate live progress during environment setup

rugpanov:
## Problem

The uv **Setting up Python environment** run shows a mute, indeterminate spinner for a minute or more. Under `--output json` the CLI streams no progress — it buffers everything and prints a single result object at the end — so the notification sat silent, with no sign it was alive or how long it had been running.

## Change

Add `python-setup/utils/setupProgress.ts`, a pure module that turns the one hard fact we have during the run — **elapsed time** — into an honest, informative progress line:

- Walks the fast **leading phases** in order (3.5 s each), mirroring the CLI's own spinner labels: `checking prerequisites…`, `resolving your Databricks compute…`, `fetching matching versions and constraints…`, `updating pyproject.toml…`.
- Then dwells on the dominant **provision** phase, rotating its real uv sub-steps (6 s each, looping): `installing the matching Python version…`, `creating the virtual environment…`, `resolving the dependency graph…`, `downloading databricks-connect and dependencies…`, `building and installing packages with uv…`, `syncing your project's dependencies…`.
- Appends an `m:ss` **elapsed counter**, updated once a second.

It **never shows a percentage and never claims a phase finished** — without a live signal from the CLI that can't be observed, so the copy stays truthful (real work, real order, estimated timing).

The ticker lives in `withElapsedProgress` (timer/clock injected so it's unit-testable, and the work deferred so a synchronous throw still clears the interval). The `withProgress` wiring in `pythonSetupDeps.ts` calls it. The orchestrator and CLI client are untouched — no new CLI args, no `--output json` change.

## Interim, but not throwaway

This is the extension-only interim. The proper fix is a structured progress stream from the CLI; when that lands, the observed phase feeds this same rendering and the leading-phase time estimate drops away — the elapsed counter and provision rotation stay as the fallback for older bundled CLIs.

## Testing

- Co-located unit tests for the pure module (18 cases): elapsed formatting incl. minute roll-over and negative/non-finite clamp, leading-phase walk, provision rotation and looping, composed message, and `withElapsedProgress` reporting + interval cleanup on resolve / reject / synchronous throw.
- `tsc -b` clean, eslint clean, prettier clean.


rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `cf1d0e67`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32462577633)

rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `948e3c21`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32465137896)

rugpanov:
🤖 Integration tests ❌ **5 of 41 test jobs failed** for `70b34665` (36 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32479739940)

misha-db (approved):

PR #2150, opened 2026-08-25
feat(aitools): flag agents that can't be installed, in the picker and the Agents tree

misha-db:
## What

When installing Databricks AI tools, some agents can't be installed for the
chosen scope — either they aren't detected on the machine, or they only support
global-scope installs (the CLI's `supports_project_scope`). Previously the
install picker let you select them anyway (the install would then fail), and the
Agents list in the configuration view gave no indication of why an agent was
unavailable.

This PR makes both surfaces explain the situation up front, using one shared
decision so they stay in lockstep.

## Changes

**Install picker** (`AiToolsCommands`)
- Deactivate un-installable agents: they start unchecked and can't be selected
  (VS Code has no true per-item disable, so the picker strips them from the
  selection), instead of being selectable and failing at install time.
- Annotate each with a short inline reason ("Not detected" / "Only supports
  global scope").

**Agents tree** (`AiToolsComponent`)
- Render blocked agents with a muted circle-slash icon, an inline reason
  ("Not detected" / "Global scope only"), an explanatory tooltip, and no install
  affordance — distinct from installed (green check + version) and installable
  ("Not installed" + inline install button) rows.

**Shared logic** (`AiToolsModel`)
- Extract the block decision into `agentInstallBlockReason()`, returning
  `"notDetected"`, `"scopeUnsupported"`, or `undefined`. Both the picker and the
  tree consume it so the two never drift.
- Plumb `supports_project_scope` through `CliWrapper` into the agent status.

## Tests

- `AiToolsModel.test.ts`: unit tests for `agentInstallBlockReason` across the
  detected × scope matrix.
- `AiToolsCommands.test.ts`: picker deactivates undetected and scope-unsupported
  agents, annotates them, and strips them from the accepted selection.
- `AiToolsComponent.test.ts`: tree renders blocked rows (muted icon, reason,
  no command) and keeps global-only agents installable at global scope.
- Full AiTools unit suite passes (99 tests).


rugpanov:
🤖 Integration tests ❌ **4 of 41 test jobs failed** for `c1d67fbe` (37 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32850036303)

packages/databricks-vscode/src/aitools/AiToolsManager.ts
```diff
@@ -82,6 +82,7 @@ function computeAgentsStatuses(
         id: agent.name,
         type: agent.managed ? "plugin" : "skills-only",
         detected: agent.detected,
+        supportsProjectScope: agent.supports_project_scope,
```
rugpanov: With an older/stale CLI that omits `supports_project_scope`, this maps to `undefined`, and `agentInstallBlockReason` then reads `!undefined === true` — blocking every detected agent at project scope. Suggest defaulting absence to "supported" (and the interface field could be optional to match the real runtime shape). Worth a small omitted-field regression test too.

```suggestion
        supportsProjectScope: agent.supports_project_scope ?? true,
```

packages/databricks-vscode/src/aitools/AiToolsModel.ts
```diff
@@ -27,6 +33,27 @@ export interface AiToolsAgentStatus {
     skillsOnly?: boolean;
 }
 
+export type AgentInstallBlockReason = "notDetected" | "scopeUnsupported";
+
+/**
+ * Why an agent can't be installed at `scope`, or undefined if it can. Shared by
+ * the install picker ({@link AiToolsAgentStatus} in AiToolsCommands) and the
```
rugpanov: Minor: this `{@link AiToolsAgentStatus}` points at the status interface, but the sentence is about the install picker consuming this helper — did you mean `{@link describeAgent}`?

packages/databricks-vscode/src/ui/configuration-view/AiToolsComponent.ts
```diff
@@ -29,6 +40,66 @@ function isHiddenAgent(agentId: string): boolean {
...
+            description: "Not detected",
+            tooltip: `${agent.displayName} was not detected on this machine.`,
+            iconPath: blockedIcon(),
+            contextValue: getContextValue("agent.blocked"),
+            installable: false,
+        };
+    }
+    if (reason === "scopeUnsupported") {
+        return {
+            description: "Global scope only",
```
rugpanov: The tree says "Global scope only" here while the picker (`AiToolsCommands.ts`) says "Only supports global scope" for the same condition. Since the block decision is now shared, worth aligning the wording so the two surfaces read identically.

rugpanov (changes requested):
**Please do not land until the CLI bump.**

Nice cleanup — the shared `agentInstallBlockReason` decision keeps the picker and the tree in lockstep, and the three-state tree rendering reads well. A couple of things before this can land:

**Prerequisite: the bundled CLI must emit `supports_project_scope` first.** The currently bundled CLI (`v1.13.0`) doesn't emit this field yet — `aitools list --output json` returns only `name, display_name, managed, detected, installed`. That's what's failing the unit suite (`CliWrapper.test.ts` asserts the field is a boolean), and it isn't fixable from this PR alone: it needs the CLI change to land, then a `cli.version` bump here so the shipped binary actually returns the field.

I've left an inline suggestion to also default the mapped field, so a *stale* CLI can't silently break project-scope installs (`bin/` is gitignored and isn't re-fetched on pull).

**One more for awareness (optional):** the tree correctly hides the install affordance on blocked rows, but the `installAgent` command handler doesn't re-check `agentInstallBlockReason` before installing — a stale node or a direct command invocation could still attempt an install the block was meant to prevent. Fine to leave to the CLI to reject; just flagging.

The Integration Tests failure is an unrelated `setup_local` cluster-start flake ("Failed to start the cluster: too many attempts"), not caused by this change.

rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `94072eaf`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32860453144)

rugpanov (approved):

rclarey (approved):

rugpanov:
🤖 Integration tests ❌ **2 of 41 test jobs failed** for `564bf366` (39 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32947703440)

PR #2151, opened 2026-08-25
Add performance telemetry for the legacy databricks-connect install

rugpanov:
## What

The legacy (pre-VPEX) Python environment setup — `EnvironmentDependenciesInstaller.install()`, reached via the `databricks.environment.setup` / `databricks.environment.reinstallDBConnect` commands — emits no success or duration telemetry.

Its only signal today is the generic `commandExecution` event, which sets `success=false` only when the command handler throws. But `install()` catches its own errors (shows a message, then returns normally), so the legacy flow reports success regardless of whether `databricks-connect` actually installed. There is no way to measure how long the install takes or how often it fails, and no way to compare it against the uv-native flow, which already emits `python_env.setup.attempt` / `python_env.setup.result`.

## Change

Emit a dedicated `python_env.dbconnect_install` event with:

- `outcome`: `ok` | `failed`
- `duration`: wall-clock ms, measured around the install itself — so it reflects the pip work, not the interpreter/version prompts answered beforehand.

A distinct event name (rather than reusing `python_env.setup.result`) keeps the legacy and uv-native flows cleanly separable in analysis and avoids overloading the uv-native result schema, whose failure phases/error codes have no legacy analogue. Categorical/numeric data only — no versions, paths, or names.

## Testing

- New co-located unit test `EnvironmentDependenciesInstaller.test.ts`: a successful install records `outcome: "ok"` with a `duration` metric; a throwing install records `outcome: "failed"`.
- `yarn test:lint` clean; the new unit tests pass. The pre-existing `CliWrapper` / `packageJsonUtils` failures in a fresh dev clone stem from the gitignored `bin/databricks` binary and version pins, and are unrelated to this change.

### DIFF
--- a/packages/databricks-vscode/src/telemetry/constants.ts
+++ b/packages/databricks-vscode/src/telemetry/constants.ts
@@ -29,6 +29,7 @@ export enum Events {
     PYTHON_ENV_SETUP_RESULT = "python_env.setup.result",
     PYTHON_ENV_DRIFT = "python_env.drift",
     PYTHON_ENV_ADOPTION = "python_env.adoption",
+    PYTHON_ENV_DBCONNECT_INSTALL = "python_env.dbconnect_install",
     AITOOLS_INSTALL = "aitoolsInstall",
     AITOOLS_UPDATE = "aitoolsUpdate",
     AITOOLS_UNINSTALL = "aitoolsUninstall",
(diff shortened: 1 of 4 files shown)


rugpanov:
🤖 Integration tests triggered for `b0839df6` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32852691248)

rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `d12252ea`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/32854924834)

packages/databricks-vscode/src/telemetry/constants.ts
```diff
@@ -617,6 +629,25 @@ export class EventTypes {
                 "none), so adoption can be sliced by compute. No cluster IDs or names",
         },
     };
+    [Events.PYTHON_ENV_DBCONNECT_INSTALL]: EventType<
+        {
+            outcome: DbConnectInstallOutcome;
```
rclarey, line 655: nit: why not `success: boolean` if there is only success or failure?

rclarey (approved):

misha-db (approved):

packages/databricks-vscode/src/telemetry/constants.ts
rugpanov: Good question! I kept it as a string `outcome` rather than `success: boolean` to line up with the uv-native flow's `python_env.setup.result`, whose field is also named `outcome` (`ok`/`failed`/`cancelled`/…). Since this event exists to compare the legacy vs uv-native setup flows, sharing the field name and the `ok`/`failed` vocabulary lets a dashboard UNION/compare both events on one column instead of translating boolean↔enum in every query. It also stays forward-compatible if a third outcome ever becomes reachable here. Happy to switch to a boolean if you'd prefer.

PR #2163, opened 2026-08-31
Prompt re-login on setup-local session expiry instead of a hard error

rugpanov:
## Why

When `databricks environments setup-local` aborts because the active profile's session can no longer be refreshed, the extension surfaced the raw CLI text as a hard error notification (which VS Code decorates with "Report this problem") and force-revealed the setup output channel. An expired session is an expected, self-service condition — the user just needs to log in again — not a defect worth reporting.

## What

- Add `isReauthRequiredError`, a precision-tuned matcher for the CLI's reauthenticate signal. It requires **both** the CLI's own `databricks auth login` remediation **and** a matching cause (`reauthenticate` / `re-authenticate` / `refresh token is invalid` / `access token could not be retrieved`), with whitespace normalized so a cause split across lines still matches.
- Use it as a **positive gate** in the setup-local spawn/parse `catch`: on a match, show a re-login prompt — a warning (not an error) with a **Login** button that runs the extension's existing re-auth flow (`databricks.connection.configureLogin`) — and record the run as not report-worthy (`not_started`, `reportOffered: false`). No log-channel reveal and no auto-retry: the user re-runs setup once connected.
- Anything the matcher does not recognize — a genuine spawn/parse defect, or a transient network failure — **falls through to the existing report path**. So a false negative (a drifted CLI wording this misses) is never worse than today's behavior, and a network error is never mislabeled as an expired session.

## Testing

- `authErrors` matcher: positive cases (verbatim CLI abort, hyphenated spelling, whitespace-split cause, case-insensitive) and precision negatives (generic parse error, network/host mention, index-fetch failure, cause-without-remediation, empty/non-string).
- Controller gate: a reauth-signalled abort shows the prompt with no report action (`not_started` / `reportOffered: false`); a no-signal abort stays on the report path.
- Deps: the prompt is a warning with a Login action that runs the re-auth command and does not reveal the log channel.
- A gateway↔matcher integration test pins the CLI-stderr format the gate keys on, so a change to either side is caught in tests rather than in production.
- `yarn build`, full unit suite (1034 passing), and `yarn test:lint` all pass.

## Note

An earlier revision of this branch used an SDK session probe; it was replaced with this positive-gate matcher after review flagged that a blind probe misclassifies transient network failures as an expired session.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
@@ -125,6 +125,7 @@ function makeDeps(
         adoptInterpreter: async () => {},
         saveState: () => {},
         notify: async () => {},
+        showReauthPrompt: async () => {},
         showError: async () => {},
         showSuccess: async () => {},
         reportEnvironment: {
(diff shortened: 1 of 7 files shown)


rugpanov:
🤖 Integration tests ❌ **1 of 41 test jobs failed** for `cef25ac2` (40 passed).
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/33403415107)

rugpanov:
🤖 Integration tests triggered for `915168b2` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/33486085583)

rugpanov:
🤖 Integration tests triggered for `0050a09d` — ⏳ running.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/33487420752)

rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `c2ddf840`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/33487908972)

misha-db (approved):

PR #2171, opened 2026-09-03
fix(run): preserve script stdout when an Upload-and-Run file fails

rugpanov:
## Problem

When running a Python file on a cluster ("Upload and Run File"), if the script raises an exception, any output it printed **before** failing is lost. The `1.2/commands` execution API returns only the traceback on an error result and drops the accumulated stdout, so the output panel shows the traceback with none of the preceding `print` output — making failures hard to debug.

## Fix

Because a single command yields a single result whose error variant has no stdout field, the captured output has to be preserved where it still exists — in the bootstrap that runs the file.

- **`resources/python/bootstrap.py`** buffers the script's stdout while it runs. On an exception that follows printed output, it emits the captured stdout, a boundary marker, then a structured JSON traceback containing only the user's frames. Errors with no prior output re-raise unchanged; successful runs and `SystemExit`/`KeyboardInterrupt` replay the buffer, so no output is lost.
- **`src/run/ErrorParser.ts`** adds `renderErrorEnvelope()` to turn that JSON payload into renderable frames (red-highlighted summary line), plus a shared boundary constant.
- **`src/run/DatabricksRuntime.ts`** splits the text result on the boundary — emitting the stdout first, then the traceback remapped to local files (so frames stay clickable) — and shares the frame-emit logic between the new path and the existing error path.

The new behaviour triggers **only** on print-then-crash. Successful runs and errors with no prior output render exactly as before.

## Verification

- `yarn build` (full `tsc`) clean; `eslint` and `prettier` clean; `bootstrap.py` compiles.
- Unit tests 4/4 pass (2 new for `renderErrorEnvelope`).
- End-to-end against a live cluster running the real `bootstrap.py`:
  - **print-then-crash** → stdout preserved, followed by a clickable traceback with only the user frame;
  - **success + output** → plain text output, unchanged;
  - **error with no prior output** → native error result, unchanged.

### DIFF
--- a/packages/databricks-vscode/resources/python/bootstrap.py
+++ b/packages/databricks-vscode/resources/python/bootstrap.py
@@ -1,12 +1,16 @@
 import runpy
 import sys
 import os
+import io
+import json
+import traceback
 
 # values will be injected by the runner
 python_file = "PYTHON_FILE"
 repo_path = "REPO_PATH"
 args = []
 env = {}
+stdout_error_boundary = "STDOUT_ERROR_BOUNDARY"
 
 # change working directory
 os.chdir(os.path.dirname(python_file))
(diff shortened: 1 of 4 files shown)


rugpanov:
🤖 Integration tests ✅ **all 41 test jobs passed** for `f039406c`.
[View run](https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/33762245288)

misha-db (approved):

PR #2176, opened 2026-09-07
Migrate setup-local invocation to orthogonal --no-constraints/--no-dbconnect flags

rugpanov:
## Summary

`databricks environments setup-local` has replaced the single behavioral flag
`--constraints-only` with two **orthogonal** negative flags:

| Flag | Effect |
|---|---|
| `--no-constraints` | Skip writing the remote Python-version / dependency pins. |
| `--no-dbconnect` | Skip adding the `databricks-connect` dependency. |

The extension still modeled the old single `mode` enum and only emitted
`--constraints-only`. This migrates the invocation to the new flags so the
setup-options picker (a follow-up) can map its tiers straight onto composable
capabilities.

## Changes

- Replace `SetupLocalInvocation.mode` with two orthogonal optional booleans —
  `skipConstraints` → `--no-constraints`, `skipDbconnect` → `--no-dbconnect`.
  `buildSetupLocalArgs` stays pure and deterministic (fixed flag order); the
  `--constraints-only` emission is dropped.
- Update the two call sites (the real setup run and the drift dry-run) to the
  flag-free default, and derive the attempt telemetry's `mode` dimension from
  `skipDbconnect` so that event is unchanged.
- Rework the arg unit tests around the new mapping (each flag alone, both
  together, default emits neither, and never `--constraints-only`).

## No behavior change

Both callers run the default (Full) setup, which emits **no** behavioral flags —
exactly as before. `--constraints-only` was only ever emitted in a mode no caller
used, so nothing changes in the argv sent to the CLI today; the new flags become
reachable only when the picker is introduced.

## Draft / follow-up

Kept as a **draft** pending the bundled-CLI dependency: the `cli.version` bump +
re-fetch will land here once a CLI **release** carrying the new flags is
available (they are merged to CLI `main` but not yet in a tagged release). The
change is safe to run against the currently bundled CLI in the meantime because
it emits no new flags.

## Testing

- `tsc --noEmit`: clean.
- Unit suite: `buildSetupLocalArgs`, `PythonSetupCliClient`, and
  `PythonSetupEnvironmentSetup` (setup + telemetry) all green.
- eslint + prettier: clean on all changed files.

### DIFF
--- a/packages/databricks-vscode/src/extension.ts
+++ b/packages/databricks-vscode/src/extension.ts
@@ -1170,7 +1170,6 @@ export async function activate(
             try {
                 const result = await pythonSetupClient.run(
                     {
-                        mode: "default",
                         dryRun: true,
                         compute: resolution.compute,
                     },
--- a/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.ts
@@ -634,10 +634,8 @@ export class PythonSetupEnvironmentSetup implements Disposable {
                 targetType: compute.kind,
                 serverlessVersion:
                     compute.kind === "serverless" ? compute.version : undefined,
-                // The attempt event keeps its default/constraints-only mode
-                // dimension; derive it from the orthogonal flag that drops
-                // databricks-connect (the CLI treats --no-dbconnect and the old
-                // --constraints-only as the same mode).
+                // --no-dbconnect is the orthogonal spelling of the legacy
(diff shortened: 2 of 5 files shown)


rugpanov:
🤖 Integration tests running for `a8453f9b` — ⏳.

rugpanov:
🤖 Integration tests ✅ **passed** for `c95891a4` (the earlier failure was an unrelated `deploy_and_run_pipeline` remote-timeout flake; it passed on shard re-run).
[View run](https://go/deco-tests/34108506399)

misha-db (approved):

rugpanov:
🤖 Integration tests ❌ **failed** for `57896939`.
[View run](https://go/deco-tests/34131461214)

PR #2178, opened 2026-09-07
Add HTTP proxy and corporate-CA support for SDK and bundled CLI

misha-db:
## What

Makes the extension work behind corporate HTTP/HTTPS proxies and against
workspaces whose TLS certificate is signed by an internal CA — for both the
in-process Databricks JS SDK and the bundled Go CLI.

## Why

Users on corp-managed machines (proxy-only networks, TLS-intercepting proxies,
internal CAs) currently hit two problems:

1. **Proxy not applied.** The SDK never routed through a proxy, and the CLI only
   picked one up from OS env vars — never from the VS Code `http.proxy` setting.
2. **Corp CA not trusted.** The SDK pins its own HTTP agent (bypassing VS Code's
   global TLS patching), so an OS-trusted internal CA was ignored and SDK calls
   failed with `SELF_SIGNED_CERT_IN_CHAIN` / `UNABLE_TO_GET_ISSUER_CERT_LOCALLY`.
   The bundled Go CLI was unaffected (it reads the OS store natively).

## What changed

**Proxy resolution (SDK + CLI)**
- New `src/utils/network/proxyAgent.ts` builds a proxy- and CA-aware agent via
  `@vscode/proxy-agent`, resolving the proxy the way VS Code core does:
  `http.proxy` setting first, then `http(s)_proxy` env vars, honoring
  `http.noProxy` / `NO_PROXY`. `createWorkspaceClient` is now the single place
  the extension constructs a `WorkspaceClient`; the auth checks (`AuthProvider`,
  `AzureCliCheck`, `DatabricksCliCheck`) route through it.
- `getProxyEnvVars` (forwarded to the CLI subprocess) now also reads the VS Code
  `http.proxy` / `http.noProxy` settings, not just OS env vars — the Go CLI
  honors a proxy only through these env vars, so this keeps the CLI and SDK
  resolving the same proxy from the same inputs.

**Certificate trust (SDK)**
 - The agent's ca is built by merging Node's bundled roots (tls.rootCertificates) + the OS trust store + an optional user PEM + the NODE_EXTRA_CA_CERTS bundle, deduped. Previously the OS store replaced the bundled roots, which could break public-root TLS; NODE_EXTRA_CA_CERTS is folded in because setting an explicit ca otherwise drops the trust Node would normally add from that env var.
 - The OS trust store is read via @vscode/proxy-agent's native readers on all supported Node versions (security on macOS, PEM bundle files on Linux). On Windows the store would be read through the native @vscode/windows-ca-certs addon, which we deliberately do not ship (it publishes no prebuilds and can't be cross-compiled on the Linux release runner) — so on Windows the SDK falls back to Node's bundled roots, and internal-CA trust goes through the databricks.proxy.caCert setting below.
  - When a proxy is configured, the merged CA and strict-SSL setting are applied to the endpoint (workspace host) TLS handshake, not just the proxy connection, via a small HttpsProxyAgent subclass — otherwise https-proxy-agent would validate the host against Node's bundled roots only.
- New databricks.proxy.caCert setting — an absolute path to a PEM bundle — as a cross-platform escape hatch (and the primary internal-CA path on Windows), merged onto the trust store.

**Strict SSL**
- New `databricks.proxy.strictSSL` setting overrides the built-in
  `http.proxyStrictSSL` (both default to verification on). Activation and the SDK
  client path now resolve strict-SSL through the same layered setting, so the
  `DATABRICKS_SDK_PROXY_STRICT_SSL` env var (read by the SDK fetch path) is
  consistent and defaults to *on* when nothing is configured.

## Graceful degradation

Every failure path is recoverable: an unreadable OS store or missing native
module falls back to Node's bundled roots, an unreadable `caCert` path is logged
and ignored, and none of these break the agent. TLS verification stays on by
default throughout.

## Tests

- Unit tests for proxy resolution, CA merge/dedup, `caCert` (success + fallback +
  unreadable path), the version-aware flag, and strict-SSL env wiring.
- `yarn workspace databricks run test:unit` — all green (1078 passing).

**Not covered by CI:** the old-Node Windows corp-CA path requires the native
module to actually be present, which depends on the win32 VSIX being built on a
Windows host. Verify the release pipeline before relying on it; the
`databricks.proxy.caCert` setting is the fallback if it isn't shipped.

### DIFF
--- a/packages/databricks-vscode/src/utils/network/proxyAgent.test.ts
+++ b/packages/databricks-vscode/src/utils/network/proxyAgent.test.ts
@@ -120,6 +120,32 @@ describe(__filename, () => {
                 false
             );
         });
+
+        it("falls back to Node's bundled CAs when the system store can't be read", async () => {
+            // @vscode/proxy-agent reads the store via tls.getCACertificates,
+            // which is absent/throws on older runtimes. Simulate that failure
+            // on the shared (required) tls module proxy-agent also sees.
+            // eslint-disable-next-line @typescript-eslint/no-require-imports
+            const tls = require("node:tls");
+            const original = tls.getCACertificates;
+            tls.getCACertificates = () => {
+                throw new Error("getCACertificates unavailable");
+            };
+            try {
+                const agent = (await getDatabricksHttpAgent(
+                    new URL("https://example.com")
+                )) as https.Agent;
+                // `ca` must be omitted (not [] / undefined) so Node keeps its
+                // bundled roots instead of trusting nothing.
+                assert.ok(!("ca" in (agent.options as https.AgentOptions)));
+                assert.strictEqual(
+                    (agent.options as https.AgentOptions).rejectUnauthorized,
+                    true
+                );
+            } finally {
+                tls.getCACertificates = original;
+            }
+        });
     });
 
     describe("createWorkspaceClient", () => {
--- a/packages/databricks-vscode/src/utils/network/proxyAgent.test.ts
+++ b/packages/databricks-vscode/src/utils/network/proxyAgent.test.ts
@@ -5,8 +5,10 @@ import {reset, spy, when} from "ts-mockito";
 import {HttpsProxyAgent} from "https-proxy-agent";
 import {HttpProxyAgent} from "http-proxy-agent";
 import {workspaceConfigs} from "../../vscode-objs/WorkspaceConfigs";
+import {WorkspaceClient} from "@databricks/sdk-experimental";
 import {
     applyProxyStrictSSLEnv,
+    createWorkspaceClient,
     getDatabricksHttpAgent,
     resetProxyAgentCaches,
     strictSSL,
@@ -117,6 +119,19 @@ describe(__filename, () => {
                 (agent.options as https.AgentOptions).rejectUnauthorized,
                 false
             );
+        });
+    });
+
+    describe("createWorkspaceClient", () => {
+        it("builds a client tagged with the extension product and a proxy-aware agent", async () => {
+            const client = await createWorkspaceClient(
+                {host: "https://example.com", authType: "pat", token: "t"},
+                new URL("https://example.com")
+            );
+            assert.ok(client instanceof WorkspaceClient);
+            assert.strictEqual(client.apiClient.product, "databricks-vscode");
+            const agent = await client.apiClient.getAgent();
+            assert.ok(agent instanceof https.Agent);
         });
     });
 });
--- a/packages/databricks-vscode/src/utils/network/proxyAgent.test.ts
+++ b/packages/databricks-vscode/src/utils/network/proxyAgent.test.ts
@@ -1,0 +1,122 @@
+import * as assert from "assert";
+import * as https from "node:https";
+import * as http from "node:http";
+import {reset, spy, when} from "ts-mockito";
+import {HttpsProxyAgent} from "https-proxy-agent";
+import {HttpProxyAgent} from "http-proxy-agent";
+import {workspaceConfigs} from "../../vscode-objs/WorkspaceConfigs";
+import {
+    applyProxyStrictSSLEnv,
+    getDatabricksHttpAgent,
+    resetProxyAgentCaches,
+    strictSSL,
+} from "./proxyAgent";
+
+describe(__filename, () => {
+    let configsSpy: typeof workspaceConfigs;
+    let existingEnv: NodeJS.ProcessEnv;
+
+    beforeEach(() => {
+        existingEnv = Object.assign({}, process.env);
+        resetProxyAgentCaches();
+        configsSpy = spy(workspaceConfigs);
+        // Defaults: strict SSL on, no proxy configured.
+        when(configsSpy.proxyStrictSSL).thenReturn(true);
+        when(configsSpy.httpProxy).thenReturn(undefined);
+        when(configsSpy.httpNoProxy).thenReturn([]);
+    });
+
+    afterEach(() => {
+        reset(configsSpy);
+        process.env = existingEnv;
+    });
+
+    describe("strictSSL / applyProxyStrictSSLEnv", () => {
+        it("reflects the effective setting", () => {
+            when(configsSpy.proxyStrictSSL).thenReturn(true);
+            assert.strictEqual(strictSSL(), true);
+
+            when(configsSpy.proxyStrictSSL).thenReturn(false);
+            assert.strictEqual(strictSSL(), false);
+        });
+
+        it("sets DATABRICKS_SDK_PROXY_STRICT_SSL only when disabled", () => {
+            when(configsSpy.proxyStrictSSL).thenReturn(false);
+            applyProxyStrictSSLEnv();
+            assert.strictEqual(
+                process.env.DATABRICKS_SDK_PROXY_STRICT_SSL,
+                "false"
+            );
+
+            when(configsSpy.proxyStrictSSL).thenReturn(true);
+            applyProxyStrictSSLEnv();
+            assert.strictEqual(
+                process.env.DATABRICKS_SDK_PROXY_STRICT_SSL,
+                undefined
+            );
+        });
+    });
+
+    describe("getDatabricksHttpAgent", () => {
+        it("returns a plain https agent when no proxy is configured", async () => {
+            const agent = await getDatabricksHttpAgent(
+                new URL("https://example.com")
+            );
+            assert.ok(agent instanceof https.Agent);
+            assert.ok(!(agent instanceof HttpsProxyAgent));
+        });
+
+        it("returns an https proxy agent when http.proxy is set", async () => {
+            when(configsSpy.httpProxy).thenReturn("http://127.0.0.1:8080");
+            const agent = await getDatabricksHttpAgent(
+                new URL("https://example.com")
+            );
+            assert.ok(agent instanceof HttpsProxyAgent);
+        });
+
+        it("honours the http.proxy setting for http hosts", async () => {
+            when(configsSpy.httpProxy).thenReturn("http://127.0.0.1:8080");
+            const agent = await getDatabricksHttpAgent(
+                new URL("http://example.com")
+            );
+            assert.ok(agent instanceof HttpProxyAgent);
+        });
+
+        it("returns a plain agent when the host matches noProxy", async () => {
+            when(configsSpy.httpProxy).thenReturn("http://127.0.0.1:8080");
+            when(configsSpy.httpNoProxy).thenReturn(["example.com"]);
+            const agent = await getDatabricksHttpAgent(
+                new URL("https://example.com")
+            );
+            assert.ok(agent instanceof https.Agent);
+            assert.ok(!(agent instanceof HttpsProxyAgent));
+        });
+
+        it("falls back to a plain http agent for http hosts without a proxy", async () => {
+            const agent = await getDatabricksHttpAgent(
+                new URL("http://example.com")
+            );
+            assert.ok(agent instanceof http.Agent);
+            assert.ok(!(agent instanceof HttpProxyAgent));
+        });
+
+        it("loads the system certificate trust store onto https agents", async () => {
+            const agent = (await getDatabricksHttpAgent(
+                new URL("https://example.com")
+            )) as https.Agent;
+            const ca = (agent.options as https.AgentOptions).ca;
+            assert.ok(Array.isArray(ca) && ca.length > 0);
+        });
+
+        it("disables certificate verification when strict SSL is off", async () => {
+            when(configsSpy.proxyStrictSSL).thenReturn(false);
+            const agent = (await getDatabricksHttpAgent(
+                new URL("https://example.com")
+            )) as https.Agent;
+            assert.strictEqual(
+                (agent.options as https.AgentOptions).rejectUnauthorized,
+                false
+            );
+        });
+    });
+});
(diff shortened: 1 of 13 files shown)


rugpanov:
🤖 Integration tests ❌ **failed** for `25219fd6`.
[View run](https://go/deco-tests/34116013157)

rugpanov:
🤖 Integration tests ❌ **failed** for `a6630e11`.
[View run](https://go/deco-tests/34120812882)

rugpanov:
🤖 Integration tests running for `0727387b` — ⏳.

rugpanov:
🤖 Integration tests ❌ **failed** for `705b4f01`.
[View run](https://go/deco-tests/34335377898)

rugpanov:
🤖 Integration tests ❌ **failed** for `80470b33`.
[View run](https://go/deco-tests/34344451741)

rugpanov:
🤖 Integration tests ❌ **failed** for `57876145`.
[View run](https://go/deco-tests/34359613315)

rugpanov:
🤖 Integration tests ❌ **failed** for `f1ad46a5`.
[View run](https://go/deco-tests/34992018982)

rugpanov:
🤖 Integration tests ❌ **failed** for `737f4ecf`.
[View run](https://go/deco-tests/35109641753)

rugpanov:
🤖 Integration tests ❌ **failed** for `158948d7`.
[View run](https://go/deco-tests/35206915056)

rugpanov:
🤖 Integration tests ❌ **failed** for `a5e67de0`.
[View run](https://go/deco-tests/35225464456)

rugpanov:
🤖 Integration tests ❌ **failed** for `70cad6cc`.
[View run](https://go/deco-tests/35246279028)

rugpanov:
🤖 Integration tests ❌ **failed** for `45d97298`.
[View run](https://go/deco-tests/35329625396)

abdelrahmanomar1515:
How simpler would this be if we the SDK stops overriding the VSCode patched ProxyAgent?
It wouldn't solve the issue where VSCode settings are not passed to the CLI, but is there anything else that would still be broken?

packages/databricks-vscode/src/utils/network/proxyAgent.ts
```diff
@@ -0,0 +1,314 @@
...
+    const agentOptions: https.AgentOptions = {
+        keepAlive: true,
+        keepAliveMsecs: KEEP_ALIVE_MSECS,
+        timeout: (httpTimeoutSeconds || DEFAULT_HTTP_TIMEOUT_SECONDS) * 1000,
+        ...(usesTls ? {rejectUnauthorized, ...(ca ? {ca} : {})} : {}),
+    };
+
+    if (proxyUrl) {
+        return isHttps
+            ? new HttpsProxyAgent(proxyUrl, agentOptions)
```
rugpanov: **Custom CA is not applied to the endpoint (workspace host) TLS handshake when a proxy is configured — I think this defeats the PR's main scenario.**

Here we hand the merged `ca` to `new HttpsProxyAgent(proxyUrl, agentOptions)`. In `https-proxy-agent@7.0.6` the constructor options become `this.connectOpts`, which govern **only** the TLS connection to the *proxy* (`https-proxy-agent/dist/index.js:96`). The final handshake to the Databricks host is:

```js
// https-proxy-agent/dist/index.js:132
return tls.connect({ ...omit(setServernameFromNonIpHost(opts), 'host','path','port'), socket });
```

`opts` here comes from `agent-base`'s `createSocket` as `{ ...options, secureEndpoint }` (`agent-base/dist/index.js:124-128`) — it does **not** merge the agent's constructor options. And `options` is the SDK's per-request object, which (`@databricks/sdk-experimental` `fetch.js:155-163`) carries `agent` + `rejectUnauthorized` but **no `ca`**. So the endpoint gets validated against Node's bundled roots only; `databricks.proxy.caCert` and the OS trust store are silently ignored whenever a proxy is set.

Net effect: a user behind a TLS-intercepting proxy with an internal CA can't keep verification on and trust their CA — the endpoint handshake fails with `SELF_SIGNED_CERT_IN_CHAIN` / `UNABLE_TO_VERIFY_LEAF_SIGNATURE`. The only workaround is `databricks.proxy.strictSSL=false` (that one *does* propagate, via the env var → per-request `rejectUnauthorized`), i.e. turning verification off entirely. The no-proxy path is fine because `new https.Agent` merges `this.options.ca` into the origin handshake.

Suggested direction: get the trust material onto the endpoint handshake, not just the proxy socket — e.g. fold the configured/OS CA into `NODE_EXTRA_CA_CERTS` (which the SDK's default handshake honors) rather than only onto the agent. And please add a test that inspects the **endpoint** tunnel's `tls.connect` options rather than `agent.connectOpts`.

packages/databricks-vscode/src/utils/network/proxyAgent.ts
```diff
@@ -0,0 +1,314 @@
...
+ * roots) and then extended with the OS trust store and the configured PEM.
+ * Setting `ca` *replaces* Node's defaults, so if we set it to only the extra
+ * certs, public-root TLS would break — hence the merge. When we have nothing to
+ * add (system store unreadable and no `caCert`), return `undefined` so the
+ * caller omits `ca` and Node keeps its defaults.
+ *
+ * Deduped because the OS store commonly re-lists the public roots already in
+ * `tls.rootCertificates`; a `Set` keeps the handed-off list minimal.
+ */
+function buildCaBundle(
```
rugpanov, line 246: **`NODE_EXTRA_CA_CERTS` stops being trusted once we set an explicit `ca`.**

On macOS/Linux the OS-store read almost always returns a non-empty list, so `buildCaBundle` returns a concrete array and the agent gets an explicit `ca`. Node uses an explicit `ca` **in place of** its default store, and `NODE_EXTRA_CA_CERTS` is only merged into that *default* store (it's not in `tls.rootCertificates`). Previously the SDK set no `ca`, so `NODE_EXTRA_CA_CERTS` was honored.

Existing users who trust their corporate CA solely via `NODE_EXTRA_CA_CERTS` (not imported into the OS keychain) will silently lose that trust after upgrading — on the no-proxy path too. Consider folding `NODE_EXTRA_CA_CERTS` file contents into the bundle (deduped), or at minimum calling out the behavior change in the setting docs / CHANGELOG.

packages/databricks-vscode/src/utils/network/proxyAgent.ts
```diff
@@ -0,0 +1,314 @@
...
+import {readFile} from "node:fs/promises";
+import {
+    createProxyResolver,
+    loadSystemCertificates,
+    LogLevel,
+    type Log,
+    type ProxyAgentParams,
+} from "@vscode/proxy-agent";
+import {HttpProxyAgent} from "http-proxy-agent";
+import {HttpsProxyAgent} from "https-proxy-agent";
```
rugpanov, line 13: Nit: `http-proxy-agent` and `https-proxy-agent` are imported directly here but neither is declared in `package.json` — they resolve only transitively via `@vscode/proxy-agent`. It builds today because esbuild bundles the hoisted copies, but a transitive bump could relocate/remove them and break the bundle silently. Please add both as explicit dependencies.

packages/databricks-vscode/src/utils/network/proxyAgent.test.ts
```diff
@@ -0,0 +1,349 @@
...
+            const agent = (await getDatabricksHttpAgent(
+                new URL("https://example.com")
+            )) as https.Agent;
+            assert.strictEqual(
+                (agent.options as https.AgentOptions).rejectUnauthorized,
+                false
+            );
+        });
+
+        it("applies TLS options when an http host uses an https proxy", async () => {
```
rugpanov: Nit: this is the only test touching the "https proxy" TLS path, but it asserts `agent.connectOpts.ca` (the **proxy-side** options) against an `http://` target (no endpoint TLS at all), so it passes while the endpoint-trust issue goes uncaught. A regression test should drive an `https://` target through a proxy and assert the CA / `rejectUnauthorized` on the **endpoint** tunnel.

rugpanov (changes requested):
Thanks for tackling this — the layering (settings → env → agent), the graceful CA fallbacks, and the test coverage are well done. But I think the headline capability doesn't actually work when a proxy is configured, plus a backward-compat regression, so requesting changes. Details inline; two non-inline points below.

**Windows: PR description vs. code disagree.** The description says the win32 VSIX "ships the native `@vscode/windows-ca-certs` module … copied into `out/` at package time," but the code does the opposite — `esbuild:base` marks it `--external`, there's no copy step in `scripts/package-vsix.sh`, and the in-code comment in `getSystemCertificates` correctly says it's deliberately not shipped and Windows falls back to bundled roots. Please reconcile the description with the code. As written, a Windows user gets no automatic system-store trust and must set `databricks.proxy.caCert` — and combined with the endpoint-CA issue below, has no working internal-CA path behind a proxy at all.

**Verified safe (noting so they don't come up again):** the `DATABRICKS_SDK_PROXY_STRICT_SSL` change (deleting the var when strict, vs. always setting `"true"`) is correct — the SDK's `fetch.js` treats an unset value as verify-on. Applying a single `http.proxy` to both schemes for the CLI matches VS Code's scheme-agnostic convention. And routing all production `WorkspaceClient` construction through `createWorkspaceClient` checks out (remaining `new WorkspaceClient` are test-only).

rugpanov:
🤖 Integration tests ✅ **passed** for `9ebf46f7`.
[View run](https://go/deco-tests/35354706597)

packages/databricks-vscode/src/utils/network/proxyAgent.ts
misha-db: The fix (proxyAgent.ts): a CaAwareHttpsProxyAgent extends HttpsProxyAgent subclass overrides connect(req, opts) and folds the CA + rejectUnauthorized into opts for the secureEndpoint case — opts (unlike connectOpts) does flow to the endpoint tls.connect. getDatabricksHttpAgent uses it for the proxy + https path (line 358-364 above).

packages/databricks-vscode/src/utils/network/proxyAgent.ts
misha-db: Done

packages/databricks-vscode/src/utils/network/proxyAgent.test.ts
misha-db: Replaced the misleading test with two endpoint-tunnel tests (capture opts passed to super.connect, assert CA membership
  + rejectUnauthorized) plus two NODE_EXTRA_CA_CERTS tests. Verified as genuine regression guards: they fail against the  old
  constructor-only wiring and pass with the subclass fix.

packages/databricks-vscode/src/utils/network/proxyAgent.ts
misha-db: Added loadNodeExtraCaCerts() and folded its file contents into buildCaBundle's deduped merge (Node roots + OS store +
  caCert + NODE_EXTRA_CA_CERTS), with log-and-ignore fallback on an unreadable path.

rugpanov:
🤖 Integration tests running for `fca7db33` — ⏳.

rugpanov (approved):
Approving — all the concerns from the previous review are addressed, and I re-verified the endpoint-CA fix end-to-end.

- **Endpoint CA through a proxy** ✅ `CaAwareHttpsProxyAgent` folds `ca`/`rejectUnauthorized` into the per-request `opts`, which `https-proxy-agent` 7.0.6 forwards into the endpoint `tls.connect` (the `omit` only strips host/path/port). The corp CA now verifies the workspace host, not just the proxy socket.
- **`NODE_EXTRA_CA_CERTS`** ✅ re-read and merged into the bundle, so trust that only lived in the default store survives an explicit `ca`.
- **Deps** ✅ `http-proxy-agent` / `https-proxy-agent` now declared.
- **Windows** ✅ description reconciled with the code (addon deliberately not shipped; falls back to bundled roots + `databricks.proxy.caCert`).
- **Tests** ✅ now assert on the endpoint tunnel, plus `NODE_EXTRA_CA_CERTS` coverage. CI is fully green.

Two optional, non-blocking test nits for whenever you touch this again:
1. The endpoint test stubs `HttpsProxyAgent.prototype.connect`, so it doesn't guard the library contract (that `opts.ca` still reaches the endpoint `tls.connect`) — an upstream `https-proxy-agent` change could pass this suite while corp TLS breaks. A light integration-style check would close that gap.
2. `beforeEach` doesn't clear an inherited `NODE_EXTRA_CA_CERTS`; on a runner that has it set, the "falls back to bundled CAs" assertion would fail. A one-line `delete process.env.NODE_EXTRA_CA_CERTS` in `beforeEach` makes the suite hermetic.

Nice work on the fix.

misha-db:
@abdelrahmanomar1515 
> How simpler would this be if we the SDK stops overriding the VSCode patched ProxyAgent? It wouldn't solve the issue where VSCode settings are not passed to the CLI, but is there anything else that would still be broken?

1. SDK builds it own proxy agent, so it we'll need SDK change in order to fallback to globalAgent
2. And as you noted, the CLI still needs the settings forwarded as env vars regardless. 
We do it here: https://github.com/databricks/databricks-vscode/pull/2178/changes#diff-de007be4c398fd85580ff970672d76e38f6bd9db6bc44eac9a49665bf849a0b5R164
3. Agree that explicit agent approach here is more code, but it's testable and doesn't depend on a host setting or VS Code version.

PR #2181, opened 2026-09-08
feat(python-setup): recover from a constraint conflict with a Retry-as-DB-Connect fallback

rugpanov:
## Why

The uv-native "set up Python environment" **Full** preset pins the cluster's
dependencies, so `uv sync` can fail when those pins conflict with the user's own
dependencies. Today that surfaces as the generic `E_PROVISION` toast ("uv could
not resolve the project's dependencies… adjust your dependencies"), which leaves
the user to work out on their own that dropping the pins — the **DB Connect**
preset (`--no-constraints`) — would resolve it while keeping matched Python +
databricks-connect.

## What

- **`E_PROVISION_CONFLICT`** — a distinct result-model error code for a genuine
  pin-vs-local conflict (a generic sync failure stays `E_PROVISION`), with its
  own actionable copy: *"The cluster dependencies conflict with your local
  dependencies, so uv sync couldn't resolve the environment."*
- **Recovery UX** — on a Full-preset conflict the generic toast is replaced with
  two buttons: **Retry DB Connect setup** and **Open pyproject.toml**. "Show
  Logs" is dropped from this toast (it's self-service — the channel is still
  revealed), and every other provision failure is unchanged.
- **Restore-then-retry** — the conflict fails *after* the runtime pins were
  merged into `pyproject.toml` (`diskMutated: true`, pre-merge file saved to the
  CLI's `backupPath`). So **Retry** first restores that backup, then re-runs with
  the `dbconnect` preset (`--no-constraints`) — `--no-constraints` alone only
  stops the retry from re-adding pins, it wouldn't remove the ones already on
  disk. The restore is atomic (temp sibling + rename) and refuses a `backupPath`
  outside the project. With no `backupPath`, Retry isn't offered and the run
  falls back to the ordinary doc-link handling.
- **`run` action arm** — `PythonSetupErrorAction` gains an in-process-callback
  arm alongside `url`/`command`, so the two runtime-stateful buttons reuse
  `showError`'s existing plumbing (label de-dup, log reveal); `showError` gains an
  `includeShowLogs` option to drop the trailing button for this toast.
- **`runGuarded` / `runResolved`** extracted from `runSetup`, so the retry
  re-enters the flow with the `dbconnect` preset directly — no re-prompt of the
  compute or the preset picker, and guarded so a click can't race an in-flight
  run (and the restore only fires when the retry actually runs). Because
  `dbconnect` skips constraints, the retry can't itself surface a recoverable
[...]

PR #2186, opened 2026-09-09
feat(python-setup): recover from a constraint conflict with a Retry-as-DB-Connect fallback

rugpanov:
## Why

The uv-native "set up Python environment" **Full** preset pins the cluster's
dependencies, so `uv sync` can fail when those pins conflict with the user's own
dependencies. The CLI now emits a distinct **`E_PROVISION_CONFLICT`** for that
case (shipped in **CLI v1.16.0**, already bundled on `main`). Rather than the
generic `E_PROVISION` toast ("uv could not resolve the project's dependencies…
adjust your dependencies"), offer a one-click recovery: re-run as **DB Connect**
(`--no-constraints`), which drops the conflicting cluster-dependency pins while
keeping matched Python + databricks-connect.

## What

- **`E_PROVISION_CONFLICT`** — a distinct result-model error code for a genuine
  pin-vs-local conflict (a generic sync failure stays `E_PROVISION`), with its
  own copy: *"The cluster dependencies conflict with your local dependencies, so
  uv sync couldn't resolve the environment."*
- **Recovery UX** (Full-preset conflict only) — the generic toast is replaced
  with **Retry DB Connect setup** and **Open pyproject.toml**. "Show Logs" is
  dropped from this self-service toast (the channel is still revealed).
- **Restore-then-retry** — the conflict fails *after* the runtime pins were
  merged into `pyproject.toml` (`diskMutated: true`, pre-merge file saved to the
  CLI's `backupPath`). So **Retry** first restores that backup, then re-runs with
  the `dbconnect` preset — `--no-constraints` alone would only stop the retry
  from re-adding pins, not remove those already on disk. The restore is atomic
  (temp sibling + rename) and refuses a `backupPath` resolving outside the
  project. When `backupPath` is absent/empty (e.g. a no-op-merge re-run), Retry
  isn't offered and the run falls back to the ordinary doc-link handling.
- **Guards** — `runGuarded`/`runResolved` let the retry re-enter with the
  `dbconnect` preset directly (no re-prompt), coalesced so a click can't race an
  in-flight run; `dbconnect` skips constraints so the retry can't loop; a stale
  toast's Retry no-ops once the project is ready.
- **Telemetry** — a `conflict_retry` value on the existing
  `python_env.setup.attempt` `trigger` dimension marks recovery runs, paired with
  the `python_env.setup.result` outcome for the recovery success rate. No new
  event. A conflict is never report-worthy (it's the user's own dependencies).

## Live against the shipped CLI

Unlike when this was first written, **CLI v1.16.0 (on `main`) emits
`E_PROVISION_CONFLICT`**, so the recovery path is now exercised end-to-end, not
dormant. Note: the CLI only populates `backupPath` on the *first* merge; a
re-run over an already-merged `pyproject.toml` is a no-op merge and returns an
empty `backupPath`, in which case the run correctly falls through to the generic
path (the truthy-`backupPath` guard).

## Testing

- `test:unit`: new coverage for the conflict copy + fallback doc link, the
  `run`-action dispatch, `openProjectFile`, the controller conflict→retry flow
  (restore-before-retry ordering, `--no-constraints` + adopt on success,
  `conflict_retry` telemetry paired with its outcome, open pyproject at the run's
  cwd, `reportOffered: false`, missing/empty-`backupPath` and non-Full fallbacks,
  restore-failure aborts without a retry attempt), and the atomic/in-project
  restore wiring.
- `tsc` + `eslint` + `prettier` clean on changed files.

Supersedes #2181, which merged into the (now-defunct) preset-picker branch rather
than `main`.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/README.md
+++ b/packages/databricks-vscode/src/python-setup/README.md
@@ -19,6 +19,9 @@ the mapped message alone (see `reportSetupIssue.ts` for the closed routing list)
 such a conflict is usually the user's own declared dependencies. When the
 published constraints are what conflict, that genuine case is served by a soft,
 conditional pointer in the output log instead (see `formatSetupFailureDetail`).
+`E_PROVISION_CONFLICT` (the CLI's distinct pins-vs-local code) follows the same
+policy; the Full-preset flow additionally offers a restore-and-retry recovery
+(see `buildConflictRecoveryActions`).
 
 **Privacy posture.** The issue body carries build metadata (error code, phase,
 env key, package manager, extension/CLI versions, OS) plus the CLI's stderr. The
--- a/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
@@ -124,6 +124,8 @@ function makeDeps(
             compute: {kind: "serverless", version: "5"},
         }),
         adoptInterpreter: async () => {},
+        openProjectFile: async () => {},
+        restoreProjectFile: async () => {},
         saveState: () => {},
         notify: async () => {},
         showReauthPrompt: async () => {},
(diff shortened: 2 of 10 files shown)


rugpanov:
🤖 Integration tests running for `5e672990` — ⏳.

misha-db (approved):

PR #2188, opened 2026-09-10
feat(aitools): record install error categories from CLI JSON output

rclarey:
## Changes

Run `aitools install` with `--output json` and parse the result so the AITOOLS_INSTALL telemetry event carries the CLI's error categories: a top-level `globalErrorCategory` and a per-agent `agentErrors` map, both category tokens only (never the free-form error/message strings)

## Tests

updated / added unit tests

### DIFF
--- a/packages/databricks-vscode/src/aitools/AiToolsManager.ts
+++ b/packages/databricks-vscode/src/aitools/AiToolsManager.ts
@@ -6,7 +6,12 @@ import {
     AiToolsScope,
     AiToolsSkill,
     CliWrapper,
+    ProcessError,
 } from "../cli/CliWrapper";
+import {
+    describeAiToolsInstallFailure,
+    summarizeAiToolsInstallErrors,
+} from "./aiToolsInstallOutput";
 import {Mutex} from "../locking";
 import {StateStorage} from "../vscode-objs/StateStorage";
 import {WorkspaceFolderManager} from "../vscode-objs/WorkspaceFolderManager";
@@ -15,6 +20,7 @@ import {Telemetry} from "../telemetry";
 import {
     AiToolsCursorPluginSource,
     AiToolsInstallSource,
+    EventReporter,
     Events,
 } from "../telemetry/constants";
 import {Loggers} from "../logger";
@@ -469,28 +475,13 @@ export class AiToolsManager implements Disposable {
 
         const recordEvent = this.telemetry.start(Events.AITOOLS_INSTALL);
         try {
-            await this.cli.aitoolsInstall(
+            await this.runInstallCli(
                 scope,
-                this.cwdForScope(scope),
-                token,
-                agentsToInstall
+                agentsToInstall,
+                recordEvent,
+                {source, cursorPlugin},
+                token
             );
-            recordEvent({
-                result: "success",
-                scope,
-                source,
-                agents: agentsToInstall,
-                cursorPlugin,
-            });
-        } catch (e) {
-            recordEvent({
-                result: "error",
-                scope,
-                source,
-                agents: agentsToInstall,
-                cursorPlugin,
-            });
-            throw e;
         } finally {
             // Always reconcile: a failed install often still installed some
             // tools (e.g. one agent's CLI was missing), so refresh the panel to
@@ -498,6 +489,68 @@ export class AiToolsManager implements Disposable {
             // Call the `*Impl` helpers directly — we already hold the mutex.
             await this.detectInstallImpl();
             await this.resolveInstalledImpl();
+        }
+    }
+
+    /**
+     * Run `aitools install` for `agents` in `scope`, record the AITOOLS_INSTALL
+     * outcome (success/error plus any error categories), and — on a non-zero exit
+     * — throw a {@link ProcessError} so {@link AiToolsCommands} surfaces it,
+     * folding in and logging the CLI's own JSON detail when it explains the
+     * failure. `recordEvent` is the caller's already-started reporter (so
+     * `duration` covers the whole flow); `baseProps` carries the source and, for
+     * the batch install, the Cursor-plugin flag. The caller reconciles model
+     * state in its own `finally`.
+     */
+    private async runInstallCli(
+        scope: AiToolsScope,
+        agents: string[],
+        recordEvent: EventReporter<Events.AITOOLS_INSTALL>,
+        baseProps: {source: AiToolsInstallSource; cursorPlugin?: boolean},
+        token?: CancellationToken
+    ): Promise<void> {
+        const {output, error} = await this.cli.aitoolsInstall(
+            scope,
+            this.cwdForScope(scope),
+            token,
+            agents
+        );
+
+        const summarizedErrors = summarizeAiToolsInstallErrors(output);
+        recordEvent({
+            result: error !== undefined ? "error" : "success",
+            scope,
+            agents,
+            source: baseProps.source,
+            ...(baseProps.cursorPlugin !== undefined
+                ? {cursorPlugin: baseProps.cursorPlugin}
+                : {}),
+            ...(summarizedErrors.globalErrorCategory !== undefined
+                ? {globalErrorCategory: summarizedErrors.globalErrorCategory}
+                : {}),
+            ...(summarizedErrors.agentErrors !== undefined
+                ? {agentErrors: summarizedErrors.agentErrors}
+                : {}),
+        });
+
+        if (error !== undefined) {
+            const errorCode =
+                "code" in error && typeof error.code === "number"
+                    ? error.code
+                    : null;
+            const message = describeAiToolsInstallFailure(output);
+            if (message) {
+                // The CLI reported the install failed in its JSON (non-zero exit, but
+                // a parseable result). Under `--output json` the CLI's stderr is empty,
+                // so log the folded detail here, then surface it to the UI by throwing.
+                logging.NamedLogger.getOrCreate(Loggers.Extension).error(
+                    `Databricks AI tools install failed:\n${message}`
+                );
+                throw new ProcessError(message, errorCode);
+            }
+            // There was no parseable JSON output, so fallback to the pre-JSON behaviour
+            // of throwing a `ProcessError`
+            throw new ProcessError(error.message, errorCode);
         }
     }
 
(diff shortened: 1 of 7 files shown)


rugpanov:
🤖 Integration tests ❌ **failed** for `839fc721`.
[View run](https://go/deco-tests/34473556587)

rugpanov:
🤖 Integration tests ✅ **passed** for `e2ed1a14`.
[View run](https://go/deco-tests/34850986177)

packages/databricks-vscode/src/aitools/AiToolsManager.ts
```diff
@@ -501,6 +492,68 @@ export class AiToolsManager implements Disposable {
...
+            const errorCode =
+                "code" in error && typeof error.code === "number"
+                    ? error.code
+                    : null;
+            const message = describeAiToolsInstallFailure(output);
+            if (message) {
+                // The CLI reported the install failed in its JSON (non-zero exit, but
+                // a parseable result). Under `--output json` the CLI's stderr is empty,
+                // so log the folded detail here, then surface it to the UI by throwing.
+                logging.NamedLogger.getOrCreate(Loggers.Extension).error(
```
misha-db, line 546: Duplicate logging per failure?
CliWrapper.aitoolsInstall logs the raw error and `runInstallCli` logs the folded JSON, producing two log entries per failure.

packages/databricks-vscode/src/telemetry/constants.ts
```diff
@@ -424,6 +426,25 @@ export class EventTypes {
...
+        agentErrors: {
+            comment:
+                "Per-agent install errors as a map of agent id to its categorical error, for " +
+                'each requested agent that did not install (failed or skipped) — e.g. {"codex":' +
+                '"UNSUPPORTED_SCOPE"}. Keyed on the agent id so a category can be attributed to a ' +
+                "specific agent (exploded into one row per agent downstream). Categories are the " +
+                "CLI's closed set; an unrecognised token collapses to 'other'. Never the free-form " +
+                "per-agent message. JSON-stringified into a property by recordEvent. Omitted when " +
+                "every requested agent installed, or when the bundled CLI predates JSON install output",
+        },
```
misha-db, line 447: comments say an unrecognised token "collapses to 'other'", but summarizeAiToolsInstallErrors emits the CLI's raw categorym no client-side normalization. So a new CLI category reaches telemetry as is.

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -223,6 +223,94 @@ export interface AiToolsListResult {
...
+    };
+    const hasAgents = Array.isArray(rec.agents);
+    const hasError =
+        typeof rec.error === "string" || typeof rec.error_category === "string";
+    if (!hasAgents && !hasError) {
+        return undefined;
+    }
+    return {
+        ...(obj as AiToolsInstallOutput),
+        agents: hasAgents ? (rec.agents as AiToolsInstallAgentResult[]) : [],
```
misha-db, line 310: nit: checks field presence only, then and casts to AiToolsInstallOutput. A doc like {"error_category":"X","error":123} passes and returns error as a number typed as string

Should we be more defensive?

packages/databricks-vscode/src/aitools/aiToolsInstallOutput.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (output === undefined) {
+        return undefined;
+    }
+    const lines: string[] = [];
+    if (output.error) {
+        lines.push(output.error);
+    }
+    for (const agent of output.agents) {
+        if (agent.status !== "installed") {
+            const detail = agent.message ?? "failed to install";
```
misha-db, line 28: Does this treat every non installed agent, including a legitimate skipped ones as a failure (with message "failed to install")?

misha-db (changes requested):

packages/databricks-vscode/src/aitools/aiToolsInstallOutput.ts
```diff
@@ -0,0 +1,61 @@
...
+        return {};
+    }
+    const summary: AiToolsInstallErrorSummary = {};
+    if (output.error_category !== undefined) {
+        summary.globalErrorCategory = output.error_category;
+    }
+    const agentErrors: Record<string, string> = {};
+    for (const agent of output.agents) {
+        if (agent.error_category !== undefined) {
+            agentErrors[agent.name] = agent.error_category;
```
rugpanov, line 54: **Normalize `error_category` to a closed set before it reaches telemetry**

This copies the CLI's `error_category` verbatim into both `globalErrorCategory` and every `agentErrors` value. But the event's own doc comment in `constants.ts` promises *"a closed category token … an unrecognised token collapses to `'other'`"*, and nothing here (or downstream, as far as I can see) actually performs that collapse.

This is inconsistent with how we handle CLI-sourced strings elsewhere in telemetry:
- `telemetry/README.md` (Privacy): *"Only categorical/enum values and durations — no file paths, cluster names or IDs … or user content."*
- `envKey` is the direct precedent — it pattern-checks the CLI value to *"keep a closed vocabulary; anything unrecognised collapses to `'other'`"*, precisely because minimally-parsed CLI JSON can carry user-chosen content.
- Other categorical props are typed as closed unions (e.g. `type?: "dabs" | "legacy" | "unknown"`); these two are `string` / `Record<string, string>`, i.e. open.

Two concrete risks:
1. **High cardinality.** An unbounded category value — especially `agentErrors`, which is a whole map JSON-stringified into a single property value — puts unbounded distinct values on the wire, which is exactly what strains/breaks telemetry ingestion and cost (separate from privacy).
2. **Leak.** If the CLI ever emits something free-form in `error_category` (a bug, or an older/newer CLI), it flows straight through unchecked.

Suggestion: define the known category set client-side and collapse anything else to `"other"` (mirroring `envKey`/`UNRECOGNISED_ENV_KEY`), and type the fields as that union rather than open `string`. If the collapse is actually meant to happen server-side, could we update the comment to say so and point at where it happens?

packages/databricks-vscode/src/aitools/AiToolsManager.ts
```diff
@@ -501,6 +492,68 @@ export class AiToolsManager implements Disposable {
...
+        const {output, error} = await this.cli.aitoolsInstall(
+            scope,
+            this.cwdForScope(scope),
+            token,
+            agents
+        );
+
+        const summarizedErrors = summarizeAiToolsInstallErrors(output);
+        recordEvent({
+            result: error !== undefined ? "error" : "success",
```
rugpanov, line 521: **`result` and the recorded categories can disagree — a zero-exit "success" can carry errors**

`result` is derived only from the process exit (`error !== undefined ? "error" : "success"`), while `summarizeAiToolsInstallErrors(output)` attaches categories independently — and it keys on `error_category` for *any* agent regardless of `status`.

So a request the CLI **skips** on a clean exit — e.g. `codex` in `project` scope (`status: "skipped", error_category: "UNSUPPORTED_SCOPE"`, the `perAgentSkip` shape), which exits 0 — records `result: "success"` **with** `agentErrors: { codex: "UNSUPPORTED_SCOPE" }`. A dashboard counting events with a non-empty `agentErrors` as failures would overcount, and the success flag would contradict the category data.

This may be intentional (the `agentErrors` comment does say "failed or skipped") — if so, could we make it explicit and add a test? Today every category test pairs categories with a non-zero `error`, so the zero-exit-with-skip path is uncovered. If skips shouldn't count, gating on `status === "failed"` (and/or on `error !== undefined`) keeps the two in sync. Minor related nit: `summarizeAiToolsInstallErrors` records a category even for `status: "installed"` agents, since it never checks `status`.

anton-107 (approved):
Reviewed the install JSON parsing, failure handling, and telemetry path. No blocking issues found.

packages/databricks-vscode/src/aitools/AiToolsManager.ts
rclarey: Yeah as the comment mentions we log here because `stderr` is empty with `--output json`, so we need special handling for that case. I didn't want to miss non-JSON reported errors so I kept the existing log in `CliWrapper.aitoolsInstall`, but you're right maybe having two lines is confusing. I'll handle both in AiToolsManager 👍

packages/databricks-vscode/src/telemetry/constants.ts
rclarey: Missed this when I removed category parsing, will update

packages/databricks-vscode/src/aitools/aiToolsInstallOutput.ts
rclarey: No, `output.agents` only contains the agents that were attempted (i.e. the ones the user selected in the picker), so that's why if the status is not `installed` then it means it failed. Also this function is only called for non-zero exit codes, so there is no legitimate skips here.

packages/databricks-vscode/src/aitools/aiToolsInstallOutput.ts
rclarey: This is intentional. The error categories are already an enum emitted from the CLI so this extension just treats them as opaque and forwards to telemetry to we can associate the error category with other extension telemetry. Parsing and emitting the same enums the CLI outputs just adds error surface where the two can drift apart on a CLI bump

packages/databricks-vscode/src/aitools/AiToolsManager.ts
rclarey: Yes this is intentional. The CLI is the source of truth for error categorization so this PR intentionally doesn't add it's own logic and forwards any error categories it receives. I will add a test for 0 exit + skip error category to codify that this is intentional.

> e.g. codex in project scope (status: "skipped", error_category: "UNSUPPORTED_SCOPE", the perAgentSkip shape), which exits 0

This is only true when codex is auto-detected. When `--agents` is passed (which this extension always does) then `UNSUPPORTED_SCOPE` because a hard error and the CLI exits 1. 

> A dashboard counting events with a non-empty agentErrors as failures would overcount

This is a dashboard error, exit code is the only indicator of "success" vs "error". The reason we always report error categories from the CLI, even for 0 exit with skips, is because otherwise we wouldn't get any telemetry about those skips (e.g. "How often is codex auto-detected but skipped for a project scoped installation?")

rugpanov:
🤖 Integration tests running for `0e776545` — ⏳.

misha-db (approved):

PR #2192, opened 2026-09-11
fix(python-setup): report the chosen preset in the setup success panel

rugpanov:
## Why

The Python-setup success panel ("View Details" log) derived its wording from `result.mode`, which encodes only the **databricks-connect** axis (`default` vs `constraints-only`) — not whether dependency pins were written. So it misreported two of the three presets:

- A **DB Connect** run (`--no-constraints`) is byte-identical to **Full** in `mode`, so the panel always printed `• Added matching Databricks constraints to pyproject.toml` — false; DB Connect writes no pins.
- The **Python** preset (`--no-constraints --no-dbconnect`) maps to `constraints-only`, which the headline mislabeled "constraints applied" — backwards.

## What

Thread the run's resolved skip flags (`skipConstraints`/`skipDbconnect`, from `presetToFlags`) through `showSuccess` into `formatSetupLog`, and key the panel off those axes instead of `result.mode` (which it no longer reads):

- headline: `Python environment ready.` when databricks-connect is skipped, else `… ready for Databricks Connect.`
- databricks-connect version line: only when the run kept databricks-connect
- constraints bullet: only when pins were actually written

| Preset | Headline | `databricks-connect` line | Constraints bullet |
|---|---|---|---|
| Full | for Databricks Connect | yes | yes |
| DB Connect (`--no-constraints`) | for Databricks Connect | yes | **no** |
| Python (`--no-constraints --no-dbconnect`) | plain "ready" | no | no |

No persisted state or CLI-contract change — the panel is transient output.

## Testing

- Unit (42 passing): `formatSetupLog` three-preset matrix; per-preset flag wiring into `showSuccess`; and the Full→DB Connect **conflict-retry** case, asserting the panel describes the retry that actually ran (no pins), not the user's original Full pick.
- `tsc` typecheck, prettier, and eslint clean on the changed files.
- Not run locally: the full integration-test suite (covered by CI on this PR).

## Reviewer notes

- The panel now reports the run's *requested* flags rather than a CLI-confirmed "pins written" signal — deliberate, because the `--output json` contract exposes no such field (the premise of this fix). On a successful run the CLI honors the flags, so there is no realistic divergence; a future CLI that silently skipped pins on a Full run would make the bullet wrong, and warnings remain the escape hatch.

### DIFF
--- a/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
+++ b/packages/databricks-vscode/src/python-setup/controllers/PythonSetupEnvironmentSetup.test.ts
@@ -1877,6 +1877,36 @@ describe("PythonSetupEnvironmentSetup constraint-conflict recovery", () => {
         expect(setup.ready).to.equal(true);
     });
 
+    it("hands showSuccess the DB Connect flags after a Full-preset conflict retry, so the panel omits the pins line", async () => {
+        const cli = makeScriptedCli([conflictResult(), SUCCESS_REAL_RUN]);
+        let successFlags: SetupPresetFlags | undefined;
+        let retryAction: PythonSetupErrorAction | undefined;
+        const setup = new PythonSetupEnvironmentSetup(
+            makeDeps({
+                cli,
+                pickSetupPreset: async () => "full",
+                showSuccess: async (_r, flags) => {
+                    successFlags = flags;
+                },
+                showError: async (_message, _detail, actions) => {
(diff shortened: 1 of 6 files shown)


rugpanov:
🤖 Integration tests running for `8d7b3673` — ⏳.

rugpanov:
🤖 Integration tests ✅ **passed** for `1ff7dd9e`.
[View run](https://go/deco-tests/34597509108)

misha-db (approved):

PR #2195, opened 2026-09-11
test(ssh): catch bundled CLI truncation and reconnect regressions

anton-107:
## Changes

Add a regression test to the unit-test suite that exercises the real bundled CLI through the extension's SSH command construction. A localhost HTTP/WebSocket peer replaces workspace provisioning and echoes an 8 MiB payload.

Assert complete upload and download, byte equality, successful exit, and a single connection. Checking payload completeness catches CLI 1.16.0's truncated streams even when it exits zero. The test requires no workspace credentials and adds explicit WebSocket test dependencies.

The peer advertises resume capability to exercise CLI 1.16.0's affected transport, but does not send ACK control frames. CLI 1.16.1 does not support those frames. A missing bundled binary skips the test locally and fails with a fetch instruction in CI.

## Tests

- Revised test: CLI 1.16.0 fails on an incomplete transfer; 1.16.1 and 1.12.1 pass.
- Verified missing-binary behavior: one skipped test locally, explicit failure in CI.
- Lint, formatting, and diff checks passed for the revised sources.
- Previous PR revision passed the full Linux and Windows unit-test jobs, including this regression test. The latest revision awaits fresh CI.
- Earlier regression validation also confirmed CLI 1.16.0 fails and 1.16.1 passes on Windows and against extension v2.17.0 source.

### DIFF
--- a/packages/databricks-vscode/src/ssh/SshCommands.transfer.test.ts
+++ b/packages/databricks-vscode/src/ssh/SshCommands.transfer.test.ts
@@ -1,0 +1,235 @@
+import assert from "node:assert/strict";
+import {spawn} from "node:child_process";
+import {randomBytes} from "node:crypto";
+import {mkdtemp, rm, writeFile} from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import {anything, instance, mock, reset, spy, verify, when} from "ts-mockito";
+import {env, EventEmitter, window} from "vscode";
+import type {
+    ExtensionContext,
+    QuickPick,
+    QuickPickItem,
+    Terminal,
+    TerminalOptions,
+} from "vscode";
+import {CliWrapper} from "../cli/CliWrapper";
+import {ClusterModel} from "../cluster/ClusterModel";
+import {ConnectionManager} from "../configuration/ConnectionManager";
+import {DatabricksWorkspace} from "../configuration/DatabricksWorkspace";
+import type {AuthProvider} from "../configuration/auth/AuthProvider";
+import {LoggerManager} from "../logger";
+import {Cluster} from "../sdk-extensions/Cluster";
+import {createSshTunnelTestServer} from "../test/sshTunnelTestServer";
+import {SshCommands} from "./SshCommands";
+
+// Remote VS Code startup transfers several MiB; CLI 1.16.0 cut streams off at 1 MiB.
+const PAYLOAD_BYTES = 8 * 1024 * 1024;
+
+describe("SSH tunnel command transfer", function () {
+    this.timeout(30_000);
+
+    it("transfers the complete IDE-sized payload through the bundled CLI without reconnecting", async () => {
+        const directory = await mkdtemp(
+            path.join(os.tmpdir(), "ssh-transfer-")
+        );
+        const peer = await createSshTunnelTestServer();
+        const accept = new EventEmitter<void>();
+        const hide = new EventEmitter<void>();
+        const changed = new EventEmitter<void>();
+        const windowSpy = spy(window);
+        const envSpy = spy(env);
+        const pickerMock = mock<QuickPick<QuickPickItem>>();
+        const picker = instance(pickerMock);
+        let ssh: SshCommands | undefined;
+        try {
+            const configFile = path.join(directory, "databrickscfg");
+            await writeFile(
+                configFile,
+                `[ssh-test]\nhost = ${peer.host}\ntoken = local-test-token\nauth_type = pat\n`
+            );
+            const context = mock<ExtensionContext>();
+            when(context.asAbsolutePath(anything())).thenCall(
+                (relative: string) => path.resolve(__dirname, "../..", relative)
+            );
+            const cli = new CliWrapper(
+                instance(context),
+                instance(mock(LoggerManager))
+            );
+            const auth = mock<AuthProvider>();
+            /* eslint-disable @typescript-eslint/naming-convention */
+            when(auth.toEnv()).thenReturn({
+                DATABRICKS_HOST: peer.host,
+                DATABRICKS_TOKEN: "local-test-token",
+                DATABRICKS_AUTH_TYPE: "pat",
+                DATABRICKS_CONFIG_PROFILE: "ssh-test",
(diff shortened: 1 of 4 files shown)


rugpanov:
🤖 Integration tests ❌ **failed** for `cad1eee6`.
[View run](https://go/deco-tests/34605133897)

packages/databricks-vscode/src/test/sshTunnelTestServer.ts
```diff
@@ -0,0 +1,141 @@
...
+    sockets.on("connection", (socket, request) => {
+        connections++;
+        const url = new URL(request.url!, "http://localhost");
+        if (url.pathname !== "/driver-proxy-api/o/1/test-cluster/7772/ssh") {
+            unexpectedRequests.push(`WebSocket ${url.pathname}`);
+        }
+        if (request.headers.authorization !== "Bearer local-test-token") {
+            unexpectedRequests.push("WebSocket auth was not forwarded");
+        }
+        const resumable = url.searchParams.has("delivered");
```
rugpanov: **[must fix]** This 64 KiB / 50 ms acknowledgement block is dead code on the happy path, so the scenario it documents is never actually tested.

I instrumented this server and ran the test against the real bundled CLI 1.16.1: the tunnel WebSocket connects as `/driver-proxy-api/o/1/test-cluster/7772/ssh?id=<uuid>` — there is **no `delivered` query param** — so `resumable` is always `false` and **zero acks are ever sent** (verified: `resumable=false`, ack count `0`). The behaviour the comment below describes ("healthy in-flight data must not be mistaken for a dead peer") is therefore not exercised; the peer degrades to a plain byte echo.

The core truncation and reconnect-count coverage is unaffected — this only weakens the ack-latency part. Suggest either gating on the param the CLI actually sends (`id`) / sending acks unconditionally **and** asserting at least one ack was exchanged, or dropping this block and its comment so the coverage claim matches reality.

packages/databricks-vscode/src/ssh/SshCommands.transfer.test.ts
```diff
@@ -0,0 +1,235 @@
...
+import {createSshTunnelTestServer} from "../test/sshTunnelTestServer";
+import {SshCommands} from "./SshCommands";
+
+// Remote VS Code startup transfers several MiB; CLI 1.16.0 cut streams off at 1 MiB.
+const PAYLOAD_BYTES = 8 * 1024 * 1024;
+
+describe("SSH tunnel command transfer", function () {
+    this.timeout(30_000);
+
+    it("transfers the complete IDE-sized payload through the bundled CLI without reconnecting", async () => {
```
rugpanov: **[nice to have]** This test hard-fails instead of skipping when the bundled CLI hasn't been fetched.

`bin/` is gitignored and neither `build` nor `test:unit` runs `package:cli:fetch` (only CI's `unit-tests.yml` does), so a fresh-checkout `yarn test:unit` produces:

```
AssertionError [ERR_ASSERTION]: CLI exit=127 ...; uploaded=0, downloaded=0, expected=8388608
/bin/sh: .../bin/databricks: No such file or directory
```

The cause is visible in the diagnostic, so this isn't a blocker — but a `this.skip()` guard (e.g. on `existsSync(cli.cliPath)`) would let local runs degrade gracefully instead of reporting a byte-count mismatch.

rugpanov:
🤖 Integration tests ✅ **passed** for `e986dfa0`.
[View run](https://go/deco-tests/34614760474)

packages/databricks-vscode/src/test/sshTunnelTestServer.ts
anton-107: Fixed in e986dfa0. Removed the unused ACK block and its comment, and corrected the PR description. CLI 1.16.1 does not support those ACK control frames, so merely switching the condition to `id` would introduce invalid protocol traffic. Reverified the resulting byte-transfer test: CLI 1.16.0 fails on incomplete transfer; 1.16.1 and 1.12.1 pass.

packages/databricks-vscode/src/ssh/SshCommands.transfer.test.ts
anton-107: Fixed in e986dfa0. The test now checks for the bundled binary before creating the server or temporary test resources. A missing binary skips the test locally; in CI it fails explicitly with the `package:cli:fetch` command, so a packaging/setup problem cannot silently remove regression coverage. Verified both paths: one pending test locally and the expected failure in CI.

packages/databricks-vscode/src/ssh/SshCommands.transfer.test.ts
```diff
@@ -0,0 +1,243 @@
...
+            await started;
+            verify(windowSpy.showErrorMessage(anything())).never();
+            verify(terminal.sendText(anything())).once();
+            assert.ok(command, "accepting compute must launch the CLI");
+            assert.ok(terminalOptions);
+
+            // Enter the same ProxyCommand path Remote SSH uses after provisioning.
+            // Only cloud setup and the remote peer are replaced; the bundled CLI is real.
+            command +=
+                " --proxy --metadata=ssh-test,7772,test-cluster --profile=ssh-test";
```
rclarey, line 149: What do we expect `command` to be here? Should it be something like
```
'<user folder>/.vscode/extensions/databricks.databricks-2.17.1/bin/databricks' ssh connect --ide=vscode --auto-approve
```
which I see when I launch the SSH tunnel from the extension?
Why do we need to add extra flags?

anton-107: without the `--proxy` flag the command takes the provisioning branch and fails in this test, which is more of a canary test for the bundled cli (it checks that the cli can handle 8MiB of data transfer)

`--metadata` also skips certain steps in the cli process

rclarey: Got it. The reason I ask is because if we're going to change the command anyway (which I agree is fine for a canary test) then maybe we can skip the 70-80 lines of mocking and setup above which are needed to get `command` and just hard code the command instead.

Not blocking though, so I'll approve this as is

rclarey (approved):

PR #2197, opened 2026-09-14
feat(telemetry): emit setupPreset on the python_env.setup.result event

rugpanov:
## Why
`setupPreset` (full / dbconnect / python) was emitted only on `python_env.setup.attempt`. The `python_env.setup.result` event carries outcome/errorCode but only the legacy two-value `mode`, which can't separate **Full** from **DB Connect** (both report `default`). In `telemetry_refined` there is no run-correlation id and a session produces several attempt/result pairs (initial, rerun, conflict_retry), so splitting provision errors/outcomes by the three-way setup mode via an attempt↔result join is unreliable.

## What
Emit `setupPreset` on the result event too — mirrored from the attempt the reporter already closes over — and add it to the `PYTHON_ENV_SETUP_RESULT` schema as an optional field. The standalone `no_compute` result (no attempt) omits it. This makes outcome / recovery split-by-mode a single-event slice, no join.

## Testing
- `yarn test:unit` (1148 passing): the result event now carries `event.setupPreset`; `no_compute` still omits it; the schema-allowlist guard test updated to expect the new key.
- `yarn test:lint`: clean.

## Reviewer notes
- **`telemetryVersion` intentionally left at `1.0`.** The `index.ts` comment asks for a minor bump on additive fields, but that convention hasn't been followed in practice — the attempt event's own `setupPreset` / `trigger` / `isGreenfield` were all added at `1.0` — and the version is global across every event. A lone bump here would be inconsistent and churn every event's version assertion without making the version an accurate schema history. Treated as pre-existing stale convention.
- **Only `setupPreset` is mirrored**, one of six attempt-only dimensions. A single run-correlation id on both events would make every dimension joinable and is the more general fix; deferred as a follow-up (higher cardinality, revisits the categorical-only telemetry discipline).

## Follow-up
Unblocks the join-free outcome-by-mode / recovery-by-mode dashboard panels on the Deco IDE Usage / VPEX Usage page.


rugpanov:
🤖 Integration tests ❌ **failed** for `0553777f`.
[View run](https://go/deco-tests/34839112459)

misha-db (approved):

PR #2205, opened 2026-09-22
Add a setting to override CLI path

### LINKED ISSUE #2199: [BUG] VSCode extension doesn't allow the use of the locally installed CLI version
**Describe the bug**
As per the [FAQ for the VSCode extension](https://docs.databricks.com/aws/en/dev-tools/vscode-ext/faqs#what-if-i-have-the-databricks-cli-already-installed), the extension should allow the use of an already installed version of the CLI. However, it doesn't do this automatically, and there is no setting to point the extension to where the extension is already installed. The only option is to use the bundled cli in `%userprofile%/.vscode/extensions/databricks.databricks-2.16.0-win32-x64/bin/databricks.exe`.

Our company's IT policies/controls prevent .exe being run unless there are installed from our internal App Store. This App Store has the databricks cli which is installed to `c:/program files/databricks/databricks.exe` and runs correctly. However, the VSCode extension can't be pointed to it.

This results in the VSCode extension not running properly as it just triggers a Windows popup for admin access which our users don't have:

```
The application you are attempting to run is not permitted in this environment.

Executable: DATABRICKS.EXE
Path: %USERPROFILE%\.VSCODE\EXTENSIONS\DATABRICKS.DATABRICKS-2.16.0-WIN32-X64\BIN
Parent Process: cmd.exe /d /c ""%USERPROFILE%\.vscode\extensions\databricks.databricks-2.16.0-win32-x64\bin\databricks.exe" "bundle" "validate" "--target" "feature_branch""

Deciding Rule: Trusted Ownership

Click here for instructions to run this software
```

**To Reproduce**

1. Install VSCode extension
2. Try to run without admin permissions on a machine where user install and running of .exe are forbidden unless officially installed from company App Store
3. Try to point VSCode extension to existing install of the databricks cli

**System information:**

Version: 1.137.0 (user setup)
Commit: 645f29cc3176500b4b5762ba887cf2a7f0ffdf2c
Date: 2026-09-08T13:43:59-07:00
Electron: 42.10.0
ElectronBuildId: 15109253
Chromium: 148.0.7778.280
Node.js: 24.18.1
V8: 14.8.178.38-electron.0
@github/copilot: 1.0.84.70.gdb75d0d
@github/copilot-sdk: 1.0.13-preview.4
OS: Windows_NT x64 10.0.26200

Databricks Extension Version: 2.16.0

rclarey:
## Changes

Closes #2199 

Add a new setting to override the bundled CLI path. If this setting is set, and the overridden CLI version is older than the bundled CLI, a warning is shown on startup

## Tests

Added unit tests


rugpanov:
🤖 Integration tests ❌ **failed** for `1fb04b7d`.
[View run](https://go/deco-tests/35702693805)

rugpanov:
🤖 Integration tests ✅ **passed** for `af4fd23c`.
[View run](https://go/deco-tests/35739418848)

packages/databricks-vscode/package.json
```diff
@@ -1856,6 +1856,10 @@
                         "type": "number",
                         "default": 5,
                         "description": "The interval in minutes at which the remote state of the bundle is refreshed."
+                    },
+                    "databricks.overrideDatabricksCLIPath": {
+                        "type": "string",
```
rugpanov, line 1861: **Security: this setting needs a non-workspace `scope`.**

With no `scope`, `overrideDatabricksCLIPath` defaults to `window` scope, so it can be set from a workspace's checked-in `.vscode/settings.json` — not just the user's own settings. On activation `warnOverridenCliDrift()` immediately spawns `<override> version`, and the path is then forwarded as `DATABRICKS_CLI_PATH` to every CLI / SDK / Terraform call. So opening a repo that ships a malicious `.vscode/settings.json` (pointing at a binary committed in that same repo) is silent arbitrary code execution once the workspace is trusted — no user action beyond opening the folder.

This is exactly the class VS Code hardened `git.path`, `python.defaultInterpreterPath`, `php.validate.executablePath`, etc. against: executable-path settings should be `machine` or `machine-overridable` so a workspace file can't redirect which binary runs. `machine-overridable` still lets remote/dev-container user settings override it, which is likely what you want here.

```suggestion
                    "databricks.overrideDatabricksCLIPath": {
                        "type": "string",
                        "scope": "machine-overridable",
```

packages/databricks-vscode/src/cli/CliWrapper.test.ts
```diff
@@ -501,6 +540,278 @@ token = dapitest5678
...
+                        Patch: patch,
+                    };
+                    /* eslint-enable @typescript-eslint/naming-convention */
+                    const overridePath = path.join(
+                        tmpDir,
+                        `databricks-${overrideOutput.Tag}`
+                    );
+                    await writeFile(
+                        overridePath,
+                        `#!/usr/bin/env bash\necho '${JSON.stringify(
```
rugpanov: **This will fail on the Windows CI runner.** `unit-tests.yml` runs the suite on `windows-server-latest`, and these `warning: true` cases write the fake override CLI as a `#!/usr/bin/env bash` script with no extension. On Windows that file is spawned through `cmd.exe` (via `execFile`) and can't execute, so `getCliVersion(overridePath)` returns `undefined`, `warnOverridenCliDrift` returns early without warning, and every warning-expected case fails `assert.equal(showWarningCalls.length, 1)`.

Options: gate the `warnOverridenCliDrift` suite behind `process.platform !== "win32"`, or on Windows write a `.cmd`/`.bat` shim (named accordingly) instead of a bash script. Worth confirming against a Windows run before merge.

rugpanov:
🤖 Integration tests running for `3fb242c7` — ⏳.

rugpanov:
🤖 Integration tests running for `bfd44399` — ⏳.

.isaac/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "sync_reminder_last_shown": "2026-09-23"
```
anton-107: don't think you meant to include this file to this PR?

perhaps add .isaac to .gitignore while you are on this?

packages/databricks-vscode/src/cli/CliWrapper.ts
```diff
@@ -1192,4 +1289,81 @@ export class CliWrapper {
...
+            return;
+        }
+
+        const env = {
+            ...EnvVarGenerators.getEnvVarsForCli(this.extensionContext),
+            ...EnvVarGenerators.getProxyEnvVars(),
+        };
+        const [overrideResult, bundledResult] = await Promise.all([
+            getCliVersion(overridePath, env),
+            getCliVersion(defaultCliPath(this.extensionContext), env),
```
anton-107: The person who filed #2199 can't run the bundled databricks.exe at all. Their company blocks it, and every attempt pops up an admin-access prompt. That prompt is the bug. With this PR, warnOverridenCliDrift still runs bin/databricks version --output json on every activation, just to learn the bundled version. So that user still gets the prompt every time they open a window. The spawn isn't needed: the pinned version is already in packageMetadata.cliVersion (packageJsonUtils.ts:83, from package.json cli.version). The check should compare against that.

packages/databricks-vscode/package.json
```diff
@@ -1856,6 +1856,11 @@
...
+                        "scope": "machine-overridable",
```
anton-107: `machine-overridable` is defined in the VS Code docs as a machine setting that **can be** overridden by workspace or folder settings. So the path can still come from the repo. `machine` is the scope that blocks this, and it still allows Remote and dev-container machine settings, which is what the reviewer wanted.

  - Quick check: put `{"databricks.overrideDatabricksCLIPath": "./evil"}` in a folder's `.vscode/settings.json`. 
  - `workspace.getConfiguration("databricks").get("overrideDatabricksCLIPath")` returns `"./evil"`. With `"scope": "machine"` it returns undefined, and the Settings UI says the setting can only be applied in user settings.

rugpanov:
🤖 Integration tests running for `28d2a213` — ⏳.

anton-107 (approved):

PR #2208, opened 2026-09-23
 Bundle Resource Explorer in Databricks Remote SSH mode

misha-db:
## Changes

Until now, Databricks Remote SSH ("remote") mode surfaced only the Unity Catalog and Docs views. It connected purely from the ambient environment via ConnectionManager.connectFromEnvironment(), which deliberately never touches the ConfigModel, so there was no bundle target and none of the bundle views could render. This PR brings the Bundle Resource Explorer, Bundle Variables, and a slimmed-down Configuration view to remote mode, along with project-folder and bundle-target pickers — all without introducing a login flow.

New: RemoteBundleInitializer (src/bundle/RemoteBundleInitializer.ts)
Bridges the environment-resolved auth onto the ConfigModel so BundleRemoteStateModel can shell out to bundle summary. It enforces the invariant that, in remote mode, the ConfigModel's auth provider always mirrors the environment:
- Re-applies auth after setTarget clears it, by listening to onDidChangeAuthProvider (which fires from inside setAuthProvider while the config mutex is held, so the re-apply queues after the clear rather than racing it).
- Re-resolves the target on a project-folder change — normal mode gets this from ConnectionManager.init(), which remote mode never runs.
- Re-applies auth on reconnect (onDidChangeState === "CONNECTED").
- Dedupes by auth-provider reference equality to avoid redundant bundle summary calls and self-looping. Two mutexes (applyAuthMutex, folderChangeMutex) serialise overlapping triggers with an acyclic lock order.

New: RemoteConfigurationDataProvider (src/ui/configuration-view/RemoteConfigurationDataProvider.ts)
A minimal Configuration view showing the active project folder (clickable to switch) and, once resolved, the bundle target with its Host/Mode children — enough to answer "which project is selected" and "which workspace will a deploy target." It drops the normal provider's BundleProjectManager gate and the login/cluster/sync/environment components that don't exist in remote mode, and suppresses BundleTargetComponent until a folder is active.

Refactor: extension.ts
Extracted the Bundle Resource Explorer and Bundle Variables registration into shared registerBundleResourceExplorer / registerBundleVariablesView helpers, called from both the normal and remote activation flows (the only difference is that remote mode passes no ConfigurationDataProvider to the decoration provider). The remote flow now wires up the initializer, the remote Configuration view, both bundle views, and lightweight selectActiveProjectFolder / selectTarget command registrations.

Refactor: shared bundle-target picker
ConnectionCommands.selectTarget and the new remote databricks.connection.bundle.selectTarget registration now share promptToSelectBundleTarget (src/bundle/activeBundleUtils.ts) — the two flows have only the ConfigModel in common.

DecorationProvider now takes an optional ConfigurationDataProvider (absent in remote mode).

package.json
- Dropped the !databricks.context.remoteMode guard from the bundle commands' enablement clauses so they're usable in remote mode.
- configurationView now shows when databricks.context.activated; dabsResourceExplorerView shows when a target is set or in remote mode; dabsVariableView when a target is set.
- Added viewsWelcome entries prompting "Select a project" when no target is resolved in remote mode, and a selectActiveProjectFolder navigation item on the resource explorer in remote mode.

## Tests

- RemoteBundleInitializer.test.ts — covers applying env auth once a target resolves, no-op when no target, re-apply on onDidChangeAuthProvider, re-apply on reconnect (and ignoring non-connected states), reference-equality dedup, folder-change re-resolution, no-op when the new folder has no target, serialisation of overlapping folder changes, and swallowing connectFromEnvironment failures.
- RemoteConfigurationDataProvider.test.ts — Local Folder + Target rows once resolved, target picker offered only when a folder is active, picker suppressed with no folder, and copy-kind stamping via getTreeItem.
- ConfigModel.test.ts (new) — locks the target/auth ordering invariant the initializer depends on: setTarget clears the auth provider on every path, so auth must always be re-applied after the target, never before.


rugpanov:
🤖 Integration tests ❌ **failed** for `bb3025c3`.
[View run](https://go/deco-tests/35887407492)

rugpanov:
🤖 Integration tests running for `756b260d` — ⏳.

